PR #65, opened 2013-06-25
Add `direnv invoke WORKDIR EXECUTABLE [ARGS...]`

pwaller:
Does what you might imagine. Establishes the direnv environment for `$WORKDIR`
before running `$EXECUTABLE $ARGS`.

On platforms where exec is available, we exec into the target executable.
Otherwise `os/exec.Command()` is used.

### DIFF
--- a/cmd_invoke.go
+++ b/cmd_invoke.go
@@ -3,12 +3,13 @@ package main
 import (
 	"flag"
 	"fmt"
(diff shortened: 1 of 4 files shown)


savannahsammons58-lab:
> unning `$EXECUTABLE $ARGS`.

Hey

PR #152, opened 2015-01-07
Added some documentation on how to set up the hook

danielmbarlow:
Please check that understand properly. Thanks.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,7 +60,9 @@ Links to binary builds are also available on
 ## Setup
 
 For direnv to work properly it needs to be hooked into the shell. Each shell
-has it's own extension mechanism:
+has it's own extension mechanism, which you can copy and paste from the 
+console after running the `direnv hook SHELL` command (where shell is the 
+name of shell you are using.) For example:
 
 ### BASH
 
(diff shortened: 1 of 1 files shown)


zimbatm:
Hi @danielmbarlow, thanks for your contribution. I have left a few comments on the changeset, hope I don't come off as defensive. How is your direnv experience so far ? I'm interested in getting general feedback too.

Cheers,
zimbatm

pwaller:
Yeah, profile loading is a bit of a mess and I've seen little in the way of clear guidance for newbies on the matter.

My preference is to put things like direnv or path modifications in my `.profile` (which bash will read) and then enable `run command as login shell` in my terminals.

image

`.bashrc` is run whenever bash starts interactively, including if you happen to invoke bash from bash (which you might do via tmux or screen), it will re-do the environment modifications even though they are already present. Whilst not necessarily harmful to re-do environment modifications, I've seen this lead to some confusion (and for example cause it to destroy `.bash_history` unders some configuraitons).

pwaller:
In fact, I have one thing in my `.bashrc`, and it's `shopt -s histappend`, which prevents the history from being destroyed. In my `.profile` I also have `export HISTFILE=~/.my_bash_history` which means if a bash instance starts up which doesn't have `histappend` set, it won't have `HISTFILE` set either and so it can't possibly destroy `~/.my_bash_history`.

zimbatm:
That's why direnv is best configured in the `.bashrc`, it is only needed for interactive bash sessions. The bash extension mechanism happen to depend on an environment variable (PROMPT_COMMAND) but it really is a bash configuration like `shopt`.

zimbatm:
Closing as it no longer merges

PR #153, opened 2015-01-19
Add ability to prune references to .envrc, which no more exists

avnik:
The problem:

When we enable reading  .envrc with  `direnv allow` command, we create empty file in .config/direnv/allow/, but when we remove entrie directory with .envrc, "allow hash" still exists.

Solution:
I add code to store path to .envrc in "allow-hash" file, and implement `prune` subcommand, which remove stalled files from .config/direnv/allow (old empty files still be untouched)

### DIFF
--- a/cmd_prune.go
+++ b/cmd_prune.go
@@ -1,0 +1,59 @@
+package main
+
+import (
+	"io/ioutil"
+	"os"
+	"path"
+	"strings"
+)
+
+var CmdPrune = &Cmd{
+	Name: "prune",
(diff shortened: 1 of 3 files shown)


zimbatm:
Hi @avnik, thanks for the PR. Apart from minor issues I think it's looking quite good. I'll merge it once those are fixed.

zimbatm:
@avnik are you thinking of resuming work on this-one ?

zimbatm:
Scratch that, it's better merged and improved later if necessary

PR #156, opened 2015-02-14
shell_fish: escape spaces

gfxmonk:
Spaces need to be backslash-escaped in values. Without this, fish treats any var with spaces as an array. Which _looks_ OK when you use the var directly in fish, but ends up as a `:`-separated string for child processes. i.e:

	$ cat > .envrc
	export FOO="1 2 3"
	direnv: error .envrc is blocked. Run `direnv allow` to approve its content.
	
	$ direnv allow
	direnv: loading .envrc
	direnv: export +FOO
	
	# looks good:
	$ echo $FOO
	1 2 3
	
	# uh-oh...
	$ env | grep "FOO"
	FOO=1:2:3

### DIFF
--- a/shell_fish.go
+++ b/shell_fish.go
@@ -55,6 +55,8 @@ func (f fish) Escape(str string) string {
 			escaped(`\n`)
 		case char == CR:
 			escaped(`\r`)
+		case char == SPACE:
+			backslash(char)
 		case char <= US:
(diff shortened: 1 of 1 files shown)


gfxmonk:
Actually, scratch this. I've hit more fish escaping bugs, gonna fix them all at once.

PR #158, opened 2015-02-14
shell_fish: use single quotes to simplify escaping logic

gfxmonk:
After I submitted #156 I found that other characters are incorrectly escaped for `fish` too - specifically the '*' character.

Rather than explicitly escaping special characters, this patch just wraps exported values in single quotes, only breaking out for special non-printable characters (\t, \xxx, etc). This makes the logic much simpler, and it now works properly when variables contain spaces and `*` (and perhaps others that I haven't encountered yet).

There's also a new test case, `dump`, which tests both fish & bash on a bunch of special characters, as well as covering #157 (so it'll fail until that it merged as well).

### DIFF
--- a/shell_fish.go
+++ b/shell_fish.go
@@ -18,16 +18,13 @@ end
 }
 
 func (f fish) Escape(str string) string {
-	if str == "" {
-		return "''"
-	}
 	in := []byte(str)
-	out := ""
+	out := "'"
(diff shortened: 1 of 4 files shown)


zimbatm:
Sweet, thanks @gfxmonk all this looks good to me

PR #169, opened 2015-06-03
Direnv should not fail if ../.envrc doesn't exist

gerhard:
`source_env ..` fails if `../.envrc` doesn't exist. I'm not convinced it should. What are your thoughts on this?

TL;DR

My use-case is a top-level dir which contains multiple apps, each with their specific `.envrc`. I store common values in the top-level dir which is not version-controlled. This global `.envrc` might not exist on all workstations. Currently, direnv fails:

```
/bin/bash: line 134: ./.envrc: No such file or directory
direnv: error exit status 1
```

I would much rather it gracefully handled the absence of `../.envrc`


gerhard:
All clear, PR attached.

zimbatm:
Thanks @gerhard

gerhard:
:+1:

PR #170, opened 2015-06-27
Add support for emacs export

zimbatm:
At some point I would like to add a universal format that can be understood by all shells instead.

The emacs.el script is still under construction, I never did any lisp so it may take a while.


zimbatm:
Closing in favour of #209

PR #171, opened 2015-07-04
Remove reference to ~/.config/direnv/direnvrc

gfxmonk:
It looks like this file is never loaded, and the path is not mentioned anywhere in the code (as far as I can tell). So I assume it's a remnant from some older version where this was supported.


zimbatm:
Thanks. I would like to add the support for it instead to make direnv adheres to the [xdg spec](http://standards.freedesktop.org/basedir-spec/basedir-spec-latest.html). It allows to unclutter the home directory.

gfxmonk:
Cool, even better :)

PR #172, opened 2015-07-04
Add use_nix() helper to stdlib

gfxmonk:
This lets users easily add [`nix`](http://nixos.org/nix/) expressions to their direnv environment, either by explicitly specifying packages:

    use nix -p python3 ocaml

..or by relying on the default `nix-shell` behaviour when you have a `default.nix` or `shell.nix` in the current directory:

    use nix

(This all relies on `nix` being installed, of course).

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -380,6 +380,17 @@ const STDLIB = "# These are the commands available in an .envrc context\n" +
 	"  rvm \"$@\"\n" +
 	"}\n" +
 	"\n" +
+	"# Usage: use_nix [...]\n" +
+	"#\n" +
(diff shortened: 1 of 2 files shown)


zimbatm:
Wow nice. Mine wasn't looking as good :)

```bash
use_nix() {
  source "$HOME/.nix-profile/etc/profile.d/nix.sh"
  export NIX_PATH=/nix/var/nix/profiles/per-user/$USER/channels
  export NIX_PROFILE=$PWD/.direnv/nix
  load_prefix "$PWD/.direnv/nix"
}
```

PR #179, opened 2015-09-06
Fix lintian spelling warning

punitagrawal:
Nitpick fix: lintian threw this warning when building the direnv package for debian.


pwaller:
Thanks!

PR #185, opened 2015-11-06
Add `use node` directive to stdlib

wilmoore:
- `use node` Loads NodeJS version from a `.node-version` or `.nvmrc` file.
- `use node <version>` Loads specified NodeJS version.
- requires environment variable $NODE_VERSIONS
- optional environment variable $NODE_VERSION_PREFIX

### DIFF
--- a/man/direnv-stdlib.1.md
+++ b/man/direnv-stdlib.1.md
@@ -167,6 +167,26 @@ Example:
 * `rvm` ...:
     Should work just like in the shell if you have rvm installed.
 
+* `use node`:
+    Loads NodeJS version from a `.node-version` or `.nvmrc` file.
+
+Example (.envrc):
+
+    set -e
+    use node
+
+Example (.node-version):
+
+    4.2.2
+
+* `use node` version:
+    Loads specified NodeJS version.
+
+Example (.envrc):
+
+    set -e
+    use node 4.2.2
+
 COPYRIGHT
 ---------
 
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -26,6 +26,27 @@ log_status() {
   fi
 }
 
+# Usage: log_error [<message> ...]
+#
+# Logs an error message. Acts like echo,
+# but wraps output in the standard direnv log format
+# (controlled by $DIRENV_LOG_FORMAT), and directs it
+# to stderr rather than stdout.
+#
+# Example:
+#
+#    log_error "Unable to find specified directory!"
+
+log_error() {
+  local color_normal=`tput sgr0`
+  local color_error="\e[0;31m"
+  if [[ -n $DIRENV_LOG_FORMAT ]]; then
+    local msg=$*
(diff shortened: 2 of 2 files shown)


zimbatm:
Really nice @wilmoore

wilmoore:
> Really nice @wilmoore

Thanks. I've been using it myself and definitely wanted to share. Love how you've made `direnv` so extensible.

> The rest of the script uses lower-cased local variable so that it's easier to distinguish them from environment variables

Good point. I'll make an update so that style is preserved.

wilmoore:
> What would you think of using the existing log_status function maybe add a log_error function to keep logging unified ?

I think that would be excellent.

wilmoore:
> I'm not sure why the sleep is for. Isn't just exiting the script enough ?

Yeah, one thing I've noticed is that if I put something to `stderr`, it pretty much doesn't show up. Sometimes there is a bit of a flash so I know it is being output, but doesn't stick around for long. Is there a way to ensure those messages always show up...perhaps using `stderr` is the problem.

zimbatm:
Alright, I'm done with the review. Ping me when you're done making the updates that you want, I'll have another look tomorrow.

wilmoore:
> Alright, I'm done with the review. Ping me when you're done making the updates that you want, I'll have another look tomorrow.

Sounds good. I'll make some updates tonight. Thanks!

zimbatm:
> Yeah, one thing I've noticed is that if I put something to stderr, it pretty much doesn't show up. Sometimes there is a bit of a flash so I know it is being output, but doesn't stick around for long. Is there a way to ensure those messages always show up...perhaps using stderr is the problem.

Weird. It's possible that stdout or stderr is not being flushed properly before exit but don't know how to address that. It should be flushed if your line terminates with a "\n".

wilmoore:
> It should be flushed if your line terminates with a "\n"

That did the trick :+1:

zimbatm:
Alright, if you can address these two little things it would be awesome.

wilmoore:
> Alright, if you can address these two little things it would be awesome.

All set...thanks :+1:

zimbatm:
@wilmoore thanks for putting all this effort into the PR !

wilmoore:
> ... thanks for putting all this effort into the PR !

No problem at all :+1:

wilmoore:
@zimbatm 

Wasn't sure where to ask so I figured I'd start here. After you merged this, I figured I'd go ahead and uninstall my previous installation of `direnv`:

```sh
brew uninstall --force direnv
```

and install from head:

```sh
brew install --HEAD direnv
```

Unfortunately, after doing this, everything stopped working. To ensure I wasn't missing something, I checked the man page:

```sh
man direnv-stdlib
```

I noticed that there was no `use node` section in that man page. That tells me that there is something amiss. I wonder if this is a homebrew issue or if there is some build artifact missing?

PR #187, opened 2015-11-11
Add fuzzy matching to `use node` stdlib function

wilmoore:
If you specify a partial NodeJS version (i.e. `4.2`), a fuzzy match is performed and the highest matching version installed is selected.

### DIFF
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -217,6 +217,10 @@ Should work just like in the shell if you have rvm installed.
 .IP \(bu 2
 \fB\fCuse node\fR:
 Loads NodeJS version from a \fB\fC\&.node\-version\fR or \fB\fC\&.nvmrc\fR file.
+.PP
+If you specify a partial NodeJS version (i.e. \fB\fC4.2\fR), a fuzzy
+match is performed and the highest matching version installed
+is selected.
 .RE
 .PP
 Example (.envrc):
@@ -232,7 +236,7 @@ Example (.node\-version):
 .PP
 .RS
 .nf
-4.2.2
+4.2
 .fi
 .RE
 .RS
--- a/man/direnv-stdlib.1.md
+++ b/man/direnv-stdlib.1.md
@@ -170,6 +170,10 @@ Example:
 * `use node`:
     Loads NodeJS version from a `.node-version` or `.nvmrc` file.
 
+    If you specify a partial NodeJS version (i.e. `4.2`), a fuzzy
+    match is performed and the highest matching version installed
+    is selected.
+
 Example (.envrc):
 
     set -e
@@ -177,7 +181,7 @@ Example (.envrc):
 
 Example (.node-version):
 
-    4.2.2
+    4.2
 
 * `use node` version:
     Loads specified NodeJS version.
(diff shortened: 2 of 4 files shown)


zimbatm:
Nice. Since you're here I was wondering, are you using the stdlib `layout node` as well in your .envrcs ? I was thinking that it might make sense to activate the layout at the same time

wilmoore:
> are you using the stdlib layout node as well in your .envrcs

I don't use `layout node` because adding `node_modules/.bin` to `$PATH` is not a necessary pattern with `node` and `npm`.

The general pattern is to utilize `package.json`.`scripts`.
https://docs.npmjs.com/cli/run-script

When you define a `.scripts`.`test:e2e`, for that project, you can now type:

```
npm run test:e2e
```

Even better, within `.scripts`.`test:e2e` you can access any binary dependencies that have been installed.

```
npm install --save-dev webdriverio
```

Then you can say in `package.json`:

```
{
  "scripts": {
    "test:e2e": "wdio ./test/config/wdio.conf.js"
  }
}
```

In other words, you don't have to type `./node_modules/.bin/wdio` because within `.scripts` it is added to the `$PATH` automatically.

This is somewhat like `bundle exec`; however, I hesitate to compare the two because `bundle exec` does  a lot of stuff that `npm run` does not need to do. For example, without `bundle exec`, a command doesn't automatically have access to installed Gems; however, with `node`, that isn't an issue since you don't need to do anything special to access installed packages. You just `require` and `node` knows to look into `node_modules`.

Hope that helps.

zimbatm:
Ok, thanks for letting me know where things stand in the node world.

I think that what you describe is just because it's not easy to add `./node_modules/.bin` into the path without direnv so they have come up with their own wrapper. But like you said each command needs to be prefixed with the `bundle exec` equivalent. If you just add `./node_modules/.bin` into the PATH then you can go ahead and just use `coffee` or whatever program is installed with the project.

wilmoore:
> I think that what you describe is just because it's not easy to add ./node_modules/.bin into the path without direnv so they have come up with their own wrapper.

You may very well be correct; however, I don't ATM have any historical context around it to say one way or the other.

> If you just add ./node_modules/.bin into the PATH then you can go ahead and just use coffee or whatever program is installed with the project.

I can definitely see how someone might like that as a workflow. In that regard, probably is good to have both options.

For those that want access to those local binaries, they can always add `layout node` to their `.envrc` file.

For me, I would probably only ever use `use node` along with a `.node-version` file. I wouldn't ever want to have to figure out why when I'm in a specific directory, the `host` program doesn't work as expected for example (maybe this is a bad example, but it's the first I could think of). If one of the project's dependencies were say: https://www.npmjs.com/package/host, that could put me in a situation where things act unexpectedly.

But again, it's up to the individual so both options seems great!

zimbatm:
Exactly. The trick is to put the non-controversial stuff in the stdlib and anything that's not 100% clear goes into the user's `~/.direnvrc`. Thanks for taking the time to explain your POV !

wilmoore:
> Exactly. The trick is to put the non-controversial stuff in the stdlib and anything that's not 100% clear goes into the user's ~/.direnvrc. 

Agreed!

> Thanks for taking the time to explain your POV !

Certainly :smile:

PR #191, opened 2015-12-25
Preserve previous BASH status code

### LINKED ISSUE #155: bash _direnv_hook clobbers last command's exit status
The exit status from `eval` is the exit status of the command that was evaluated and the current bash hook uses `eval` to process `direnv export bash` ahead of an existing `$PROMPT_COMMAND`.  This approach as the side effect of eating the exit status of the last shell command that was executed.

For example, executing /usr/bin/false returns 1:
```
$ /usr/bin/false; echo $?
1
```
and executing /usr/bin/true returns 0:
```
$ /usr/bin/true; echo $?
0
```

Now it's fairly common for a bash prompt to provide some indication of whether or not the previous command was successful or not. A really basic prompt would simply print out the exit status of the previous command by referencing `$?`. For example:
```
$ prompt_command() { export PS1="[rc=$?]\\$ "; }
$ export PROMPT_COMMAND=prompt_command
[rc=0]$ 
```

With this prompt command in place, calling `true` or `false` results in the expected prompt:
```
[rc=0]$ true
[rc=0]$ false
[rc=1]$ 
```

If I then evaluate the direnv hook, the behavior changes:
```
[rc=1]$ eval "$(direnv hook bash)"
[rc=0]$ true
[rc=0]$ false
[rc=0]$ echo $?
1
```

You'll notice that the prompt's view of the exit status does not reflect the status of the previously executed command at the prompt.  This is due to the fact that `$PROMPT_COMMAND` now evaluates `direnv export bash` prior to `prompt_command`:
```
[rc=0]$ echo $PROMPT_COMMAND 
_direnv_hook;prompt_command
[rc=0]$ 
```

If I simply change the bash `_direnv_hook` function from:
```bash
_direnv_hook() {
  eval "$(direnv export bash)";
};
```
to
```bash
_direnv_hook() {
  local previous_exit_status=$?;
  eval "$(direnv export bash)";
  return $previous_exit_status;
}
```
things begin to behave as expected:
```
[rc=0]$ true
[rc=0]$ echo $?
0
[rc=0]$ false
[rc=1]$ echo $?
1
[rc=0]$ 
```

Basically, `_direnv_hook` should preserve the exit status at entry and restore it at exit. If it doesn't, it breaks a number of customized bash prompts.

miketheman:
Credit to @sykesm for reporting in detail.

Fixes #155

### DIFF
--- a/shell_bash.go
+++ b/shell_bash.go
@@ -6,7 +6,9 @@ var BASH bash
 
 const BASH_HOOK = `
 _direnv_hook() {
+  local previous_exit_status=$?;
   eval "$(direnv export bash)";
+  return $previous_exit_status;
 };
 if ! [[ "$PROMPT_COMMAND" =~ _direnv_hook ]]; then
   PROMPT_COMMAND="_direnv_hook;$PROMPT_COMMAND";


zimbatm:
Thanks @miketheman, I had forgotten about this

PR #198, opened 2016-02-14
Makefile: simplify "man" task using md2man-rake(1)

sunaku:
No need to expand `*.1.md` into `*.1` extensions and track dependencies in
make(1) because md2man-rake(1)'s "md2man:man" task already does it all.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,8 +1,5 @@
 DESTDIR ?= /usr/local
 
-MAN_MD = $(wildcard man/*.md)
-ROFFS = $(MAN_MD:.md=)
-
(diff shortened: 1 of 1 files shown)


zimbatm:
Thanks. With the proposed change md2man needs to be installed even if the .md files haven't changed. This is going to be an issue for packagers like homebrew as it's adding another build dependency (usually only developers need to install md2man since the result is committed into the repo). Ideally I would like to get rid of the ruby dependency to make contributions easier.

zimbatm:
Closing as I think it's important not to add md2man as a direct build dependency to packagers

PR #205, opened 2016-03-12
Correct install instructions for zsh, bash

ptim:
Update bash and zsh entries according to: https://github.com/direnv/direnv#setup

(previously, the direnv hook was being printed to stdout when i sourced my profile, rather than being evaluated)

Other shells might need editing, but I only primarily use zsh..

Also, immediately source the profile.. is that too cheeky?

### DIFF
--- a/docs/install.rst
+++ b/docs/install.rst
@@ -91,14 +91,14 @@ Bash
 
 ::
 
-    $ echo 'eval "direnv hook bash"' >> ~/.bashrc
+    $ echo 'eval $("direnv hook bash)"' >> ~/.bashrc && source ~/.bashrc
 
 Zsh
 ^^^
 
 ::
 
-    $ echo 'eval "direnv hook zsh"' >> ~/.zshrc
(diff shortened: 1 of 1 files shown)


docs/install.rst
```diff
@@ -91,14 +91,14 @@ Bash
 
 ::
 
-    $ echo 'eval "direnv hook bash"' >> ~/.bashrc
+    $ echo 'eval $("direnv hook bash)"' >> ~/.bashrc && source ~/.bashrc
```
zimbatm: `$("` -> `"$(`

zimbatm:
Yes, thanks for the PR. I don't know why it wasn't caught before but it was obviously wrong.
There is just a little fix and then I can merge.

zimbatm:
Fixed in 3aebe0dde8b23e71a224e2934367685d822c0951

zimbatm:
I'm not sure if I will go forward with the sphinx doc

Issue #173, opened 2015-07-04
idea: support $DIRENV_CACHE


gfxmonk:
I have a nontrivial direnv setup to cache the vars required for complex enviornments. The basic setup is that I use `nix-shell`, but for large dependency trees this can take a handful of seconds, which is annoying to wait for when I open a new shell.

So I have this script, which serializes the vars to a file and then just uses `direnv_load cat $vars`: https://github.com/gfxmonk/app-customisations/blob/bd20e835741a1c3e49162101cf431ef9d54cbd39/home/direnv-std/nix-shell#L30

Never mind the details of the script, the only relevant point is that I set `$DIRENV_CACHE` to this cached vars file. The script _tries_ to regenerate this file when its inputs change, but it's not always correct so I occasionally need to explicitly remove this cache.

To do that, I made this little script [`direnv-reload`](https://github.com/gfxmonk/app-customisations/blob/bd20e835741a1c3e49162101cf431ef9d54cbd39/bin/direnv-reload) which removes `$DIRENV_CACHE` if it's set, and then runs the typical `direnv reload`. What do you think about rolling this script's functionality into `direnv reload` itself? It would provide a standard way for .envrc scripts to use caching, while regenerating the cache whenever reload is explicitly requested. I'm not sure how common this is likely to be (it obviously requires a much more complex .envrc script), but the integration effort on direnv's part is pretty minimal.

We may want to name the variable something more explicit just to make it impossible to accidentally set, maybe something like $DIRENV_DELETE_ON_RELOAD or something painfully obvious ;)

CMCDragonkai:
How are you using nix shell with direnv?

gfxmonk:
Well the caching code is a little complex, but without caching it's pretty simple (PR in #172):

    direnv_load nix-shell --show-trace --run 'direnv dump'

zimbatm:
It's a great idea.

Most of the necessary mechanism is in place to integrate that into direnv directly I think. It would be nice if the user could just add a `direnv cache` directive into his `.envrc` to enable caching and then have `direnv reload` also invalidate the cache like you said.

Cache would be stored under `${XDG_CACHE_DIR:-$HOME/.cache}/direnv/$shasum.cache` and only contain the relevant diff to avoid the issue with dynamic variables like the `SSH_AUTH_SOCK`.

There are two drawbacks to this approach which is why the cache wouldn't simply be enabled for everyone:
[...]

Issue #98, opened 2014-01-30
.ruby-version


nyarly:
The emerging standard for choosing a ruby-version is to put a version number into a file called ".ruby-version" - I think that something like use ruby could work if I could source that file there (e.g. `use ruby .ruby_version` or `use ruby $(cat .ruby_version)` or something. Is that doable?

Issue #77, opened 2013-11-20
Support for tracking multiple file's mtime


zimbatm:
direnv currently only tracks the current `.envrc` for modification. For example if you use the `dotenv` or source another file, the user has to type `direnv reload` when those files are changed.

Instead of creating tons of DIRENV_ environment variables we could use a single one in which the whole state is serialized + base64 encoded.

zimbatm:
that's implemented actually, thanks @nyarly :)

PR #206, opened 2016-03-20
Expanding mtime to file watches

nyarly:
This is fairly big PR, but it was hard to limit the scope - it's kind of a fundamental change but a really nice one. It would be rad if this got merged as is, but I think that's beyond expectation. If there's changes needed, it'd be nice to hear as soon as possible so that it's still fresh in the meat caches.

I've swapped out the mtime in the RC struct for a FileTimes list that records the path and mtime for mutliple files. The watch structure is serialized to DIRENV_WATCHES the same way DIRENV_DIFF is.

The upshot is that direnv can now watch more than just the .envrc and its allow file for changes. This addresses #173, #98 as well as #77.

## Summary of Changes

Obviously, there's the FileTime[s] structs and associated methods. There were a lot of changes to `direnv export` and `direnv edit` to handle the change is how things were tracked.

### New Private Commands
`direnv watch <path>` adds the path to the watchlist and prints the new export command for the DIRENV_WATCHES env var. The upshot is that stdlib and direnvrc functions can `eval $(direnv watch filename)` in order to add a file to the list of watches. (export captures the value at the end of loading the .envrc so that changes during its execution can be picked up)

`direnv current <path>` checks to see if <path> is current or not - it exits 0 if the path is current or 1 if it's unknown, stale, missing or newly added. Almost useless outside of .envrc (or .direnvrc), but could be very useful for writing a caching stdlib function.

## Mea Culpas
A few changes not strictly in scope:

The behavior of the watch stuff was tricky to debug, so I wound up extending direnv's debugging - log.go now has a log_debug function that is a no-op unless `DIRENV_DEBUG=1` is set. It makes use of the log package instead of Fprintf'ing to Stderr, which is nice because it makes it possible to track file and line number.

cmd_export.go got a heavy refactoring. I had a hard time reasoning about it in one giant method, which is why it got broken up like that.

Maybe the worst thing: I updated the version to 2.8 - I needed an easy way to make sure I was running the right version when I was taking it for a test run, but I don't think that's exactly kosher. Still, the change of interface requires at least that, I think?


zimbatm:
Love this. I will take a look at the code now.

env.go
```diff
@@ -25,6 +25,12 @@ func GetEnv() Env {
 	return env
 }
 
+func (env Env) Clean() {
```
zimbatm: Maybe rename to `CleanState()` or `CleanContext()`

stdlib.sh
```diff
@@ -153,7 +153,10 @@ source_env() {
   if ! [[ -f $rcpath ]]; then
     rcpath=$rcpath/.envrc
   fi
+
   rcfile=$(user_rel_path "$rcpath")
+  watch_file $rcpath
```
zimbatm: the `"$rcpath"` should be quoted in case it contains some spaces

stdlib.sh
```diff
@@ -166,6 +169,18 @@ source_env() {
...
 
+# Usage: watch_file <filename>
+#
+# Adds <path> to the list of files that direnv will watch for changes - useful when the contents
+# of a file influence how variables are set - especially in direnvrc
+#
+watch_file() {
+  local file=${1/#\~/$HOME}
+
+  eval $($direnv watch $file)
```
zimbatm: quoted `"$file"`.

Shell escaping is tricky. I always use shellcheck on all of my scripts because of that.

stdlib.sh
```diff
@@ -486,7 +501,7 @@ use_nix() {
 
 ## Load the global ~/.direnvrc if present
 if [[ -f ${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc ]]; then
-  source_env "${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc" >&2
+  source_env ${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc >&2
 elif [[ -f $HOME/.direnvrc ]]; then
-  source_env "$HOME/.direnvrc" >&2
+  source_env $HOME/.direnvrc >&2
```
zimbatm: I'm not sure why the quotes were removed here. On OSX the home directly can be "/Users/Foo Bar" I think

version.go
```diff
@@ -1,3 +1,3 @@
 package main
 
-const VERSION = "2.7.0"
+const VERSION = "2.8.0"
```
zimbatm: I agree but I prefer to do the release separately. I like to bump the version and update the CHANGELOG.md at the same time (and then tag that commit).

zimbatm:
:tada: Got a few comments above ^ and Travis might have to be kicked around a bit. I suspect that we lost go 1.4 compatibility but that's not a bit deal. The docs also need updating a bit but I'm happy to do that independently.

Overall I really like the idea and the implementation is great. The export function is the hairiest and I like what you did with it. I also like the storage of mtimes and the overall cleanup that you did.

If you wanted to take things even further, all the state could be stored in a single `DIRENV_STATE` object that would contain the dir, diff and mtimes. That would open the door to adding other features like multiple switchable context per folder (for example to test a library with different versions of ruby).

nyarly:
Updated - ran shellcheck and ensured that all the checks are cleared, including setting some source= directives; has the advantage of hinting at shellcheck for other devs.

zimbatm:
Looking good. Anything more ?

nyarly:
I think that's it. I like the idea of a unified STATE but... One think at a
time, eh?

zimbatm:
Sure thing :)

I'm going to do some more testing, adding the docs and also make use of the new feature in `source_env`, `dotenv`, ... in the next couple of days. Once I'm happy that it's stable I'll make a release, probably next weekend.

PR #220, opened 2016-05-25
Permit empty NODE_VERSION_PREFIX variable

f440:
I use [n](https://github.com/tj/n) (It's very famous node.js manager). It install node.js to non-prefixed directory.

```
% ls n/versions/node
0.10.36  0.12.12  4.3.2  5.7.1  6.2.0
```

direnv can change node.js search prefix via the `NODE_VERSION_PREFIX` variable, but can't accept  empty string. The PR is fixed it.


zimbatm:
Fair enough, thanks

PR #242, opened 2016-12-10
Add `use guix` and update README to mention GNU Guix

cbaines:
I've been using both the direnv package from GNU Guix, and the `use guix` helper that @zimbatm wrote for me for many months now. Its been working really well, and I haven't heard about any issues, so please consider adding it to the stdlib :)


zimbatm:
Thanks, looking good! I'll make a release in a few.

PR #248, opened 2017-01-27
When MANPATH is empty, prepend to a default value in path_add

### LINKED ISSUE #247: Using load_prefix in environment with empty MANPATH breaks man
When `$MANPATH` is empty, `man` will use various configurations to construct an equivalent search path hierarchy. I think those configurations are platform dependent, but see, for example, [`man 5 manpath'](http://man7.org/linux/man-pages/man5/manpath.5.html).

That is, if `$MANPATH` is empty, `man` knows where to look for man pages. In contrast, If `$MANPATH` is not empty, `man` will only search in those paths.

If `$MANPATH` is empty, `load_prefix` (which calls `path_add MANPATH`), will set `$MANPATH` to a non-empty value, which will probably break `man`.

For example:
```
$ unset MANPATH
$ man man
# See man's man page
$ echo 'load_prefix foo' > .envrc
direnv: error .envrc is blocked. Run `direnv allow` to approve its content.
$ direnv allow
direnv: loading .envrc
direnv: export +CPATH +LD_LIBRARY_PATH +LIBRARY_PATH +MANPATH +PKG_CONFIG_PATH ~PATH
$ man man
No manual entry for man
```

Instead of looking in the "global" *and* "local" locations, man now only looks in the "local" locations, which is unexpected.

I'm not sure of a cross-platform way of fixing this, though `man -w` looks promising. Perhaps `path_add` could check if `$1` is `MANPATH`. See https://github.com/direnv/direnv/compare/master...mdawaffe:fix/MANPATH-prepending. I didn't make a PR because I was unsure of the approach.

mdawaffe:
If `$MANPATH` is empty, `man` still knows where to look.

If `$MANPATH` is not empty, `man` will only look in those paths.

Instead of setting `MANPATH=$dir` in `path_add`, prepend `$dir` to the default MANPATH-equivalent calculated with `man -w`.

For Darwin, see "SEARCH PATH FOR MANUAL PAGES" section of [`man man`](https://developer.apple.com/legacy/library/documentation/Darwin/Reference/ManPages/man1/man.1.html).

For Linux, see [`man 5 manpath`](http://man7.org/linux/man-pages/man5/manpath.5.html).

Fixes #247.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -257,7 +257,10 @@ path_add() {
 
   if [[ -z $old_paths ]]; then
     if [[ $1 = MANPATH ]]; then
-      # If empty, prepend $dir to the default
+      # If MANPATH is empty, man still knows where to look.
+      # If MANPATH is not empty, man will only look in MANPATH.
+      # So if we set MANPATH=$dir, man will only look in $dir.
+      # Instead, prepend to `man -w` (which outputs man's default paths).
       old_paths="$dir:$(man -w)"
     else
       old_paths="$dir"
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -256,7 +256,12 @@ path_add() {
   dir=$(expand_path "$2")
 
   if [[ -z $old_paths ]]; then
-    old_paths="$dir"
+    if [[ $1 = MANPATH ]]; then
+      # If empty, prepend $dir to the default
+      old_paths="$dir:$(man -w)"
+    else
+      old_paths="$dir"
+    fi
   else
     old_paths="$dir:$old_paths"
   fi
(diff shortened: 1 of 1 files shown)


zimbatm:
> man  uses  a  sophisticated  method of finding manual page files, based on the invocation options and environment variables, the /private/etc/man.conf configuration file, and some  built  in  conventions and heuristics.

-- man(1) Mac OS X Manual Page

This is gold.

zimbatm:
Thanks a lot for the PR @mdawaffe.

I split the function in two to avoid accumulating too much heuristic in path_add. I don't think someone would ever want a MANPATH with only the designated paths but it's possible.

mdawaffe:
Much clearer; thanks.

PR #255, opened 2017-02-22
Fix/252 node version sorting

### LINKED ISSUE #252: Partial NodeJS version does not return highest matching version
To reproduce:

```
$ cat .envrc
use node 7
$ find $NODE_VERSIONS -maxdepth 1 -mindepth 1
/Users/mjs/.local/share/node/versions/node-v6.9.1
/Users/mjs/.local/share/node/versions/node-v7.1.0
/Users/mjs/.local/share/node/versions/node-v7.2.0
/Users/mjs/.local/share/node/versions/node-v7.4.0
```

Actual:

`node-v7.1.0` is matched and used.

Expected:

`node-v7.4.0` (highest version of the 7 series of releases) is matched and used.

Discussion:

The code to [find the highest matching version](https://github.com/direnv/direnv/blob/master/stdlib.sh#L510) does not work as expected. There's a few problems:

- field separator is `.`, but this appears at points in the string in locations other than the version number.
- The `-r` option of sort appears to do nothing when combined with `-k`.

mdawaffe:
As #252 points out, there are two problems with the sorting of node versions.
1. The path to the node versions may contain a `.`, which is the separator used by `sort`.
2. `sort`'s `-r` and `-k` flags don't combine as expected.

This PR:
1. Strips the path prefix from the strings to sort, and
2. Fixes the sorting by moving the reversal flag to each individual key's options.

It also splits the multi-pipe command into several lines for readability, but I don't know how portable that is.

Fixes #252

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -525,6 +525,8 @@ use_node() {
     head -1
   )
 
+  node_prefix="$NODE_VERSIONS/${NODE_VERSION_PREFIX-"node-v"}$node_prefix"
+
   if [[ ! -d $node_prefix ]]; then
     log_error "Unable to find NodeJS version ($version) in ($NODE_VERSIONS)!"
     return 1
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -519,7 +519,7 @@ use_node() {
     while IFS= read -r line; do echo "${line#${NODE_VERSIONS%/}/${NODE_VERSION_PREFIX-"node-v"}}"; done |
 
     # Sort by version: split by "." then reverse numeric sort for each piece of the version string
-    sort -r -t . -k 1,1n -k 2,2n -k 3,3n |
+    sort -t . -k 1,1rn -k 2,2rn -k 3,3rn |
 
     # The first one is the highest
     head -1
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -510,7 +510,16 @@ use_node() {
   fi
 
   node_wanted=${NODE_VERSION_PREFIX-"node-v"}$version
-  node_prefix=$(find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" | sort -r -t . -k 1,1n -k 2,2n -k 3,3n | head -1)
+  node_prefix=$(
+    # Look for matching node versions in $NODE_VERSIONS path
+    find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" |
+
+    # Sort by version: split by "." then reverse numeric sort for each piece of the version string
+    sort -r -t . -k 1,1n -k 2,2n -k 3,3n |
+
+    # The first one is the highest
+    head -1
+  )
 
   if [[ ! -d $node_prefix ]]; then
     log_error "Unable to find NodeJS version ($version) in ($NODE_VERSIONS)!"
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -514,6 +514,10 @@ use_node() {
     # Look for matching node versions in $NODE_VERSIONS path
     find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" |
 
+    # Strip possible "/" suffix from $NODE_VERSIONS, then use that to
+    # Strip $NODE_VERSIONS/$NODE_VERSION_PREFIX prefix from line.
+    while IFS= read -r line; do echo "${line#${NODE_VERSIONS%/}/${NODE_VERSION_PREFIX-"node-v"}}"; done |
+
     # Sort by version: split by "." then reverse numeric sort for each piece of the version string
     sort -r -t . -k 1,1n -k 2,2n -k 3,3n |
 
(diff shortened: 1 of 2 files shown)


wilmoore:
@mdawaffe - the following list is what I get when I run `ls $NODE_VERSIONS` locally:

```
node-v0.10.36
node-v0.12.17
node-v0.12.18
node-v0.12.2
node-v0.12.7
node-v0.9.0
node-v4.2.2
node-v4.3.0
node-v4.3.2
node-v4.7.0
node-v5.9.0
node-v6.9.5
```

What's the ouptut from your patch given the above list (or similar)?

@zimbatm - what's the best way to write a simple test to validate the above?

mdawaffe:
Here's my setup:

```
$ env | grep NODE_VER
NODE_VERSION_PREFIX=
NODE_VERSIONS=/Users/mdawaffe/.nvm/versions/node

$ /bin/ls -1 "$NODE_VERSIONS"
v6.10.0
v6.9.0
v6.9.4
v7.2.0
v7.2.1
v7.3.0
v7.4.0
```

Current method:
```
$ node_wanted=v7; \
  find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" | \
  sort -r -t . -k 1,1n -k 2,2n -k 3,3n
/Users/mdawaffe/.nvm/versions/node/v7.2.1 # Should be 7.4.0
/Users/mdawaffe/.nvm/versions/node/v7.2.0
/Users/mdawaffe/.nvm/versions/node/v7.3.0
/Users/mdawaffe/.nvm/versions/node/v7.4.0
```

Method proposed in this PR:
```
$ node_wanted=v7; \
  find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" | \
  while IFS= read -r line; do echo "${line#${NODE_VERSIONS%/}/${NODE_VERSION_PREFIX-"node-v"}}"; done | \
  sort -t . -k 1,1rn -k 2,2rn -k 3,3rn
v7.4.0 # Success! Is 7.4.0
v7.3.0
v7.2.1
v7.2.0
```

Your setup (as I understand it):
```
$ env | grep NODE_VER
NODE_VERSIONS=/Users/mdawaffe/node-test

$ /bin/ls -1 "$NODE_VERSIONS"
node-v0.10.36
node-v0.12.17
node-v0.12.18
node-v0.12.2
node-v0.12.7
node-v0.9.0
node-v4.2.2
node-v4.3.0
node-v4.3.2
node-v4.7.0
node-v5.9.0
```

Current method:
```
$ node_wanted=node-v4; \
  find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" | \
  sort -r -t . -k 1,1n -k 2,2n -k 3,3n
/Users/mdawaffe/node-test/node-v4.2.2 # Should be 4.7.0
/Users/mdawaffe/node-test/node-v4.3.0
/Users/mdawaffe/node-test/node-v4.3.2
/Users/mdawaffe/node-test/node-v4.7.0
```

Method proposed in this PR:
```
$ node_wanted=node-v4; \
  find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" | \
  while IFS= read -r line; do echo "${line#${NODE_VERSIONS%/}/${NODE_VERSION_PREFIX-"node-v"}}"; done | \
  sort -t . -k 1,1rn -k 2,2rn -k 3,3rn
4.7.0 # Succes! Is 4.7.0
4.3.2
4.3.0
4.2.2
```

Ignore the fact that the current method returns paths and the method proposed in this PR returns version strings. That difference is normalized at https://github.com/direnv/direnv/pull/255/files#diff-5140bf46d14d26b672dfebb247c545aaR528 to return paths.

stdlib.sh
```diff
@@ -510,7 +510,22 @@ use_node() {
...
+  node_prefix=$(
+    # Look for matching node versions in $NODE_VERSIONS path
+    find "$NODE_VERSIONS" -maxdepth 1 -mindepth 1 -type d -name "$node_wanted*" |
+
+    # Strip possible "/" suffix from $NODE_VERSIONS, then use that to
+    # Strip $NODE_VERSIONS/$NODE_VERSION_PREFIX prefix from line.
+    while IFS= read -r line; do echo "${line#${NODE_VERSIONS%/}/${NODE_VERSION_PREFIX-"node-v"}}"; done |
+
+    # Sort by version: split by "." then reverse numeric sort for each piece of the version string
+    sort -t . -k 1,1rn -k 2,2rn -k 3,3rn |
```
zimbatm: what happens if node releases a version with a component that contains two digits? Like `6.10.0`

stdlib.sh
```diff
@@ -510,7 +510,22 @@ use_node() {
...
+
+    # The first one is the highest
+    head -1
+  )
+
+  node_prefix="$NODE_VERSIONS/${NODE_VERSION_PREFIX-"node-v"}$node_prefix"
```
zimbatm: can you extract `local node_version_prefix=${NODE_VERSION_PREFIX-"node-v"}`

stdlib.sh
zimbatm: Ok it's fine

zimbatm:
Thanks! Will make a new release in a few.

PR #256, opened 2017-02-22
Typo in MANPATH_add always generates "PATH missing" error

mdawaffe:
PR #248 fixed some MANPATH issues, but it also introduced some copy/paste typos.

https://github.com/direnv/direnv/blob/v2.11.1/stdlib.sh#L270-L284

```
Usage: MANPATH_add <path>
...
dir=$(expand_path "$2")
```

There is no `$2`.

This PR switches to the correct parameters in `MANPATH_add()`.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -279,8 +279,8 @@ path_add() {
 MANPATH_add() {
   local old_paths="${!1}"
   local dir
-  dir=$(expand_path "$2")
-  export "$1=$dir:${old_paths:-$(man -w)}"
+  dir=$(expand_path "$1")
+  export "MANPATH=$dir:${old_paths:-$(man -w)}"
 }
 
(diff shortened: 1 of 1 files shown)


stdlib.sh
```diff
@@ -279,8 +279,8 @@ path_add() {
 MANPATH_add() {
   local old_paths="${!1}"
```
zimbatm: old_paths also needs to be fixed to import the existing $MANPATH

zimbatm:
Sorry for the wait, I'm pushing a new release today with your fix. I made some adjustments and pushed your changes to master in 11db56325e6a35f9cbfaa5f3c336e9a7429d6a25

stdlib.sh
mdawaffe: Thanks for catching that.

PR #264, opened 2017-03-24
Only output color if $TERM is not dumb

robnagler:
Probably should use github.com/fatih/color
Broken, sorry, b/c I don't know how to integrate "go get" in Makefile
I'm not a go programmer. :(
See https://github.com/fatih/color/issues/50
Visit respectmyterm.com to learn why
Thanks!

### DIFF
--- a/log.go
+++ b/log.go
@@ -2,16 +2,20 @@ package main
 
 import (
 	"fmt"
+	"github.com/mattn/go-isatty"
 	"log"
+	"os"
 )
 
 const (
-	debugLogFormat   = "DBG-direnv: %s"
-	defaultLogFormat = "direnv: %s"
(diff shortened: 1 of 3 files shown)


zimbatm:
Thanks for your contribution!

Do you mind vendoring the `github.com/mattn/go-isatty` dependency? This is the first external dependency in this project so I haven't thought exactly how it should integrate.

robnagler:
Unfortunately, I'm not a go developer. I tried adding it as a submodule in `vendor/`, but the make is still failing even when adding `GO15VENDOREXPERIMENT=1`. I read a bit about this, and it seems you'll need packaging tool like `godep`.

zimbatm:
Alright, that works too ^_^

PR #290, opened 2017-08-18
stdlib: configurable direnv_layout_dir

zimbatm:
Introduce the `$direnv_layout_dir` variable to let the user select
out-of-folder layouts. The main purpose of making this configurable is
to allow the user to store the `.direnv` folder outside of the
project's directory. For example if it interracts badly with backups,
source control or the IDE.

The variable can be changed in the `.envrc` file or in the global
`~/.config/direnv/direnvrc` file. In that case on might want to hash the
project path like that:

```
: ${XDG_CACHE_HOME:=$HOME/.cache}
pwd_hash=$(echo -n $PWD | shasum | cut -d ' ' -f 1)
direnv_layout_dir=$XDG_CACHE_HOME/direnv/layouts/$pwd_hash
```

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -598,7 +598,7 @@ const STDLIB = "#!bash\n" +
 	"\n" +
 	"## Load the global ~/.direnvrc if present\n" +
 	"if [[ -f ${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc ]]; then\n" +
-	"  source_env \"${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc\" >&2\n" +
+	"  source \"${XDG_CONFIG_HOME:-$HOME/.config}/direnv/direnvrc\" >&2\n" +
 	"elif [[ -f $HOME/.direnvrc ]]; then\n" +
-	"  source_env \"$HOME/.direnvrc\" >&2\n" +
+	"  source \"$HOME/.direnvrc\" >&2\n" +
 	"fi\n"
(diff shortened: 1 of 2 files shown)


williamsmj:
I've only tested layout_python, but this works for me, i.e. if `/project/.envrc` says `layout_python_dir=/foo/bar` then the virtual environment that would normally go in `/project/.direnv/` goes in `/foo/bar`, and is activated and deactivated correctly.

However, there's a problem with the example code you give in this PR that sets a global `direnv_layout_dir` using a hash: at least on my system, whatever directory the `.envrc` is in `pwd_hash` is that of my home directory. My shell is not good enough to figure out what's going in with subshells. Perhaps this is an issue with the environment in which `pwd_hash` is being evaluated.

zimbatm:
Ok that's an issue with how the direnvrc was being loaed. Do you want to give it another go?

williamsmj:
Perfect! Works as expected. Thank you very much!

ymmv, but I modified `pwd_hash` slightly to make a listing of the layouts directory a bit more human-readable:
```
pwd_hash=$(basename $PWD)-$(echo -n $PWD | shasum | cut -d ' ' -f 1)
```

zimbatm:
thanks for testing out the change! time to merge

jtojnar:
Since this is what I see in search results for `direnv_layout_dir`, you can add the following to `.direnvrc` to avoid polluting the project directories.

```bash
direnv_layout_dir() {
	: ${XDG_CACHE_HOME:="$HOME/.cache"}
	pwd_hash="$(echo -n "$PWD" | shasum | cut -d ' ' -f 1)"
	mkdir -p "$XDG_CACHE_HOME/direnv/layouts/$pwd_hash"
	echo "$XDG_CACHE_HOME/direnv/layouts/$pwd_hash"
}
```

`mkdir -p` will not be necessary with https://github.com/nix-community/nix-direnv/commit/91f83adf52c515d4078aaccbdd42b5a4729fbcf6.

PR #295, opened 2017-09-15
pwd: Don't use -P to remove symlinks

purpleidea:
If you use -P you get the absolute "physical" path removing any
symlinks, however it is then used in a context where it is expecting a
relative path to find the .envrc file. As a result, it can often be
incorrect and cause `pushd` errors.

Fixes GH#294

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -147,7 +147,7 @@ const STDLIB = "#!bash\n" +
 	"#\n" +
 	"find_up() {\n" +
 	"  (\n" +
-	"    cd \"$(pwd -P 2>/dev/null)\"\n" +
+	"    cd \"$(pwd 2>/dev/null)\"\n" +
 	"    while true; do\n" +
 	"      if [[ -f $1 ]]; then\n" +
 	"        echo \"$PWD/$1\"\n" +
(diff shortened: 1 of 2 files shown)


purpleidea:
wtf? Somehow the patch you merged, was not the patch I sent. In particular my name in the commit changed. What happened?

mponton:
I'm checking this on my phone but I see the proper patch and your name in the commit. Some kind of issue in your browser session?

purpleidea:
Check the sha1 ID of my patch in my branch and what was committed.

mponton:
This is normal when the PR is merged no? Look at all the other PRs, they have their commit id, the merge produce another one. Sorry if I misunderstand your concern...

purpleidea:
It's obvious the patch was mangled. Look at the name field in the two ;)

zimbatm:
This is what happens when using the "Squash & Merge" button on github

purpleidea:
It changed my commit name.
That's a major GH bug as far as I'm concerned.
I'd recommend merging properly with the CLI for now.
This breaks a lot of things, for example, if I do `git branch -d
that-branch` my local git doesn't think it's merged!

PR #298, opened 2017-09-26
stdlib: make direnv_layout_dir lazy

### LINKED ISSUE #297: Layout within nested environments broken
I just updated to direnv v2.13.0 and I see un expected behavior. I have a tree of directories with nested environments that source the top level using `source_up`. In the highest level I have a custom Python virtualenv defined by using `layout python python2.7`. In the children I would have custom settings but I would not define a new layout for python.

Up to direnv v2.12.2 the way it worked is that if I moved into a child directory a Python virtualenv would be configured by the parent `.envrc` within a `.direnv` directory within the parent. All the children would share the same version of Python. 

When I upgraded to direnv v2.13.0, now every time I move into a child directory, a new version of the virtualenv is created within a `.direnv` directory in the children directories, instead of reusing the parent `.direnv`. 

Is this a bug or a change in behavior?

zimbatm:
This fixes an issue where the first .envrc fixes the $direnv_layout_dir
in combination with source_up or source_env.

Fixes #297

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -9,7 +9,17 @@ const STDLIB = "#!bash\n" +
 	"\n" +
 	"# Config, change in the direnvrc\n" +
 	"DIRENV_LOG_FORMAT=\"${DIRENV_LOG_FORMAT-direnv: %%s}\"\n" +
-	"direnv_layout_dir=$PWD/.direnv\n" +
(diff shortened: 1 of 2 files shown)


elventear:
Works for me 👍

PR #311, opened 2017-11-28
Add option to python to specify virtualenv

peterfpeterson:
A reasonable way to use `virtualenv` is to enable the system site packages. This only works if you are selecting the `python` that the version of `virtualenv` was built with. To use, set your `.envrc` to be
```shell
layout python3 --system-site-packages
```
This still works with the traditional methods, but matching the two tools gives more flexibility.


zimbatm:
Hi, thanks for your PR.

Do you mind extending your explanation a bit? As I am not a python developer there are a few things that I don't know.

Do you mean that there is a virtualenv-3 that should only be used with python3? In what cases would virtualenv-3 be installed, is it something distro-specific?

I am also note sure what `--system-site-packages` does. Is it a way to import some packages from the system and avoid unnecessary rebuilds?

Thanks for your patience :)

peterfpeterson:
On many linux systems, you can install multiple versions of python and associated tools side-by-side with lots of symlinks between them. For example, on fedora 26:
* python -> python2 -> python2.7 -> python2.7.4
* python3 -> python3.6 -> python3.6.3

Similarly, `virtualenv` has lots of versions with links. The [`--system-site-packages`](https://virtualenv.pypa.io/en/stable/userguide/#the-system-site-packages-option) option adds a directory tree in your virtual environment to install extra things into (via `pip`), but will still fall back on the site-packages for the python that virtualenv was built with. What it boils down to, is that `--system-site-packages` should only be used to create a python environment for the python version that the particular virtualenv runs in.

The way that I wrote `layout_python2` and `layout_python3` will fall back to system defaults for virtualenv if the python version specific version isn't found. I think this change really only affects people running in the system/virtualenv hybrid mode which is only available using this one flag. To be more explicit: I install numpy with a system package, and with this flag I don't need to reinstall it in the virtual environment.

I'd add this to the documentation, but it is a little unclear if it really is duplicated in a bunch of places, or it just seems like that.

MemoryReload:
@peterfpeterson Does `layout python` command support this option?

peterfpeterson:
@MemoryReload This is a change to `layout python`. I don't quite follow what you are asking.

@zimbatm The usage of this (via `layout python2` or `layout python3` is to automatically detect the partner virtualenv command. I made the assumption that nobody would specify the virtualenv command without specifying the python command. You're pointing out that it is a bad assumption. The current code assumes that the python virtual environment command is `virtualenv` as well. I'll change it to supply the virtual environment command via a flagged option, or looking falling through a list of virtualenv then venv.

MemoryReload:
@peterfpeterson Since as the wiki said, `layout python2` is a hand short method for `layout python 2.x.x `(x here may be you precise version number), I just want to know whether this --system-site-packages option is support or not for `layout python 2.x.x `. Thanks!

peterfpeterson:
[`--system-site-packages`](https://virtualenv.pypa.io/en/stable/userguide/#the-system-site-packages-option) only works if the version of python you want in your environment is the same version as the one at the system level. I normally create two environments, one with `python2` and one with `python3`. Fedora currently has side-by-side installs of many packages so it works out by changing which `virtualenv` command you use.

MemoryReload:
@peterfpeterson   Thanks for your explanation. 👍

peterfpeterson:
I'm suspicious that there is a better way to do this with a `for` and `break`.

zimbatm:
There is too much heuristics happening here and I really dislike heuristic-ridden code. Every code branch is another way for the code to rot and fail, it makes it harder to debug, more bug-fix releases and then I am usually the one trying to support users in the issue tracker. If python doesn't have a simple mapping between virtualenv and python version then maybe I should just remove `layout_python` from the stdlib.

Instead of trying to handle all of the cases I suggest to: (a) pick a sane default if any exists, and (b) let the user pass the python and virtualenv versions as arguments if he has a weird setup.

Eg: `PYTHON_BIN=my/bin/python VIRTUALENV_BIN=my/bin/virtualenv layout python`

peterfpeterson:
I agree totally with simplicity being best. I didn't even think of injecting values in `.envrc`. Since the python version is an optional argument already I only added injecting `VIRTUALENV_BIN`. There are also a few more checks to give the user better error messages when something goes wrong.

peterfpeterson:
I'm closing this because #376 will solve the issue better.

PR #312, opened 2017-11-29
Add anaconda support

peterfpeterson:
[anaconda](https://www.anaconda.com/) is a popular virtual environment for python. This adds a layout for it.

Normal conda usage would require a user to
```shell
conda env create # defaults to environment.yml file
source activate <envname>
```


zimbatm:
Hi,

sounds good. Is it possible to have the conda env live under the `direnv_layout_dir()`?

The main issue with the current approach is that two projects can have the same name. It happens to me quite a lot since I'm checking out a lot of github projects. That's why all the other layouts try to store their state in the `.direnv` folder under the project's root.

peterfpeterson:
Unfortunately, the environments are universal to the combination of installation and user of conda. They have [docs here](https://conda.io/docs/user-guide/tasks/manage-environments.html). virtualenv sandboxes much better.

zimbatm:
alright. potentially we could use something like `pwd -P | sha1sum` to create a unique name. but if the user selects the name by hand it works for me too.

zimbatm:
pushed a few tweaks, let me know if you agree with them

peterfpeterson:
I just tried this out and it works as it did before. Merge when you are ready.

zimbatm:
thanks!

PR #314, opened 2017-11-30
Add support for Pipenv layout

nvie:
This adds a new layout, similar to `layout_python`, but uses Pipenv to build the virtualenv from the Pipfile located in the same directory.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -440,7 +440,7 @@ layout_pipenv() {
 
   local VENV=$(pipenv --bare --venv 2>/dev/null)
   if [[ -z $VENV || ! -d $VENV ]]; then
-    pipenv install
+    pipenv install --dev
   fi
 
   export VIRTUAL_ENV=$(pipenv --venv)
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -434,7 +434,7 @@ layout_python3() {
 #
 layout_pipenv() {
   if [[ ! -f Pipfile ]]; then
-    echo 'No Pipfile found.  Use `pipenv` to create a Pipfile first.' >&2
+    log_error 'No Pipfile found.  Use `pipenv` to create a Pipfile first.'
     exit 2
   fi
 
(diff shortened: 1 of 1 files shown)


stdlib.sh
```diff
@@ -427,6 +427,26 @@ layout_python3() {
...
 }
 
+# Usage: layout pipenv
+#
+# Similar to layout_python, but uses Pipenv to build a
+# virtualenv from the Pipfile located in the same directory.
+#
+layout_pipenv() {
+  if [[ ! -f Pipfile ]]; then
+    echo 'No Pipfile found.  Use `pipenv` to create a Pipfile first.' >&2
```
zimbatm: is there a `pipenv init` that could make sense here?

nvie: That command does not exist. I'm not sure if invoking that would work here, since it would have to decide between Python 2 or Python 3 on the user's behalf.

Do you think dropping that message and then returning with a 0 exit code is better? I noticed that direnv keeps throwing the error after every command you type if you're stuck in this state.

zimbatm: No that's good, I just wanted to check if there is anything helpful that we could do instead of returning an error. Since it's not the case the error is good.

Actually you should use `log_error` to respect the direnv log format.

The error after every command is a but that I meant to fix. It should only fail once and stay in the errored state.

zimbatm: Pushed the change for the error logging in 4748fbe

nvie: Oh, much nicer — TIL about `log_error`.

nvie: 👍

nvie:
I also added f79c226edb81e05f69d776390173339fb3e7aa29, which should make the environment completely compatible with `pipenv shell` itself.

zimbatm:
Ok great, looks like it's ready to be merged. Any last words? :)

nvie:
Nope! Let's have it, cheers 🍻 !

nvie:
@zimbatm Thanks for the quick merge. Would you be able to push out a new release that includes this?

zimbatm:
There are few more python-related PRs in the pipeline. I will do a "python" release once they have all been processed.

PR #324, opened 2018-01-05
use_nix: unset IN_NIX_SHELL

ttuegel:
nix-shell sets the environment variable IN_NIX_SHELL when evaluation happens
inside a shell. direnv picks up this variable and sets it outside
nix-shell. Subsequent evaluation can fail if IN_NIX_SHELL is set but we are not,
in fact, in nix-shell. Therefore, use_nix should restore the original value of
IN_NIX_SHELL after loading the environment.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -635,11 +635,13 @@ use_node() {
 # (e.g `use nix -p ocaml`).
 #
 use_nix() {
+  local orig_IN_NIX_SHELL="$IN_NIX_SHELL"
   direnv_load nix-shell --show-trace "$@" --run 'direnv dump'
   if [[ $# = 0 ]]; then
     watch_file default.nix
     watch_file shell.nix
   fi
+  IN_NIX_SHELL="$orig_IN_NIX_SHELL"
 }
 
 # Usage: use_guix [...]


zimbatm:
I don't understand what problem it's trying to solve exactly. Since it's re-exporting all the variables from nix-shell, isn't it essentially like it's in the nix shell?

ttuegel:
The main use of `IN_NIX_SHELL` is for files that change their behavior depending on whether they are loaded by `nix-build` or `nix-shell`. If direnv sets `IN_NIX_SHELL`, then `nix-build` fails for such files.

zimbatm:
sorry I don't get it, shouldn't nix-build filter the environments from the cli? is nix installed in single or multi-user mode?

ttuegel:
> sorry I don't get it, shouldn't nix-build filter the environments from the cli?

The __build__ is isolated from the environment, but the __evaluation__ is NOT because Nix has a `getEnv` builtin. And the evaluation determines what goes into the build.

zimbatm:
Got it :)

I just tweaked the commit a bit and pushed it as 99e43f697ca10be5ec28b0185f9cfd9c456dc527 in master.

ttuegel:
Thanks!

Issue #180, opened 2015-09-08
project whitelist DIRENV_ALLOW


JeanMertz:
I found an [issue](https://github.com/direnv/direnv/issues/23) discussing a `DIRENV_ALLOW` whitelist path to allow all `.envrc` files inside that path.

It doesn't look like this was ever implemented. We're using Direnv heavily in our project, with each directory having a different set of env vars being set. However, having to allow each file every time a developer makes a change is becoming cumbersome.

In general, we trust the developers working on this project, so it would make sense to whitelist all direnv files in this repository.

Is there a way to do this? Or are there some ideas to add this feaure?

JeanMertz:
Also: symlinking to an existing (and allowed) direnv still wants you to allow the symlink, even though its contents are already alowed somewhere else. Is this a bug?

zimbatm:
Hi @JeanMertz ,

I'm not against adding a `DIRENV_ALLOW` setting, just never took the time to implement it. Do you want to give it a shot ? The auth logic is contained in the `rc.go` file.

Since that ticket I've been convinced to also make the path as part of the hash so two identical files can have different hashes. The reason is that an innocuous-looking file can be devastating in a different context. I know it's a bit paranoid but it's also good for security.

zimbatm:
Fixed in master!

JeanMertz:
Wow, sorry for not commenting on this in the last couple of years 🙈 nice to see this being implemented, awesome 👍

zimbatm:
no worries. I just released https://github.com/direnv/direnv/releases/tag/v2.15.0 that contains this feature

ssbarnea:
@zimbatm Apparently this feature does not work on version 2.16.0 on Linux, even after setting this to `.bash_profile`:

```

export DIRENV_ALLOW=$HOME
eval "$(direnv hook bash)"
```

I checked and the environment variable is set but when I cd to a a subfolder which contains the .envrc, I still get:
```
direnv: error .envrc is blocked. Run `direnv allow` to approve its content.
```
Running manually `direnv allow` would be useless as these .envrc files are generate during build and the who point was to auto load them when the user is switching to the build folder.

Still, I used the same feature on MacOS without problems. 

Also this feature is too important to be missed from mentioned when doing "direnv help".

LaurensRietveld:
Correction to post above, path should be `$HOME/.config/direnv/config.toml` (this bit me for a while ;) )

zimbatm:
thanks, updated the example

[...]

PR #328, opened 2018-02-02
Allow whitelisting specific .envrc paths.

anp:
This is intended to cover the case where you may have frequently changing .envrc files whose source you trust (for example, a version-controlled envrc in a team environment). The whitelisting can be revoked, at which point the normal allow/deny operations resume effect.

This has been previously requested in https://github.com/direnv/direnv/issues/180. To answer questions on that issue, on my team we do commit .envrc files to automatically set developers up with project-specific configuration and local PATH overrides. It would be really nice to provide devs with a script that will whitelist all of the .envrc paths within our repo, so they don't need to allow, cd out, and cd back in every time they pull or rebase and a file has changed.

One thing to note is that I've chosen to whitelist the absolute path of the file for less ambiguity, although I'm honestly not sure whether that's more or less surprising behavior than whitelisting the path the user provides, symlinks or no.

I've tested this locally on my machine, but I honestly don't quite understand how the tests in this repository are set up so I haven't added automated ones. The logic is pretty simple, but if you'd like it to come with test coverage I can poke around some more.

Happy to discuss if this approach doesn't work for you!

### DIFF
--- a/cmd_allow.go
+++ b/cmd_allow.go
@@ -25,7 +25,7 @@ var CmdAllow = &Cmd{
 			return
 		}
 
-		rc := FindRC(rcPath, config.AllowDir())
+		rc := FindRC(rcPath, config.AllowDir(), config.WhitelistDir())
 		if rc == nil {
 			return fmt.Errorf(".envrc file not found")
 		}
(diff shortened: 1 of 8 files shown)


zimbatm:
I guess this can be closed now. Implemented in #332

PR #332, opened 2018-02-18
Add a TOML configuration file, support whitelisting envrc's regardless of their content's hash

anp:
With this PR, direnv now supports a configuration file at `CONFIG_DIR/config.toml`, and adds support for a `whitelist` key. Any `.envrc` files which match the patterns are allowed, regardless of whether the specific path & contents of the envrc file have been explicitly `allow`ed.

Here's an example TOML file:

```toml
[whitelist]
prefix = ["/Users/adam/code/foo"]
exact = [
    "/Users/adam/code/bar",
    "/Users/adam/code/quux/.envrc"
]
```

In this example, any envrc underneath the `foo` directory (including subdirectories) is trusted, even if its contents have changed. The topmost envrc in the `quux` directory is trusted as well, but not those of subdirectories.

Addresses https://github.com/direnv/direnv/issues/180
Supersedes https://github.com/direnv/direnv/pull/328

### DIFF
--- a/Gopkg.lock
+++ b/Gopkg.lock
@@ -2,14 +2,20 @@
 
 
 [[projects]]
+  name = "github.com/BurntSushi/toml"
+  packages = ["."]
+  revision = "b26d9c308763d68093482582cea63d69be07a0f0"
+  version = "v0.3.0"
+
+[[projects]]
   branch = "master"
   name = "github.com/direnv/go-dotenv"
   packages = ["."]
-  revision = "ed7765a41787829cf0a04dd562883d9d4e5cd8a6"
+  revision = "4738ec4932d7ea82d158b6900f0c7a12f116e5e5"
 
 [solve-meta]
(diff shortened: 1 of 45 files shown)


anp:
I'll take a look at these failing tests soon.

zimbatm:
It looks like it's failing the test if the config.toml is missing. It should just ignore that step if it's the case.

anp:
Agreed! There's one edge case that I think it should handle, which is warning the user (over stderr?) if the config file exists but can't be parsed. Going to add that now, let me know if you'd like to handle that case differently.

zimbatm:
yes good thinking!

anp:
Updated, my original intent was to not error in this circumstance, but my unfamiliarity with Go meant I was accidentally assigning a value to the `err` value for the whole functions! Fixed and tests are passing for me locally. Will wait for CI.

PR #338, opened 2018-02-24
Fix build failure with golang-1.10

punitagrawal:
direnv tests fails when built from sources with golang-1.10. This was
reported in debian bug #890852.

The failure seems to be caused by incorrect escaping of '%' in 'direnv
stdlib'. Fix this by correcting the escape sequence.

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -599,7 +599,7 @@ const STDLIB = "#!bash\n" +
 	"\n" +
 	"    # Strip possible \"/\" suffix from $NODE_VERSIONS, then use that to\n" +
 	"    # Strip $NODE_VERSIONS/$NODE_VERSION_PREFIX prefix from line.\n" +
-	"    while IFS= read -r line; do echo \"${line#${NODE_VERSIONS%/}/${node_version_prefix}}\"; done |\n" +
+	"    while IFS= read -r line; do echo \"${line#${NODE_VERSIONS%%/}/${node_version_prefix}}\"; done |\n" +
 	"\n" +
 	"    # Sort by version: split by \".\" then reverse numeric sort for each piece of the version string\n" +
 	"    sort -t . -k 1,1rn -k 2,2rn -k 3,3rn |\n" +


zimbatm:
thanks @punitagrawal . I extended your work a little bit in #339. By not using printf at all it will avoid awkward escaping situations.

punitagrawal:
Thanks @zimbatm for picking up the fix and improving on it. As you can
probably tell, golang doesn't come naturally to me. :)

I'll pick it up for the 2.15 update of the package.

zimbatm:
Let me do a 2.15.1 release so you don't have to keep patches in debian

zimbatm:
https://github.com/direnv/direnv/releases/tag/v2.15.1

zimbatm:
@punitagrawal do you think it would make sense to move your repo to direnv/direnv-deb maybe?

punitagrawal:
On Sat, Feb 24, 2018 at 7:45 PM, zimbatm <notifications@github.com> wrote:

> https://github.com/direnv/direnv/releases/tag/v2.15.1
>

Ah thanks I didn't see your response until after I uploaded 2.15.0. There
were a couple of lintian warnings I addressed in the upload as well.

I don't mind carrying a patch - especially if it's been merged upstream. I
will upload an update in the next couple of days.

zimbatm:
I think it would help with discoverability. It's hard to predict in what ways users are ready to contribute so my strategy is to lay everything in front of them. It's nice to wake up to a fresh PR :)

The story around including the deb/ folder in upstream is not super clear. I don't mind adding it but my understanding is that you would still have to maintain your own branches and tags.

punitagrawal:
On Sun, Feb 25, 2018 at 12:23 PM, zimbatm <notifications@github.com> wrote:

> I think it would help with discoverability. It's hard to predict in what
> ways users are ready to contribute so my strategy is to lay everything in
> front of them. It's nice to wake up to a fresh PR :)
>

Ack! No arguments there.


> The story around including the deb/ folder in upstream is not super clear.
> I don't mind adding it but my understanding is that you would still have to
> maintain your own branches and tags.
>

After having looked into this a bit more I think it makes sense when the
upstream authors are also maintaining the package - otherwise it isn't much
gain.


On moving the package repo - please go ahead and create the repo. It can
start out being a clone of the current package repo if it's easy or I can
repopulate if needed. The repo location can be changed with the next update
of the package.

Thanks.

zimbatm:
I had to create a new repo, github doesn't support hosting two forks in the same org apparently (even if I did the rename before)

punitagrawal:
On Mon, Feb 26, 2018 at 6:12 PM, zimbatm <notifications@github.com> wrote:

> I had to create a new repo, github doesn't support hosting two forks in
> the same org apparently (even if I did the rename before)
>

Looks to be all working from my end. I've pushed the package repo branches
to start with.

zimbatm:
Great! Feel free to ping me if there is anything else

PR #347, opened 2018-03-20
Add the usage of source_up to the README

ktaragorn:
This is far more useful than source_env (so much that possibly it should be default)


zimbatm:
thanks!

PR #353, opened 2018-04-06
fixed instructions to compile from source

gherlein:
if you follow the existing directions the make fails - these steps work properly

### DIFF
--- a/README.md
+++ b/README.md
@@ -46,6 +46,7 @@ Dependencies: make, golang
 ```bash
 go get github.com/direnv/direnv
 cd $GOPATH/src/github.com/direnv/direnv
+go get ./...
 make
 sudo make install
 # or symlink ./direnv into the $PATH
--- a/README.md
+++ b/README.md
@@ -44,9 +44,10 @@ nope
 Dependencies: make, golang
 
 ```bash
-git clone https://github.com/direnv/direnv
-cd direnv
-make install
+go get github.com/direnv/direnv
+cd $GOPATH/src/github.com/direnv/direnv
+make
+sudo make install
 # or symlink ./direnv into the $PATH
 ```
 


README.md
```diff
@@ -44,9 +44,11 @@ nope
 Dependencies: make, golang
 
 ```bash
-git clone https://github.com/direnv/direnv
-cd direnv
-make install
+go get github.com/direnv/direnv
+cd $GOPATH/src/github.com/direnv/direnv
+go get ./...
```
zimbatm: this shouldn't be necessary since everything is in the vendor/ folder. what error did you get?

gherlein:
If you clone the repo outside $GOPATH you get:

gherlein@jupiter:~/src/direnv$ make
go fmt
go build  -o direnv
cmd_dotenv.go:7:2: cannot find package "github.com/direnv/go-dotenv" in any of:
	/usr/lib/go-1.10/src/github.com/direnv/go-dotenv (from $GOROOT)
	/home/gherlein/go/src/github.com/direnv/go-dotenv (from $GOPATH)
Makefile:38: recipe for target 'direnv' failed
make: *** [direnv] Error 1

The steps I provided ensure you build inside $GOPATH.  That was the error I was talking about

I agree that `go get ./...` is not needed.

README.md
ssbarnea: If someone clones the repository and tried to build, they will fail. Same happens if they try to build directly using go build or if they try to use make to build.

This is because go language has some requirements regarding the location and layout of the source code.

Ideally the Makefile should take care of this and allow user to build regardless of the location where the direnv source code was cloned. Sadly it does no do this at this moment.

zimbatm:
thanks everyone for your feedback. I spent a bit of time and reworked the makefile so direnv doesn't need to be checked out in the GOPATH anymore in fa4fef76aada582ce2aadc2788239e695581bbc7

Issue #303, opened 2017-10-25
Moving directly into a subdirectory that is a symlink causes direnv not to load


peterhoeg:
Let's say I'm currently in ```$HOME``` and inside ```$HOME/foo``` is ```bar``` and ```baz```.

```bar``` is a normal directory.
```baz``` is a symlink to another directory outside ```$HOME/foo```.

If I do ```cd foo/bar``` everything is OK.

If I do ```cd foo/baz``` I get:
```
environment: line 187: pushd: ../.envrc: No such file or directory
direnv: referenced ../.envrc/.envrc does not exist
environment: line 196: popd: directory stack empty
```

If I do ```cd foo; cd baz``` everything is OK too.

**Comment by zimbatm** [comment]
Hi Peter,

what version of direnv are you currently using? There was a fix released in 1.13.0 for a similar issue. Can you just try upgrading to the latest version and try again if it's not already the case?

**Comment by peterhoeg** [comment]
```
peter@dolores:~ $ direnv --help
direnv v2.13.1
```

Following ```nixos-unstable```.

**Comment by zimbatm** [comment]
ok, probably some of the bash builtins is resolving to the absolute path... Can you ping me again next week? I don't have the time to do a proper investigation right now.

**Comment by peterhoeg** [comment]
Sure - this is hardly a high priority item considering ```cd foo; cd baz``` works. Thanks for your hard work

**Comment by peterhoeg** [comment]
ping @zimbatm

**Comment by zimbatm** [comment]
sorry it's on my backlog but most likely won't have the energy to investigate this

**Comment by etcusrvar** [comment]
Not a fully tested proper fix, but this does fix the issue for me (on Linux)-- tests if current dir is a symlink.
```
diff --git a/stdlib.sh b/stdlib.sh
index b64c65f..842d537 100644
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -178,7 +178,11 @@ source_env() {
   local rcpath=${1/#\~/$HOME}
   local rcfile
   if ! [[ -f $rcpath ]]; then
-    rcpath=$rcpath/.envrc
+    if [[ -L "$(pwd)" ]]; then
+      rcpath="$(pwd)/$rcpath"
+    else
+      rcpath=$rcpath/.envrc
+    fi
   fi
```

etcusrvar:
Created PR: https://github.com/direnv/direnv/pull/354

etcusrvar:
[...]

PR #354, opened 2018-04-17
Issue #303- check if dir is a symlink

etcusrvar:
https://github.com/direnv/direnv/issues/303

Enabled and fixed the test case that was added and commented out 5 years ago.

Tested on Linux and macOS. (I ran the .sh test script, and made the same changes to the .tcsh script but didn't run it.)

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -180,7 +180,11 @@ const STDLIB = "#!bash\n" +
 	"  local rcpath=${1/#\\~/$HOME}\n" +
 	"  local rcfile\n" +
 	"  if ! [[ -f $rcpath ]]; then\n" +
-	"    rcpath=$rcpath/.envrc\n" +
+	"    if [[ -L \"$(pwd)\" ]]; then\n" +
+	"      rcpath=\"$(cd \"$(dirname \"$(pwd)/$rcpath\")\" && pwd)/.envrc\"\n" +
+	"    else\n" +
+	"      rcpath=$rcpath/.envrc\n" +
+	"    fi\n" +
 	"  fi\n" +
 	"\n" +
 	"  rcfile=$(user_rel_path \"$rcpath\")\n" +
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -178,7 +178,11 @@ source_env() {
   local rcpath=${1/#\~/$HOME}
   local rcfile
   if ! [[ -f $rcpath ]]; then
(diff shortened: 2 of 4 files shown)


zimbatm:
thanks! I am glad that you took the time fixing this issue.

Do you know if it would work if the parent folder was a symlink too? I am sure that your patch fixes the common case but am suspecting that it wouldn't work with deeper symlink structures.

etcusrvar:
You are correct-- I verified that this fix as is will only work if `$(pwd)` itself is a symlink. It won't work if there is a symlink somewhere down the path. For that to work, there would have to be something like a `realpath` or `readlink -f` check. Something like this:

      if [[ "$(pwd)" != "$(readlink -f "$(pwd)")" ]]; then
        rcpath="$(cd "$(dirname "$(pwd)/$rcpath")" && pwd)"
      fi

I have this working locally on both Linux and macOS and can add to this PR if you think that using `realpath` or `readlink` is an acceptable solution. There are compatibility/portability issues with using either of these commands, and I don't know how important that is to the project.

zimbatm:
direnv has the internal expand_path command for cross-platform purposes.

In regards to the symlink, I don't think that we should ever resolve them. The goal of the symlinks is to teleport some of the filesystem onto another part of the tree. When resolving the symlinks it's not respecting that.

hexmode:
What needs to happen to get this merged? I'm running into this problem.

blalor:
Just ran into this, as well; would love to see it merged!

zimbatm:
I would love to get this fixed as well but fear that it's opening a rabbit hole. There are many ways that folders can be symlinked together. This is fixing one scenario and I don't see a solution that solves all the potential combinations.

zimbatm:
It's mainly of my own doing, direnv doesn't have enough code coverage that I am able to merge changes with confidence. Since I am the one answering to all issues and also fixing most of the bugs, I can already see how this change will have ripple effects. I am not looking forward to debug the user's convoluted symlink setups ^_^

hexmode:
>... direnv doesn't have enough code coverage that I am able to merge changes with confidence.

Would it be fair to say, then, that improving code coverage is a prerequisite for this?  I just opened #409 for that.

zimbatm:
thanks @hexmode. I really appreciate every contributions but have limited time. It would be nice to have a second core contributor joining the projects but power law distribution and all.

etcusrvar:
stale

Issue #356, opened 2018-04-20
add support for elvish


sbinet:
it would be great to add support for `elvish`:
- https://elvish.io/

a fish-like shell, written in Go.

zimbatm:
Does elvish have any hooks that direnv can use to execute before the prompt command?

Adding new output formats is fairly straight-forward. Eg: https://github.com/direnv/direnv/blob/master/shell_tcsh.go

zimbatm:
possibly the `edit:before-readline` hook would work. TBC

sbinet:
Let me see if I can come up with a PR...

PR #357, opened 2018-04-22
direnv: add support for elvish

sbinet:
Fixes direnv/direnv#356.

### DIFF
--- a/shell_elvish.go
+++ b/shell_elvish.go
@@ -1,0 +1,42 @@
+package main
+
+import (
+	"bytes"
+	"encoding/json"
+)
+
+type elvish struct{}
+
+var ELVISH = elvish{}
+
+func (elvish) Hook() (string, error) {
+	return `## hook for direnv
+@edit:before-readline = $@edit:before-readline {
+	try {
+		m = (direnv export elvish | from-json)
+		keys $m | each [k]{
+			if (==s $k 'null') {
+				unset-env $k
+			} else {
+				set-env $k $m[$k]
+			}
+		}
+	} except e {
+		nop
+	}
+}
+`, nil
+}
+
+func (sh elvish) Export(e ShellExport) string {
+	buf := new(bytes.Buffer)
+	err := json.NewEncoder(buf).Encode(e)
+	if err != nil {
+		panic(err)
+	}
+	return buf.String()
+}
+
+var (
+	_ Shell = (*elvish)(nil)
+)
--- a/Makefile
+++ b/Makefile
@@ -50,6 +50,8 @@ man: $(ROFFS)
 test: build
 	go test
 	./test/direnv-test.sh
+	go get github.com/elves/elvish
+	elvish ./test/direnv-test.elv
 
 install: all
 	install -d $(DESTDIR)/bin
--- a/README.md
+++ b/README.md
@@ -115,6 +115,19 @@ Add the following line at the end of the `~/.cshrc` file:
 eval `direnv hook tcsh`
 ```
 
+### Elvish
+
+Run:
+
+```
+$> direnv hook elvish > ~/.elvish/lib/direnv.elv
+```
+
+and add the following line to your `~/.elvish/rc.elv` file:
+
+```
+use direnv
+```
 
 ## Usage
 
--- a/README.md
+++ b/README.md
@@ -2,7 +2,7 @@ direnv -- Unclutter your .profile
 =================================
 
 `direnv` is an environment switcher for the shell. It knows how to hook into
-bash, zsh, tcsh and fish shell to load or unload environment variables
+bash, zsh, tcsh, fish shell and elvish to load or unload environment variables
 depending on the current directory. This allows project-specific
 environment variables without cluttering the `~/.profile` file.
 
--- a/man/direnv.1
+++ b/man/direnv.1
@@ -57,38 +57,81 @@ has it's own extension mechanism:
 
 .SS BASH
 .PP
-Add the following line at the end of your "\~/.bashrc" file:
+Add the following line at the end of the \fB\fC\~/.bashrc\fR file:
 
 .PP
-\fB\fCeval "$(direnv hook bash)"\fR
+.RS
+
+.nf
+eval "$(direnv hook bash)"
+
+.fi
+.RE
 
 .PP
 Make sure it appears even after rvm, git\-prompt and other shell extensions
-that manipulate your prompt.
+that manipulate the prompt.
 
 .SS ZSH
 .PP
-Add the following line at the end of your "\~/.zshrc" file:
+Add the following line at the end of the \fB\fC\~/.zshrc\fR file:
 
 .PP
-\fB\fCeval "$(direnv hook zsh)"\fR
+.RS
+
+.nf
+eval "$(direnv hook zsh)"
+
+.fi
+.RE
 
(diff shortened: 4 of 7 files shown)


sbinet:
@xiaq do you think this is a good way to achieve `direnv` support for `elvish` ?
(I am an elvish beginner...)

shell_elvish.go
```diff
@@ -0,0 +1,56 @@
...
+var ELVISH = elvish{re: strings.NewReplacer(
+	`"`, "\"",
+	"'", `\'`,
+)}
+
+func (elvish) Hook() (string, error) {
+	return `fn -direnv-hook {
+	direnv export elvish > .direnv.elv
+	-source .direnv.elv
+	/bin/rm .direnv.elv
```
zimbatm: is it possible to eval the output directly?

shell_elvish.go
```diff
@@ -0,0 +1,56 @@
...
+	return "E:" + k + " = \"" + sh.escape(v) + "\";\n"
+}
+
+func (sh elvish) unset(k string) string {
+	return "del E:" + k + ";\n"
+}
+
+func (elvish) escape(v string) string {
+	v = strings.Replace(v, `"`, `\"`, -1)
+	v = strings.Replace(v, `"`, `\"`, -1)
```
zimbatm: is `"` escaped twice on purpose?

shell_test.go
```diff
@@ -12,6 +12,14 @@ func TestBashEscape(t *testing.T) {
 	assertEqual(t, `$'\xc3\xa9'`, BashEscape("é"))
 }
 
+func TestElvishEscape(t *testing.T) {
+	escape := (elvish{}).escape
+	assertEqual(t, ``, escape(""))
+	assertEqual(t, `escape'quote`, escape("escape'quote"))
+	assertEqual(t, `foo bar`, escape("foo bar"))
+	assertEqual(t, `é`, escape("é"))
```
zimbatm: Can you add a test for `"` and `\` as well ?

shell_elvish.go
sbinet: nope... oops :)

shell_test.go
sbinet: done.

sbinet:
PTAL

shell_elvish.go
zimbatm: Okay, maybe add a note about the eval.

There are two issues with that implementation:
* two shells entering at the same time would overwrite the same file
* if the `-source` is failing the `.direnv.elv` would be staying around

To work around that issue, use `ioutil.TempFile("", "direnv.elv")` to allocate a new random filename. Because the tmpdir is usually tmpfs it will disappear on next reboot.

shell_elvish.go
```diff
@@ -0,0 +1,51 @@
...
+			out += sh.unset(k)
+		default:
+			out += sh.export(k, *v)
+		}
+	}
+	return out
+}
+
+func (sh elvish) export(k, v string) string {
+	return "E:" + k + " = \"" + sh.escape(v) + "\";\n"
```
zimbatm: is the `\n` required or just nice to have for the formatting?

shell_elvish.go
```diff
@@ -0,0 +1,51 @@
...
+}
+
+func (sh elvish) unset(k string) string {
+	return "del E:" + k + ";\n"
+}
+
+func (sh elvish) escape(v string) string {
+	return strings.Replace(v, `"`, `\"`, -1)
```
zimbatm: I think that you first need to escape `\`. Otherwise `\"` would get escaped to `\\"` instead of `\\\"`

xiaq: The easier way to escape strings in Elvish is to use single quoted strings: Add single quotes at both ends and double the single quotes (replacing `'` with `''`). It is not possible to use `strings.Replace` to build a double-quoted strings because you would need to escape `"` and `\` at the same time.

shell_elvish.go
xiaq: My understanding of `direnv` is that to support a new shell (Elvish here), it needs to know two things:

* A piece of hook code to execute at shell startup (`direnv hook`)

* A way to serialize environment changes that can be understood by the shell (`direnv export`).

Now for most shells, the output of `direnv export` is a piece of shell code, but as I understand it, it doesn't have to be; the place where the output of `direnv export` used is in the hook code, and we control the producer as well as the consumer. For Elvish, it makes more sense to make this **not** shell code; evaluating arbitrary code is not well supported and generally frowned upon. Instead you can

* In `Export`, simply output `e` as a JSON map.

* In the hook function, use `from-json` to read the JSON map, iterate through it, and use `{set,unset}-env` to manipulate the environment (blame me for not documenting them; https://github.com/elves/elvish/issues/666).

shell_elvish.go
```diff
@@ -0,0 +1,51 @@
...
+var ELVISH = elvish{}
+
+func (elvish) Hook() (string, error) {
+	return `fn -direnv-hook {
+	direnv export elvish > .direnv.elv
+	-source .direnv.elv
+	/bin/rm .direnv.elv
+}
+
+edit:before-readline = [{ -direnv-hook }]
```
xiaq: Do not overwrite the entire hook list. Use:

```
edit:before-readline = [$@edit:before-readline { -direnv-hook }]
```

or equivalently

```
@edit:before-readline = $@edit:before-readline { -direnv-hook }
```

You can also inline the hook function in here.

shell_elvish.go
zimbatm: Brilliant. There is already a `direnv dump json` so all that needs to be done is on the elvish side. Most shells don't support JSON as an input so I didn't expect that :)

shell_elvish.go
sbinet: I've found it was easier to debug things and/or inspect the output of `direnv export`

zimbatm:
> Due to lack of eval, the setup for Elvish is a bit more complicated. What I recommend is something like: direnv hook > ~/.elvish/lib/direnv.elv; echo 'use direnv' >> ~/.elvish/rc.elv.

@xiaq is there a system path or LD_LIBRARY_PATH equivalent for elvish modules? If yes I could also distribute the direnv.elv as part of the direnv package.

sbinet:
PTAL.

I've added some instructions to the `README`, using directly what @xiaq recommended.
but I suppose one could put the `direnv.elv` file in an elvish module and install it via `epm`:
- https://elvish.io/ref/epm.html

zimbatm:
It looks good to me but I don't have any working knowledge of elvish. It would be nice to add a minimal test-case to make sure it keeps working in the future.

zimbatm:
nice! is there an easy way to download the elvish binary to make it part of the Travis-CI test run?

sbinet:
Right :)
I think I'll install with 'go get'. It's Go after all.

sbinet:
PTAL

shell_elvish.go
```diff
@@ -0,0 +1,42 @@
...
+	} except e {
+		nop
+	}
+}
+`, nil
+}
+
+func (sh elvish) Export(e ShellExport) string {
+	buf := new(bytes.Buffer)
+	err := json.NewEncoder(buf).Encode(e)
```
Mic92: The same export function can be also reused for `xonsh`.

PR #358, opened 2018-04-27
Suggest default go src directory as location for cloning git repo

kpthomas:
For people who install go just for direnv and don't do additional configuration, the `make install` script only works inside directories that go expects.

Using both ubuntu and macOS the script failed with error messages that mentioned the Go install path and default `~/go/src` folder. Moving the direnv repository to the expected folder allowed the script to complete as expected.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,7 @@ nope
 Dependencies: make, golang
 
 ```bash
+cd ~/go/src
 git clone https://github.com/direnv/direnv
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks @kpthomas. I reworked the Makefile so direnv doesn't need to be in the GOPATH anymore in fa4fef76aada582ce2aadc2788239e695581bbc7

PR #369, opened 2018-06-06
hook: pin the direnv path

zimbatm:
Because direnv does thing like allow you to change the $PATH environment
variable, it's wise that direnv itself doesn't depend on that to
function properly. That way if there is a misconfiguration, the user can
edit the broken `.envrc` without having to restart the shell.


ssbarnea:
@zimbatm I was hit by this bug today, can you please make a new release? This bug alone seems like a good enough reason for a new release.

PR #371, opened 2018-07-05
handle `$PIPENV_PIPFILE`

nazarewk:
Adds handling of `PIPENV_PIPFILE` environment variable,
allowing this:
```bash
export PIPENV_PIPFILE="${PWD}/app/Pipfile"
layout_pipenv
```
instead of this:
```bash
cd "${PWD}/app"
layout_pipenv
cd -
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -6,3 +6,4 @@
 /test/allow
 /test/config/direnv/allow
 *.sw?
+/.idea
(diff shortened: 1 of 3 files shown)


nazarewk:
yes
https://github.com/pypa/pipenv/blob/master/docs/advanced.rst#-configuration-with-environment-variables

.gitignore
```diff
@@ -6,3 +6,4 @@
 /test/allow
 /test/config/direnv/allow
 *.sw?
+/.idea
```
zimbatm: I don't mind it but just so you know, this is usually better handled in a [global gitignore](https://stackoverflow.com/questions/7335420/global-git-ignore#7335487) as it's unrelated to the project.

PR #373, opened 2018-07-07
Don't download elvish during the build

### LINKED ISSUE #372: Don't download elvish during build
Support for elvish also added a step to the download the elvish shell and build it from source. This doesn't fit with the packaging workflow where external dependencies are expressed separately.

Also, it is seen as a security risk to download external components during the build - doesn't seem to be the issue here as it's used only during for testing.

punitagrawal:
Elvish needs to be added as an external dependency rather than downloaded and built during the build.

Drop the step to download elvish in the Makefile.

Closes #372

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -50,7 +50,6 @@ man: $(ROFFS)
 test: build
 	go test
 	./test/direnv-test.sh
-	go get github.com/elves/elvish
 	elvish ./test/direnv-test.elv
 
 install: all


zimbatm:
Good idea. As a Nix contributor I should know better than allow fetches in the Makefile :)

Do you mind creating a wrapper script that prepares the test environment for Travis?

zimbatm:
Thanks @punitagrawal , I reworked the Makefile entirely in fa4fef76aada582ce2aadc2788239e695581bbc7

punitagrawal:
Thanks for fixing the issue here - and from the looks of it some more as well.

(Apologies for not responding earlier, I was away from computer)

zimbatm:
no worries at all, your patch gave me the push to finally do things properly :)

PR #376, opened 2018-07-08
WIP: stdlib: introduce layout_virtualenv

zimbatm:
Instead of trying to fix `layout_python`, introduce a new layout that uses optparse to pass optional paths to python and virtualenv.

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -414,7 +414,7 @@ const STDLIB = "#!bash\n" +
 	"  PATH_add vendor/bin\n" +
 	"}\n" +
 	"\n" +
-	"# Usage: layout python <python_exe>\n" +
+	"# Usage: layout python <python_exe> [-v <virtualenv_exe>] [-- [virtualenv-options]]\n" +
 	"#\n" +
 	"# Creates and loads a virtualenv environment under\n" +
 	"# \"$direnv_layout_dir/python-$python_version\".\n" +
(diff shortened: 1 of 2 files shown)


peterfpeterson:
Generally I prefer named options over having to remember the order. It would be a good upgrade to the other functions as well, but would break almost everyone's usage of direnv. I'm suspicious that would not make any fans. 

I suggest that the various `layout_python*` use this new function to reduce repetitive code. The versioned ones could be something like
```shell
layout_python2() {
  layout_virtualenv -p python2 "$@"
}

layout_python3() {
  layout_virtualenv -p python3 "$@"
}
```
and the vanilla one becomes (haven't tested)
```shell
layout_python() {
  local python=${1:-python}
  [[ $# -gt 0 ]] && shift
  layout_virtualenv -p $python "$@"
}
```

This would also make it clearer that direnv's "standard" way to work with python is virtualenv.

zimbatm:
Ok glad to see that we agree. I am trying hard to not break backward-compatibility with direnv which is why I am introducing this new function. Moving towards named argument also seems like a good general trend for the stdlib as a whole.

I will integrate your ideas as well.

Do you see scenarios where the updated functions would break in backward-compatibility?

peterfpeterson:
I just tested this out with an existing `.envrc` whose contents are
```shell
layout python2 --system-site-packages
```
I had forgotten that one has to escape the extra arguments to `virtualenv` so they get through `getopts`. Specifically, the layout needs to now be
```shell
layout python2 -- --system-site-packages
```
which I think is just something that needs to be added to the documentation of direnv in general if the move is towards named arguments.

zimbatm:
Good catch. Since the extra options are not documented, how likely do you think it is that users would depend on this. Is it common to tweak the virtualenv options for python developers?

peterfpeterson:
I might be the only person to give options to virtualenv other than the python version. The bigger concern is that I don't think people realize that they need to tell bash to stop parsing arguments. It is really just a documentation issue.

peterfpeterson:
Because of the way I have it installed (built on a separate system and copy the binary), I don't have the man pages installed. `direnv --help` doesn't have the details for the individual plugins/modules. Therefore, I tend to use the online docs at http://direnv.net or https://github.com/direnv/direnv/wiki. If that fails, I look at the source code on github.

PR #378, opened 2018-07-18
adds experimental curl based installer

ssbarnea:
Curl installer is supposed to ease deployment by automating
all install steps.

This first version will build the code but in the future
we may want to add option for binary install.

### DIFF
--- a/README.md
+++ b/README.md
@@ -42,6 +42,12 @@ nope
 
 ## Install
 
+If you are brave enough to try the experimental curl based installer run:
+
+```
+curl -s https://github.com/direnv/direnv/get-direnv | bash
+```
+
 ### From system packages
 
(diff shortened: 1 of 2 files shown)


zimbatm:
Thanks, I like the idea.

It should do a few more things to make it useful, otherwise it's not going to be worth the extra complexity.

* pull the pre-built binaries. You can get the latest version from https://direnv.net/version.txt and then pull the binaries from https://github.com/direnv/direnv/releases
* run shellcheck and shfmt against the script
* wrap the whole script in a function that is executed at the end. This avoids issues with half-downloaded scripts being executed.
* inform or even ask the user for changing any files in his home

zimbatm:
okay, further PRs should go target the `curl-install` branch

PR #392, opened 2018-09-06
Document possibility to unset vars

jakubholynet:
It was not clear from the current documentation that `.envrc` can not only add variables via `export` but can also remove them via `unset`.


zimbatm:
thanks!

PR #400, opened 2018-09-29
deny: persist the denial to avoid shell abuse with error messages

### LINKED ISSUE #259: Ability to silence direnv nagging about allowing an envrc
In some cases there may be an `.envrc` file that I don't want to allow.
Currently if I use this folder, direnv tells me about allowing .envrc every time I run a command on this project.

Is there any way to silence this message or otherwise reduce the amount of times it is shown?
Something like:

    direnv ignore 1d

would stop nagging for 1 day.

kalbasit:
This commit extends the `deny` command to write down `$XDG_CONFIG_HOME/.direnv/deny/<hash>` similar to how the `allow` command persists the allow based on the hash.

closes #259

### DIFF
--- a/config.go
+++ b/config.go
@@ -113,6 +113,11 @@ func (self *Config) AllowDir() string {
 	return filepath.Join(self.ConfDir, "allow")
 }
 
+// DenyDir returns the directory under which we store the denials
+func (self *Config) DenyDir() string {
+	return filepath.Join(self.ConfDir, "deny")
+}
+
 func (self *Config) LoadedRC() *RC {
 	if self.RCDir == "" {
 		log_debug("RCDir is blank - loadedRC is nil")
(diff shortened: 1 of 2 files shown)


twe4ked:
Thanks for the PR Wael.

Any updates on getting this merged @zimbatm? Thanks :)

zimbatm:
Sorry for the lack of quick feedback, I have been mulling over this for a bit.

I would prefer is this was combined with #406 somehow. Introduce a new `$XDG_CONFIG_HOME/direnv/allow` path and then store the desired outcome in the file as a structure. It doesn't feel right to split the authorization in two places like that.

zimbatm:
Closing this as the abuse has been largely solved by remembering and not reloading on each prompt. I think it's useful to see the denial when first entering the folder.

PR #404, opened 2018-10-11
Add alias '--version' to version command. Closes #377.

### LINKED ISSUE #377: direnv --version does not work
As most tools, direnv should be able to accept the `--version` argument in order to display its current version.

I know that calling direnv without arguments does print its help screen which already includes the first line with version but this is a bit too much info for the case where you just want to check its version.

zerok:


zimbatm:
thanks zerok!

Issue #405, opened 2018-10-17
direnv is tracking env vars coming from before direnv allow


kalbasit:
Consider the following `.envrc`

```shell
$ export CACHED_VAR=found
```

And the following shell commands:

```shell
$ export VAR_SHOULD_NOT_BE_CACHED=found
$ direnv allow
$ echo $CACHED_VAR
$ echo $VAR_SHOULD_NOT_BE_CACHED
```

You would expect to see:

```shell
$ echo $VAR_SHOULD_NOT_BE_CACHED
found
$ echo $CACHED_VAR
found
```

And if you reset the environment (re-open your shell) and inspect the environment, you would not find `VAR_SHOULD_NOT_BE_CACHED`.

```shell
$ echo $VAR_SHOULD_NOT_BE_CACHED
$ echo $CACHED_VAR
found
```

The problem that I am having with my `.envrc, shell.nix` given below, is
that somehow `VAR_SHOULD_NOT_BE_CACHED` in the environment after the
reset. I am using direnv from the master branch.

```shell
use_nix() {
  set -e

  local shell="shell.nix"
  if [[ ! -f "${shell}" ]]; then
    shell="default.nix"
  fi

  local nixpkgs_version=".nixpkgs-version.json"
  if [[ ! -f "${nixpkgs_version}" ]]; then
    fail "use nix: .nixpkgs-version.json not found in the folder"
  fi

  if [[ ! -f "${shell}" ]]; then
    fail "use nix: shell.nix or default.nix not found in the folder"
  fi

  local dir="$(direnv_layout_dir)"
  local default="${dir}/default"
  if [[ ! -L "${default}" ]] || [[ ! -d $(readlink "${default}") ]]; then
    local wd="${dir}/env-$(hashFile "${shell}")-$(hashFile "${nixpkgs_version}")"
    mkdir -p "${wd}"

    local drv="${wd}/env.drv"
    if [[ ! -f "${drv}" ]]; then
      log_status "use nix: deriving new environment"
      IN_NIX_SHELL=1 nix-instantiate --add-root "${drv}" --indirect "${shell}" > /dev/null
      nix-store -r $(nix-store --query --references "${drv}") --add-root "${wd}/dep" --indirect > /dev/null
    fi

    rm -f "${default}"
    ln -s $(basename "${wd}") "${default}"
  fi

  local drv=$(readlink "${default}/env.drv")
  local dump="${dir}/dump-$(hashFile ".envrc")-$(hashFile ${drv})"

[...]

PR #407, opened 2018-11-03
add optional "bash" argument to `direnv dump`

zimbatm:
This is useful in scenarios where an environment is cached between runs

Fixes some of #405 . Only for commands with a --pure option similar to nix-shell.


kalbasit:
oh interesting, let me try my envrc with this patch and see if it helps.

zimbatm:
made a few more improvements

PR #410, opened 2018-11-07
Xdg data home

zimbatm:
This breaks backward-compatibiliy by moving where the allow files are being stored


zimbatm:
finally got around to finishing this :sweat_smile:

PR #412, opened 2018-11-14
Fix typo in readme.

mcpotter:
I noticed that truly wasn't spelled correctly; a quick little thing that should be rectified.


zimbatm:
thanks!

Issue #428, opened 2018-12-14
Support virtualenvs created via "python -m venv"


pmav99:
## Issue description
Unfortunately, there are more than one ways to create a virtualenv in python. 

The most common one is of course the `virtualenv` package which you can get from [PyPi](https://pypi.org/project/virtualenv/). But you can also use the `venv`  module which is [in the standard library](https://docs.python.org/3/library/venv.html). They are roughly equivalent but not 100% compatible ([example](https://bugs.python.org/issue30811#msg313131)). You could probably even say that `venv`, which is a much newer project, is the official way, even though `virtualenv` has a wider userbase.

Anyway, the current implementation of `layout python` does support `virtualenv` but not the `venv` module. The problem is easily reproduced if virtualenv is uninstalled.

My guess is that direnv assumes that the virtualenv script is always present, which might not always be the case. #376  Might also want to take this under consideration.

### Steps to reproduce

1. Make sure that `virtualenv` is not in the path (e.g. uninstalled)
2. Create a new virtualenv using the `venv` module: `python -m venv ./.venv
3. Create an `.envrc` with `layout python`
4. Activate direnv: `direnv allow`
5. You get the following error and the venv is not activated:
```
direnv: loading .envrc                                                                                                                                                                                             
environment: line 447: virtualenv: command not found
direnv: export +VIRTUAL_ENV ~PATH
```

## Technical details

* `direnv version`: 2.18.2
* OS release: 4.19.8-arch1-1-ARCH #1 SMP PREEMPT Sat Dec 8 13:49:11 UTC 2018 x86_64 GNU/Linux

zimbatm:
Since they are not 100% equivalent it might be better to create a new layout for them. I guess `venv` is python bringing virtualenv to the stdlib and so has less support(?).

It would be nice if direnv could source stdlib functions from externally-maintained sources as I don't use python enough to make good judgement calls.

For now I recommend overriding the `layout_python` function in your `~/.config/direnv/direnvrc` with the snippet from https://github.com/direnv/direnv/wiki/Python#venv-stdlib-module

pmav99:
Oh, I didn't know about that wiki page :)
The layout for venv works just fine. 

Not sure if I should close the issue or not though. You might want to keep it open in order to add the layout to direnv, but even if you don't, adding a configuration file is no biggie.

zimbatm:
[...]

PR #429, opened 2018-12-21
stdlib: use venv module for python if available

Mic92:
This removes the dependency on virtualenv, because venv is included
in stdlib since python3.3.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -129,7 +129,7 @@ test-go-fmt:
 	[ $$($(GO) fmt | tee /dev/stderr | wc -l) = 0 ]
 
 test-shellcheck:
-	shellcheck stdlib.sh
+	shellcheck -e SC2239 stdlib.sh
 
 test-bash:
 	bash ./test/direnv-test.bash
--- a/stdlib.go
+++ b/stdlib.go
@@ -19,8 +19,8 @@ const STDLIB = "#!bash\n" +
 	"DIRENV_LOG_FORMAT=\"${DIRENV_LOG_FORMAT-direnv: %s}\"\n" +
 	"\n" +
 	"# This variable can be used by programs to detect when they are running inside\n" +
-	"# of a .envrc evaluation context. All variables starting with `DIRENV_` are\n" +
-	"# ignored by the direnv diffing algorithm and so it won't be re-exported.\n" +
+	"# of a .envrc evaluation context. It is ignored by the direnv diffing\n" +
+	"# algorithm and so it won't be re-exported.\n" +
 	"export DIRENV_IN_ENVRC=1\n" +
 	"\n" +
 	"# Usage: direnv_layout_dir\n" +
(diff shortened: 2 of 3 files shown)


Mic92:
https://docs.python.org/3/library/venv.html

Mic92:
Should I silent this warning?

```
^-- SC2239: Ensure the shebang uses an absolute path to the interpreter.
```

Mic92:
Seems like there is no use for this: https://github.com/direnv/direnv/issues/428

Mic92:
These are the options for `virtualenv`:


```
virtualenv --help
Usage: virtualenv [OPTIONS] DEST_DIR

Options:
  --version             show program's version number and exit
  -h, --help            show this help message and exit
  -v, --verbose         Increase verbosity.
  -q, --quiet           Decrease verbosity.
  -p PYTHON_EXE, --python=PYTHON_EXE
                        The Python interpreter to use, e.g.,
                        --python=python3.5 will use the python3.5 interpreter
                        to create the new environment.  The default is the
                        interpreter that virtualenv was installed with
                        (/nix/store/9vcy8i7mcwmcp486krc22v7851spcdqj-
                        python3-3.7.1/bin/python3.7m)
  --clear               Clear out the non-root install and start from scratch.
  --no-site-packages    DEPRECATED. Retained only for backward compatibility.
                        Not having access to global site-packages is now the
                        default behavior.
  --system-site-packages
                        Give the virtual environment access to the global
                        site-packages.
  --always-copy         Always copy files rather than symlinking.
  --relocatable         Make an EXISTING virtualenv environment relocatable.
                        This fixes up scripts and makes all .pth files
                        relative.
  --no-setuptools       Do not install setuptools in the new virtualenv.
  --no-pip              Do not install pip in the new virtualenv.
  --no-wheel            Do not install wheel in the new virtualenv.
  --extra-search-dir=DIR
                        Directory to look for setuptools/pip distributions in.
                        This option can be used multiple times.
  --download            Download preinstalled packages from PyPI.
  --no-download, --never-download
                        Do not download preinstalled packages from PyPI.
  --prompt=PROMPT       Provides an alternative prompt prefix for this
                        environment.
  --setuptools          DEPRECATED. Retained only for backward compatibility.
                        This option has no effect.
  --distribute          DEPRECATED. Retained only for backward compatibility.
                        This option has no effect.
  --unzip-setuptools    DEPRECATED.  Retained only for backward compatibility.
                        This option has no effect.
```

and this is for `venv`:

```
usage: venv [-h] [--system-site-packages] [--symlinks | --copies] [--clear]
            [--upgrade] [--without-pip]
            ENV_DIR [ENV_DIR ...]

Creates virtual Python environments in one or more target directories.

positional arguments:
  ENV_DIR               A directory to create the environment in.

optional arguments:
  -h, --help            show this help message and exit
  --system-site-packages
                        Give the virtual environment access to the system
                        site-packages dir.
  --symlinks            Try to use symlinks rather than copies, when symlinks
                        are not the default for the platform.
  --copies              Try to use copies rather than symlinks, even when
                        symlinks are the default for the platform.
  --clear               Delete the contents of the environment directory if it
                        already exists, before environment creation.
  --upgrade             Upgrade the environment directory to use this version
                        of Python, assuming Python has been upgraded in-place.
  --without-pip         Skips installing or upgrading pip in the virtual
                        environment (pip is bootstrapped by default)

Once an environment has been created, you may wish to activate it, e.g. by
sourcing an activate script in its bin directory.
```

If you want to go for 100% bug-compatibility this might be a problem if people used `virtualenv` with some options. At the moment it seems that they mostly removed options that were deprecated in virtualenv. The alternative to my implementation would be a new function that you mentioned.

eddies:
@Mic92 If you rebase on master, you can drop your second commit that adds the shellcheck 2239 exception, since that's been merged as part of #432

eddies:
@zimbatm this PR can be closed since it was made part of #505

PR #431, opened 2018-12-23
Support cygwin paths.

vshih:
I assume this will not break native or MinGW environments.


zimbatm:
@vshih the Travis windows support is not generally available yet so I installed AppVeyor instead. I am not terribly interested in doing the Windows support myself but if you want to finish that PR then it would be ok to rebase your work on top of it: https://github.com/direnv/direnv/pull/435/files

Issue #284, opened 2017-07-13
Add support for .env files


fgblomqvist:
I am currently using a tool that only supports the standard .env format (`var=value`), but I would love to be able to use direnv as well. Though, that would require me to have two separate files with the same content in them. One .env and one .envrc. Seems a bit annoying to me.

Would it be possible to add support for .env as well? Is there a reason to why this does not exist?
Other than that direnv seems like a really well-written little tool :+1:

jameysharp:
I just discovered `direnv` today, and I don't use `.env` files myself, so please forgive me if this response is not useful, but:

Is the `dotenv` function in the direnv stdlib not what you're looking for? It looks to me like you should be able to write a `.envrc` containing only this:

```
dotenv
```

and, if I'm reading the documentation correctly, `direnv` will load the `.env` file that's in the same directory.

zimbatm:
Yes that's correct @jameysharp. There are still two files per project but at least you don't have to duplicate the content between both.

fgblomqvist:
Thanks @jameysharp for pointing that out! It's not ideal, but it'll have to do! :)

On a side note, @zimbatm, since .env files are pretty common, maybe the `dotenv` function should be easier to find (document it right in the Readme?) or maybe at least link to where the stdlib docs can be found online (if they can). Just a couple of suggestions.

schickling:
Is there a way to have a `.envrc` file in a parent directory that contains `dotenv` so that it loads `.env` files in child directories.

Ideally there would also be a way to tell direnv to globally load `.env` files (i.e. treat them the same as `.envrc` files).

zimbatm:
The current execution model of direnv doesn't really allow for that kind of mode. Each of the folders will need that `.envrc` right now. It might change in direnv 3.0 where I plan to overhaul the configuration.

The easiest option right now is to add an alias to your shell: `load_dotenv() { echo "dotenv" > .envrc; direnv allow }`

timuckun:
It would be nice to be able to use the same .env. Right now I have to create a symlink between them (the ruby dotenv library is able to parse the direnv format).

Ideally direnv could handle key=value format without needing to type EXPORT.

timwis:
I'd love this too. In the communities I interact with most frequently, .env is more common than .envrc. it would be great if this worked with either.

nitrocode:
I use `.env` for a lot of different projects. Why put `dotenv` in your `.envrc` file when a softlink is even easier.

```
[...]

PR #438, opened 2019-01-03
Add automatic ".env" compatibility

aude:
This is work towards using `.env` if there is no `.envrc` (#284).

I love what direnv does, and would really like to use it more :smile: Having automatic "fallback" to `.env` would remove a lot of friction for me. I figured the best way to achieve that would be to help implement it.

The code is WIP now. It works, but might be a bit wrong/incoherent. I would appreciate any feedback, I'm quite new to Go.

1. Naming and coding style could be off with regard to the rest of the project, just change it however you'd like.
2. `.env` files are not loaded "layered" now, same as `.envrc`. I would personally like `.env` loading to work by default like `source_up`. Thoughts?
3. I changed `DIRENV_DIR` to `DIRENV_FILE`, to keep track of the RC file used. Is that breaking anything?
4. RC file "detection" is done by simply checking the file name. I think this is a bit cheap, maybe there should be a `type` attribute on the `RC` object instead. Thoughts?
5. I haven't added any tests for `.env` loading. I'd like to add some, but could take some advice what should be tested.


zimbatm:
This introduces a few issues with backward-compatibility:
* people rely on the existence of `$DIRENV_DIR` to check if an environment has been loaded for example
* the `.env` might get loaded even if this is not the user's intention
I don't have a great solution to fix that unfortunately. Ideally the behaviour would be configurable in some way.

aude:
I have been using this fork for the last week now without issues, for what it's worth.

aude:
Added back `DIRENV_DIR`, so that's not an issue anymore.

alvis:
Thank you @aude. This PR is highly anticipated!
@zimbatm: shell we merge this?

aude:
I did not mean to close this PR. I just rebased it on master, to keep the PR up to date.

I think one approach here could be to merge it and get it out to people. Then we could add improvements as we go, if needed.

I've been using this code successfully since January 3, 2019, so that could vouch for it working.

schickling:
:shipit:

zimbatm:
@aude I added some minimal testing, which probably needs to be fixed. Having it work on your machine is not nearly sufficient to make sure it doesn't break other people's workflows as direnv is used in many different contexts. Please also amend the documentation to advertise the fact that direnv is now also looking for .env files.

aude:
Thanks for taking a look. I'll take a look at the tests and add documentation.

aude:
Sorry for letting you guys down here. I lost motivation some months ago, I'm currently not motivated to finish the PR.

Hoping that changes, can't promise anything though.

joaomcarlos:
:shipit: !

varac:
Ping, any progress here ?

zimbatm:
@varac do you want to pick this up?

JakeElder:
Is this likely to be merged? I'd like to see this functionality included. Great work with direnv, thank you.

zimbatm:
It depends on somebody stepping up and fixing the PR.

JakeElder:
> It depends on somebody stepping up and fixing the PR.

Understood. I'd like to have a go but have very little time, knowledge of bash and experience with go, so I'm not best equipped 🤓. Thanks again for the package.

robert-r-meyer:
@aude @zimbatm I'd be happy to finish this off given commit permissions to the fork.

felipecrs:
I created a PR as continuation of this: https://github.com/direnv/direnv/pull/845

It should be ready to review, any feedback is welcome.

zimbatm:
Released in v2.30.1!

Issue #288, opened 2017-08-12
.envrc a symlink does not throw error


brikou:
I want to easily switch .envrc (based on environment in my case).

Here is the way to reproduce my issue...

```sh
cd $(mktemp -d)

cat << 'EOF' > .envrc--production
export ENV=production
EOF

cat << 'EOF' > .envrc--development
export ENV=development
EOF

ln --symbolic --force .envrc--development .envrc
# direnv: error .envrc is blocked. Run `direnv allow` to approve its content.

direnv allow

echo $ENV
# development
```

If I want to switch env to production I then run:

```sh
ln --symbolic --force .envrc--production .envrc
# (...no error thrown)

echo $ENV
# development (instead of production, ... nothing loaded)
```

No error is throw and $ENV is still set to development. I still can force it with `direnv reload`, but it is ugly.

WDYT, is it a bug or a feature :)

zimbatm:
Hey it's a cool idea. I would like to treat this as a bug :)

Can you run `direnv status` before and after the switch and also `stat .envrc--development .envrc--production`?

brikou:
You can safely copy/paste the following snippet on your own, and so be able to reproduce the whole thing.

I can wait to have this ~~feature available~~, bug solved. BTW direnv is amazing!

```sh
cd $(mktemp -d)

cat << 'EOF' > .envrc--production
export ENV=production
EOF

cat << 'EOF' > .envrc--development
export ENV=development
EOF

ln --symbolic --force .envrc--development .envrc
# direnv: error .envrc is blocked. Run `direnv allow` to approve its content.

direnv allow

echo $ENV
# development

direnv status
# direnv exec path /usr/local/bin/direnv
# DIRENV_CONFIG /home/brikou/.config/direnv
# DIRENV_BASH /bin/bash
# Loaded RC path /tmp/tmp.5tvGz3Yoyq/.envrc
# Loaded watch: ".envrc" - 2017-08-12T15:28:21+02:00
# Loaded watch: "../../home/brikou/.config/direnv/allow/be0ef2f322890e4e7fd2d29c957c996f9d66517f837b46059bfe4a0c9cd88ed2" - 2017-08-12T15:28:21+02:00
# Loaded watch: "../../home/brikou/.direnvrc" - 2017-08-12T14:04:03+02:00
# Loaded RC allowed false
# Loaded RC allowPath 
# Found RC path /tmp/tmp.5tvGz3Yoyq/.envrc
[...]

PR #452, opened 2019-01-30
file_times: use Lstat as well as Stat

grahamc:
Stat only looks at the target of symlinks, and Lstat
examines the symlink itself.

This change allows changes to both the symlink and
the symlink's target to trigger updates.

### DIFF
--- a/file_times.go
+++ b/file_times.go
@@ -29,6 +29,8 @@ func (times *FileTimes) Update(path string) (err error) {
 	var modtime int64
 	var exists bool
 
+	// Handle the path with Stat, which looks at the other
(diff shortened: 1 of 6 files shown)


grahamc:
You bet, I added a test, pushed it, verified it failed, and then reapplied my patch and it passed.

wbolster:
awesome, thanks! [stat + lstat it is!](https://github.com/direnv/direnv/issues/288#issuecomment-394628684)

grahamc:
See https://github.com/direnv/direnv/pull/457 for  further fixup.

PR #457, opened 2019-02-05
file_times: check Stat and Lstat

grahamc:
Only trigger a reload if either Stat or Lstat don't match the expected
time exactly.

Before this patch, if direnv was watching a symlink, direnv always
thought the file was stale, because the Lstat didn't match the Stat
results.

### DIFF
--- a/file_times.go
+++ b/file_times.go
@@ -29,9 +29,7 @@ func (times *FileTimes) Update(path string) (err error) {
 	var modtime int64
 	var exists bool
 
-	// Handle the path with Stat, which looks at the other
-	// end of symlinks
-	stat, err := os.Stat(path)
+	stat, err := getLatestStat(path)
 	if os.IsNotExist(err) {
 		exists = false
 	} else {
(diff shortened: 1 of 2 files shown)


grahamc:
I tried to write a test for it, but wasn't able to make it work. Essentially, it looked like this:

```bash
test_start "symlink-unchanged"
  # when using a symlink, do not reload on every prompt
  export A=2
  direnv_eval
  test_eq "${A}" "1"
  echo "$A"
  direnv_eval
  echo "$A"
  test_eq "${A}" "1"
  echo "$A"
test_stop
```

the test never reported the broken behavior.

file_times.go
```diff
@@ -134,20 +134,32 @@ func (times *FileTimes) CheckOne(path string) (err error) {
 
 func (time FileTime) Check() (err error) {
 	stat, err := os.Stat(time.Path)
+	lstat, lerr := os.Lstat(time.Path)
```
zimbatm: can you extract the Stat+Lstat function to getLatestModTime(file) function?

grahamc: I could!

Issue #443, opened 2019-01-19
Support for loading shell completion scripts


jcpetruzza:
It would be nice to be able to load new {bash/zsh/etc}-completions from an `.envrc` script. A typical use case would be when `use nix` provisions some commands that you don't have globally installed (or are for different versions so their command-line args differ).

This could work for example like this:

  - The stdlib provides a new command, say, `add_completions` to register completion scripts inside `.envrc`. E.g.
```bash
add_completions bash /path/to/foo/share/bash-completions/completions/
add_completions zsh /path/to/foo/share/zsh/site-functions/
``` 
  - For each supported shell, their `direnv export` action emits, along with the `export`/`unset` commands,  code to add (and remove) the registered completions.

1oglop1:
I was exactly looking for this feature. For example I use `awscli` which requires `complete -C aws_completer aws` but I do not want to have enabled for the system because my `awscli` is installed in specific virtualenv

zimbatm:
As long as the implementation includes proper tests. direnv is missing too many tests at the moment that I feel confident shipping new features. The DIRENV_DIFF format needs to be changed to accommodate for the diffing of more things than environment variables as well.

jcpetruzza:
If we can get something like `shell_specific` in #464 merged in, then one can implement this as follows.

  - `add_completions bash dir` emits bash-specific code that:
    1. Runs `complete` and captures the output. This gives us all existing completions before loading the env 
    2. Runs `source` on all files in `dir`
    3. Runs `complete` again and diffs the output with that of step 1 (there will only be additions and modifications).
    4. Registers an ON_UNLOAD action that for each modified entry `xxx`, runs `complete -r xxx` if it was a new entry (deleting the entry), or the definition from step 1 if was a redefinition.

  - `add_completions zsh` emits zsh-specific code that:
    1. Gets all currents completions with something like
        ```zsh
        for k in ${(k)_comps:#-*(-|-,*)}; do
          printf "%q;%q" $k ${_comps[$k]}
        done
        ``` 
    2. Adds `dir` to the `rpath` array
    3. Runs `compinit` so that all new completion defs get indexed
    4. Gets the current completions again and diffs the output with that of step 1
    5. Registers an ON_UNLOAD action that:
        1. Runs `compdef -d xxx` on every command `xxx` that appeared in the diff of step 4.
        2. Runs `compinit -D` to force a reindex of completions (otherwise, definitions that were added when loading the env may persist).

[...]

Issue #73, opened 2013-08-20
Picking up aliases and functions


pwaller:
Just come across a case where someone was expecting functions and aliases to be picked up from the source script.

I realise what a pain this is. Any way we can reasonably make that work?

pwaller:
I'm going to close this one since there hasn't been much interest and I haven't missed it much myself. But if there is demand, please open up another issue. Even better if you can suggest ways to implement it.

jneen:
I would use this feature.  At the very least support for `export -f`.  I have projects, both personal and for work, that require particular startup scripts tailored to my environment, and it'd be really useful to just throw a function in .envrc.

aeonaut:
+1! I could definitely use this. I've been using my own little `cd` hook for a while to load some functions and aliases when I change into a directory; `direnv` is a much more robust solution but not if I have to still keep my old hook around. As mentioned above, I use aliases like `alias s="rails server"`, `alias c="rails console"`, and at the moment `alias ee="stty echo"` as a workaround for a bug in [pry](https://github.com/pry/pry) that routinely kills my terminal echo -- I don't really want any of those in my global namespace. Another use case: I have a local function called `migrate` (for database migrations) in my Rails project that shadows a completely different (and rarely used) executable that has nothing to do with Rails. (As mentioned above, this is perhaps a less pressing use case than the aliases because I could just drop that in a ./bin directory and `PATH_add ./bin`; still, it's handy to throw that function in my .envrc).

airtoxin:
+1 for `export -f`

zimbatm:
Unfortunately BASH_FUNC_* environment variables cause issues for other shell so we had to ignore them. See https://github.com/direnv/direnv/pull/157

If you want to maintain all your aliases in the .envrc you could add something like that in your main ~/.direnvrc file:

```bash
export_function() {
  local name=$1
  local alias_dir=$PWD/.direnv/aliases
  mkdir -p "$alias_dir"
  PATH_add "$alias_dir"
  local target="$alias_dir/$name"
  if declare -f "$name" >/dev/null; then
    echo "#!/usr/bin/env bash" > "$target"
    declare -f "$name" >> "$target" 2>/dev/null
    echo "$name" >> "$target"
    chmod +x "$target"
  fi
}
```

then in your .envrc you can write something like:
```bash
woot() {
  echo woot
}
export_function woot
```

An executable file is then automatically generate on each .envrc load inside of the project's .direnv/aliases folder and made available on the PATH.

[...]

PR #464, opened 2019-02-26
Add shell-specific actions and use them to implement aliases

jcpetruzza:
This implements a mechanism to send shell-specific code to the host shell from the `.envrc` file, along with a simple protocol to register clean-up actions to be performed when unloading the environment. The supported shells at the moment are `bash`, `zsh`, `tcsh` and `fish` (`elvish` is missing only because it doesn't have an `eval` function, or I couldn't find it).

Based on this, we implement the exporting of aliases from `.envrc`  (#73). The same mechanism should let us support loading/unloading shell-completion scripts (#443) and, more in general, run arbitrary commands on entry/exit of the environment.

The core of `direnv` was left mostly unaffected; this is a rundown of the changes:

  * The first 2 commits add tests for `tcsh` and `zsh`, so that we can then test the new functionality on all shells. For the latter, we can simply share the test script with `bash` as long as we stay within bourne shell code. 

  * The next 3 commits fix minor issues in tests.

  * Next, we introduce a `quote` command to the stdlib. Intuitively, `quote fish foo bar` will add `foo bar;` to the output of `direnv export fish`, etc. This works by accumulating the quotes for shell `xxsh` on the `DIRENV_QUOTES_xxsh` environment variable (excluded from `DIRENV_DIFF`).

  * `quote` is a low-level operation: it will only send commands when the environment is loaded. The next 5 commits add the higher-level `shell_specific` command to the stdlib that allows one to also register clean up actions. If an `.envrc` file were to include the following:
    ```bash
    shell_specific zsh "echo 'rm -f foo'" "touch foo"
    ```
    Then loading the environment with `direnv export zsh` would have, conceptually, the following additional effect:
    ```bash
    export DIRENV_ON_UNLOAD_zsh=$(echo 'rm -f foo')
    touch foo
    ```
     `DIRENV_ON_UNLOAD_zsh` is honoured by `direnv export zsh` when unloading the environment, which would then include the extra effect:
    ```bash
    rm -f foo
    ```
  * The second argument to `shell_specific` is conceptually a function computing the clean-up action. In the above example, we use `echo 'rm -f foo'` as a constant function that always returns `rm -f foo`. For a more realistic example, when implementing aliases for `bash`, we can exploit that `alias foo` outputs the full definition of `foo` (including the `alias` keyword) and use something like:
    ```bash
    shell_specific bash 'alias foo || echo unalias foo' 'alias foo=ls -lt'
    ```  
  * In reality, `DIRENV_ON_UNLOAD_xxsh` is a comma-separated list of gzenv-encoded quotes. In order to be able to add elements to this list from the envrc, we added a `direnv gzenv [encode|decode]` private command.

  * The last commit adds a `use_aliases` function to the stdlib, so that the alias-exposing functionality becomes opt-in. The usage is:
    ```bash
    alias hello=world  # hello will not be exported

    use aliases
    alias foo=bar  # foo will be exported
    ```

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -143,6 +143,7 @@ test-fish:
 
 test-tcsh:
 	tcsh -e ./test/direnv-test.tcsh
+	tcsh    ./test/direnv-test-lax.tcsh
 
 test-zsh:
 	zsh ./test/direnv-test.zsh
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -145,8 +145,7 @@ test-tcsh:
 	tcsh -e ./test/direnv-test.tcsh
 
 test-zsh:
-	# Currently missing
-	#zsh ./test/direnv-test.zsh
+	zsh ./test/direnv-test.zsh
 
 ############################################################################
 # Installation
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -142,8 +142,7 @@ test-fish:
 	fish ./test/direnv-test.fish
 
 test-tcsh:
-	# Currently broken
-	#tcsh -e ./test/direnv-test.tcsh
+	tcsh -e ./test/direnv-test.tcsh
 
 test-zsh:
 	# Currently missing
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ eval "$(direnv hook zsh)"
 Add the following line at the end of the `~/.config/fish/config.fish` file:
 
 ```fish
-eval (direnv hook fish)
+direnv hook fish | source
 ```
 
 ### TCSH
--- a/cmd_apply_dump.go
+++ b/cmd_apply_dump.go
@@ -33,7 +33,7 @@ var CmdApplyDump = &Cmd{
 
 		diff := env.Diff(dumpedEnv)
 
-		exports := diff.ToShell(BASH)
+		exports := diff.ToShell(BASH, nil)
 
 		_, err = fmt.Println(exports)
 		if err != nil {
--- a/cmd_dotenv.go
+++ b/cmd_dotenv.go
@@ -45,7 +45,7 @@ var CmdDotEnv = &Cmd{
 			return err
 		}
 
-		str := env.ToShell(shell)
+		str := env.ToShell(shell, nil)
 		fmt.Println(str)
 
 		return
--- a/cmd_export.go
+++ b/cmd_export.go
@@ -14,6 +14,7 @@ type ExportContext struct {
 	env      Env
 	oldEnv   Env
 	newEnv   Env
+	quotes   ShellQuotes
 }
 
 func (self *ExportContext) loadConfig() (err error) {
@@ -70,6 +71,9 @@ func (self *ExportContext) updateRC() (err error) {
 
 func (self *ExportContext) loadRC() (err error) {
 	self.newEnv, err = self.foundRC.Load(self.config, self.oldEnv)
+	if err == nil {
+		self.quotes = self.newEnv.GetShellQuotes()
+	}
 	return
 }
 
@@ -77,6 +81,7 @@ func (self *ExportContext) unloadEnv() (err error) {
 	log_status(self.env, "unloading")
 	self.newEnv = self.oldEnv.Copy()
 	cleanEnv(self.newEnv)
+	self.quotes = nil
 	return
 }
 
@@ -87,6 +92,7 @@ func (self *ExportContext) resetEnv() {
 		delete(self.newEnv, DIRENV_DIFF)
 		self.foundRC.RecordState(self.oldEnv, self.newEnv)
 	}
+	self.quotes = nil
 }
 
 func cleanEnv(env Env) {
@@ -123,7 +129,7 @@ func (self *ExportContext) diffString(shell Shell) string {
 	}
 
 	diff := self.env.Diff(self.newEnv)
-	return diff.ToShell(shell)
+	return diff.ToShell(shell) + self.quotes[shell]
 }
 
 func exportCommand(env Env, args []string) (err error) {
--- a/cmd_export.go
+++ b/cmd_export.go
@@ -72,16 +72,18 @@ func (self *ExportContext) updateRC() (err error) {
 func (self *ExportContext) loadRC() (err error) {
 	self.newEnv, err = self.foundRC.Load(self.config, self.oldEnv)
 	if err == nil {
-		self.quotes = self.newEnv.GetShellQuotes()
+		unloading_quotes := self.oldEnv.GetOnUnloadShellQuotes()
+		loading_quotes := self.newEnv.GetShellQuotes()
+		self.quotes = MergeShellQuotes(unloading_quotes, loading_quotes)
 	}
 	return
 }
 
 func (self *ExportContext) unloadEnv() (err error) {
 	log_status(self.env, "unloading")
+	self.quotes = self.env.GetOnUnloadShellQuotes()
 	self.newEnv = self.oldEnv.Copy()
 	cleanEnv(self.newEnv)
-	self.quotes = nil
 	return
 }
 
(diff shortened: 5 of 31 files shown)


zimbatm:
thanks @jcpetruzza, I already integrated all your test improvements.

Sorry but the rest of the PR changes direnv too much and I am not comfortable merging this. I can see direnv potentially managing more things than just environment variables but it would have to be done declaratively where each shell hook would be expanded to handle the diffing.

jcpetruzza:
@zimbatm Thanks for your response! I think it is fine not to merge the rest of this PR as it is. I was initially going to submit only the minimal number of changes that would allow people to implement things like aliases or shell-completion on their own (and share the implementations on the wiki), but I felt without a fully-developed example it wouldn't be clear that the idea works, etc, so I ended up adding a `use_alias` example to the stdlib. I see the code in this PR as a working proof-of-concept that I would like to use to discuss the design of a direnv extension to handle additional effects that is good to be merged in.

The main take-aways, for me, from implementing this prototype are:
  1. From the rc script one needs to be able to signal to the running `direnv export xxsh` command, (via `direnv dump`) that there is some xxsh-specific action to perform, so that this can be ultimately sent to the hook. I don't think one can use anything better than a reserved environment variable for this. A design choice here is whether `direnv dump` interprets it and then outputs something more complex than an `Env`, or if it just passes it through as an environment variable. In this implementation, I used a `DIRENV_QUOTES_xxsh` variable, let it pass through `direnv dump` and the "signal" was just the raw xxsh-command to run. If I understand correctly, you'd prefer, as "signal", some sort of message (e.g., "define this alias...", etc) that then the `func (sh xxsh) Export(..)` can turn into xxsh-specific code. I think that in any case, the possibility to send raw xxsh-code should be there as well, so people can experiment with ideas without having to modify the direnv implementation; this could be achieved with a message that'd say "run this xxsh-code".
  1. Whatever effect is ultimately executed in the hook, we will normally want to reverse it when unloading the env. The problem is that in general we won't know how to reverse the effect until the code in the hook runs (e.g. one needs to know what aliases are overwritten to be able to restore them). The only way I can see to implement this is by having the hook set an environment variable that will be used by direnv during the unloading of the env. Here I've used `DIRENV_ON_UNLOAD_xxsh` for this. One would want to use some gzenv-encoding for this (as is done for `DIRENV_WATCHES`), so an additional direnv command will probably need to be implemented as well.

Do you think this could be implemented in any other way?

I'd propose then the following design for the extension:

  * We add two reserved environment variables: `DIRENV_LOAD_ACTIONS` and `DIRENV_UNLOAD_ACTIONS`. They will contain a gzenv-encoded list of "actions".
  * `DIRENV_LOAD_ACTIONS` is to be set by the rc-script and honored by direnv during an environment load.
  * `DIRENV_UNLOAD_ACTIONS` will be set by the eval-call installed by the shell-hook, and honored by direnv during an environment unload.
  * An "action" is a pair ("action_type", "action_payload"). Initially, the action type will be one of: `quote_bash`, `quote_zsh`, `quote_tcsh`, `quote_fish`; and the payload a command for that shell. More higher-level actions can be added in the future as needed.
  * We add a `direnv add_action` command, analogous to `direnv watch`. It uses `DIRENV_IN_ENVRC` to decide whether to target `DIRENV_LOAD_ACTIONS` or `DIRENV_UNLOAD_ACTIONS`.
  * A function `add_action` is added to the stdlib, analogous to `watch_file`.
  * A function `on_direnv_unload` is installed by the hook of each shell (similar to `add_action`), to be used by the code that needs to install an unload-action.

Because this is a cleaned-up version of what is already in this PR, we have a validation that it should work. This also doesn't require many changes to direnv internals, imo. One can then play around, implement something cool like "use direnv to activate a local nix profile and have the shell-completion scripts updated", and share it on the wiki :slightly_smiling_face: 

Thoughts?

zimbatm:
sorry I haven't had much time to think and this is a bit of a deep subject.

One of the fundamental difference between environment variables and the rest is that env vars are inherited between processes. So I can `cd` into a project with my bash shell, start zsh and provided both have direnv installed, everything will keep working as before. I can `cd ..` in the zsh shell and unload the env, etc... If I exit the zsh shell, I will be back to the bash shell with it's state intact as well.

Now if I think about aliases, what should happen is that the bash hook, somehow keeps track of it's own aliases. If a special DIRENV_ALIASES environment variable is created after an eval, it would have to backup the old aliases and apply the new ones itself. And somehow on every prompt command it would have to check if DIRENV_ALIASES has changed so that it could work when invoked as a sub-shell. Here I am assuming that the aliases are simple and compatible with all of the shells. It also puts the burden on each shell implementation to implement their own aliasing mechanism.

This is the reverse of your LOAD/UNLOAD actions as the responsibility is shifted from the .envrc to the specific shell for doing the backup and restore. But it means that aliases can actually be restored which I don't think your approach handles.

Another reason for my reluctance is that I would prefer to merge all the DIRENV_ environment variables into one DIRENV_STATE at some point, so introducing new variables is not ideal. For that I think your gzenv utility is quite helpful. I realize that it's not directly related to the PR and I would be happy to be working with you on that first if that's alright with you.

jcpetruzza:
The point about subshells is a very good one, and I hadn't thought about that, really. You don't even need to go the bash -> zsh route, already bash -> bash and zsh -> zsh is not working in this prototype, since an alias defined by the envrc will "disappear" from the subshell. 

> Now if I think about aliases, what should happen is that the bash hook, somehow keeps track of it's own aliases. [...] And somehow on every prompt command it would have to check if DIRENV_ALIASES has changed so that it could work when invoked as a sub-shell.

I think I would do this differently, though. The code emitted by `direnv hook` could set some random UUID in a DIRENV_HOOK_ID envrironment variable, and the contents of the proposed DIRENV_STATE could contain the hook id of the shell that produced it by calling `direnv export xxsh`. That way, when you launch a subshell, it will get a different DIRENV_HOOK_ID and the call to`direnv bash xxsh` associated to the prompt-command can then notice that given there is a DIRENV_STATE for the right envrc file and the hook id is different, the envrc needs to be rerun (or the output re-emitted since we can presumably reproduce it from the state).

> Here I am assuming that the aliases are simple and compatible with all of the shells. It also puts the burden on each shell implementation to implement their own aliasing mechanism.

> This is the reverse of your LOAD/UNLOAD actions as the responsibility is shifted from the .envrc to the specific shell for doing the backup and restore. But it means that aliases can actually be restored which I don't think your approach handles.

Wait, this prototype is already restoring the aliases! :slightly_smiling_face: Look at the test for the scenario called "alias": it checks that on entering a directory two aliases `foo` and `bar` are set according  to the definition of the envrc and, on exiting the directory, are restored to their previous value (in the case of `foo`, by removing it). And works on `bash`, `zsh`, `fish` and `tcsh`, which required, in each case, shell-specific code handle it.

If I understand correctly, what you are proposing is to make the hook more smart, but also more complex (first is has to know about aliases, and tomorrow about shell-completions, etc), while in my approach the hook remains just as simple as it is right now and users can add new shell-specific extensions without having to touch direnv internals, since they just need to use the envrc to emit the shell-specific code that will eventually run on the hook (not the most beautiful thing to write, but well...)

> Another reason for my reluctance is that I would prefer to merge all the DIRENV_ environment variables into one DIRENV_STATE at some point, so introducing new variables is not ideal. For that I think your gzenv utility is quite helpful. I realize that it's not directly related to the PR and I would be happy to be working with you on that first if that's alright with you.

Moving towards a single DIRENV_STATE variable makes sense to me and I'd be happy to help with that. It would be a biggish change, though, so it's probably better to do it incrementally. Do you want to open an issue to lay out the plan and track the progress?

varac:
What's the state of the PR ? I'd love to define directory specific aliases with direnv !

seanking2919:
Any update on this?

zimbatm:
It's not very likely to happen. @jcpetruzza did a fantastic job, and I don't feel comfortable merging this at the same time. It's too much new surface of untested code for this little one-man project.

If somebody were to abstract the changes record to include aliases and the shell process id, on top of environment variables, then it might be doable.

PR #466, opened 2019-02-27
Corrects reverse patching when using exec cmd.

MenBrial:
The cleaned-up environment created after reverse patching the current environment was not used for the exec cmd.


zimbatm:
I am ashamed of this one, and it has been around for a long time now. thanks for the fix!

PR #473, opened 2019-03-07
Fix bash hook

edubxb:
When the direnv is the first element added to PROMPT_COMMAND
(PROMPT_COMMAND is empty) the trailing semicolon character is unneeded,
and, in some cases, breaks the prompt due an invalid syntax when more
elements are added.


zimbatm:
thanks @edubxb, much appreciated

PR #482, opened 2019-03-28
fix arity mismatch error

babariviere:
this also remove `echo $e` which is annoying when there is no .envrc in current or parent folders

### DIFF
--- a/shell_elvish.go
+++ b/shell_elvish.go
@@ -13,12 +13,15 @@ func (elvish) Hook() (string, error) {
 	return `## hook for direnv
 @edit:before-readline = $@edit:before-readline {
 	try {
-		m = ("{{.SelfPath}}" export elvish | from-json)
-		keys $m | each [k]{
-			if (==s $k 'null') {
-				unset-env $k
-			} else {
-				set-env $k $m[$k]
+		m = [("{{.SelfPath}}" export elvish | from-json)]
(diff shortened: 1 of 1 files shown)


zimbatm:
nope :) This would prevent direnv from reloading the environment automatically whenever the `.envrc` changes.

Is it possible to fix the `echo $e` issue independently instead?

babariviere:
Oh sorry, I hadn't think about that

We could just export an empty json to avoid error

zimbatm:
no problem :) sounds good, let me know if you need any help for the fix

babariviere:
I have changed this pull request to instead remove "nil" error

zimbatm:
thanks :)

babariviere:
I'm sorry, I have made a mistake. In if conditions, it shouldn't be 1 but 0. Really sorry :disappointed:

zimbatm:
@babariviere are you going to submit another PR fix? maybe I should have asked for some additional tests :p

babariviere:
I don't think another PR is necessary haha. I think so but I don't know how to test with elvish

zimbatm:
fixed in 9f5ee3eaacc90b8b6ca5a639c197781d92e9ff97

babariviere:
Thanks ! (and sorry at the same time)

PR #493, opened 2019-04-17
direnv exec: DIR is always necessary

Profpatsch:
```
$ direnv exec bash -c "true"
direnv: error stat bash: no such file or directory
$ direnv exec bash
direnv: error stat bash: no such file or directory
```

Meaning leaving out the DIR argument was never really an option.


zimbatm:
thanks!

PR #499, opened 2019-05-28
[use_nix] Passing arguments to use_nix should not determine whether files are watched

kampka:
The current implementation of use_nix disables watching the `default.nix` and `shell.nix` files as soon as any arbitrary arguments are passed.
This makes some sense if you assume that one passes only packages to nix-shell that should be used instead of the shell files.
However, nix-shell accepts a lot of other useful options, like cache control, build concurrency control and others. Turning of watching the default files in these cases is not expected.
This pull request decouples the fact of whether the default nix files should be watched from the nix-shell arguments. Instead, a specific argument is added that can be passed to configure this behavior.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -683,12 +683,19 @@ use_node() {
 #
 # Load environment variables from `nix-shell`.
 # If you have a `default.nix` or `shell.nix` these will be
-# used by default, but you can also specify packages directly
-# (e.g `use nix -p ocaml`).
+# used by default unless NIX_NO_WATCH_DEFAULT is set to "true"
+# or by passing `--no-watch-default` as first parameter
+# Arguments are passed to nix-shell as is, eg. you can specify packages directly
+# using `use nix -p ocaml`.
 #
 use_nix() {
+  local NIX_NO_WATCH_DEFAULT="${NIX_NO_WATCH_DEFAULT:-false}"
+  if [[ "${1:-}" == "--no-watch-default" ]]; then
+    NIX_NO_WATCH_DEFAULT="true"
+    shift
+  fi
   direnv_load nix-shell --show-trace "$@" --run "$(join_args "$direnv" dump)"
-  if [[ $# == 0 ]]; then
+  if [[ "${NO_WATCH_DEFAULT:-false}" == "false" ]]; then
     watch_file default.nix
     watch_file shell.nix
   fi
--- a/stdlib.go
+++ b/stdlib.go
@@ -685,12 +685,19 @@ const STDLIB = "#!/usr/bin/env bash\n" +
 	"#\n" +
 	"# Load environment variables from `nix-shell`.\n" +
 	"# If you have a `default.nix` or `shell.nix` these will be\n" +
-	"# used by default, but you can also specify packages directly\n" +
-	"# (e.g `use nix -p ocaml`).\n" +
+	"# used by default unless NIX_NO_WATCH_DEFAULT is set to \"true\"\n" +
+	"# or by passing `--no-watch-default` as first parameter\n" +
+	"# Arguments are passed to nix-shell as is, eg. you can specify packages directly\n" +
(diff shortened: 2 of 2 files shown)


stdlib.sh
```diff
@@ -683,12 +683,19 @@ use_node() {
...
 #
 use_nix() {
+  local NIX_NO_WATCH_DEFAULT="${NIX_NO_WATCH_DEFAULT:false}"
+  if [[ "$1" == "--no-watch-default" ]]; then
+    NIX_NO_WATCH_DEFAULT="true"
+    shift
+  fi
   direnv_load nix-shell --show-trace "$@" --run "$(join_args "$direnv" dump)"
-  if [[ $# == 0 ]]; then
+  if [[ "${NO_WATCH_DEFAULT:false}" == "false" ]]; then
```
kalbasit:
```suggestion
  if [[ "${NO_WATCH_DEFAULT:-false}" == "false" ]]; then
```

https://wiki.bash-hackers.org/syntax/pe#use_a_default_value

stdlib.sh
```diff
@@ -683,12 +683,19 @@ use_node() {
...
 # If you have a `default.nix` or `shell.nix` these will be
-# used by default, but you can also specify packages directly
-# (e.g `use nix -p ocaml`).
+# used by default unless NIX_NO_WATCH_DEFAULT is set to "true"
+# or by passing `--no-watch-default` as first parameter
+# Arguments are passed to nix-shell as is, eg. you can specify packages directly
+# using `use nix -p ocaml`.
 #
 use_nix() {
+  local NIX_NO_WATCH_DEFAULT="${NIX_NO_WATCH_DEFAULT:false}"
```
kalbasit:
```suggestion
  local NIX_NO_WATCH_DEFAULT="${NIX_NO_WATCH_DEFAULT:-false}"
```

https://wiki.bash-hackers.org/syntax/pe#use_a_default_value

stdlib.sh
```diff
@@ -683,12 +683,19 @@ use_node() {
...
+  if [[ "$1" == "--no-watch-default" ]]; then
```
kalbasit: what if `$1` does not exist? I use `set -u` within my `.envrc` and this will fail. Please make sure there are arguments first.

kampka: Valid point, fixed.

kampka:
@kalbasit thanks for your feedback
I updated the commit according to your suggestions.

stdlib.sh
```diff
@@ -683,12 +683,19 @@ use_node() {
...
 # If you have a `default.nix` or `shell.nix` these will be
-# used by default, but you can also specify packages directly
-# (e.g `use nix -p ocaml`).
+# used by default unless NIX_NO_WATCH_DEFAULT is set to "true"
+# or by passing `--no-watch-default` as first parameter
+# Arguments are passed to nix-shell as is, eg. you can specify packages directly
+# using `use nix -p ocaml`.
 #
 use_nix() {
+  local NIX_NO_WATCH_DEFAULT="${NIX_NO_WATCH_DEFAULT:-false}"
```
zimbatm: uppercased variables are reserved to environment variables in this bash script

zimbatm:
I can't accept this PR as it would break backward-compatibility. If it was the reverse, `--keep-watches` option that disables the default then it would be fine.

kampka:
@zimbatm The main argument of this PR is that the current default is not great and probably also pretty much unexpected. If the code needs to stay compatible, the resulting code would only add unnecessary complexity compared to
```
use nix
watch_file shell.nix
```

There is virtually no gain, so I'm withdrawing this.
Thanks for looking through it though :+1:

zimbatm:
Yeah. Backward-compatibility is one of the core feature of direnv. It's the type of tool that you should be able to install and not think about it, just like `grep` or `sed`. So we are stuck with some less than ideal APIs some time.

PR #505, opened 2019-06-10
stdlib: venv and pyenv

eddies:
This is a follow-on to #429, rebased on master, with a couple updates:

* Only invoke python once to determine the Python version and whether venv is available (since invoking the python interpreter is relatively slow)
* Explicitly check for the existence of virtualenv (since it's not part of the standard library)
* Use pkgutil instead of import to check for the availability of venv and virtualenv so we don't output a multiline traceback to stderr if an import fails*
* Don't export VIRTUAL_ENV if neither venv nor virtualenv are available (for folks who set PS1 with the value of VIRTUAL_ENV to indicate that a virtual environment has been successfully activated, this would otherwise be quite misleading)

This PR also adds `/test/scenarios/inherited/.envrc` to .gitignore, which is generated by `direnv-test.bash`.

Finally, this PR also adds a commit to support [pyenv](/pyenv/pyenv) (like rbenv--in fact, pyenv was forked from it).

\* For example, in Python < 3.3, `python -c "import venv"` yields:
```
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named venv
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@
 /test/allow
 /test/direnv
 /test/config/direnv/allow
+/test/scenarios/inherited/.envrc
--- a/stdlib.go
+++ b/stdlib.go
@@ -425,9 +425,10 @@ const STDLIB = "#!/usr/bin/env bash\n" +
 	"\n" +
 	"# Usage: layout python <python_exe>\n" +
 	"#\n" +
-	"# Creates and loads a virtualenv environment under\n" +
+	"# Creates and loads a virtual environment under\n" +
 	"# \"$direnv_layout_dir/python-$python_version\".\n" +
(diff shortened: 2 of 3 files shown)


eddies:
This is a separate question that's not exactly to do with this PR, but since I was adding an entry for a test artifact to the .gitignore, I thought I'd ask: shouldn't `stdlib.go` also be in .gitignore (and deleted from the repo), since it's a generated file?

eddies:
@zimbatm anything I can do that would help get this PR merged?

zimbatm:
Sorry for the wait!

`stdlib.go` is checked in to allow users to `go get github.com/direnv/direnv` and it also makes the packager's life a bit easier.

Issue #500, opened 2019-06-04
Using layout_pipenv causes "direnv: error exit status 1" in new directory


metafeather:
**Describe the bug**
Using **`layout_pipenv`** causes `direnv: error exit status 1` in a new directory.

This is due to `layout_pipenv` raising the error when trying to determine if a virtualenv already exists, and `set -e` is activated in the .envrc sub-shell (see #245)

**To Reproduce**
1. `cd $(mktemp -d -t my_python_package)`
2. `touch Pipfile`
3. `echo "layout_pipenv" > .envrc`
4. `direnv allow`
5. `fail: direnv: error exit status 1`

**Expected behavior**
On first run of direnv allow a new virtualenv should be created and activated. On subsequent runs the virtualenv should be activated.

**Desktop (please complete the following information):**
 - OS: MacOS Mojave 10.14.5 using builtin Terminal.app
 - `bash --version`: GNU bash, version 3.2.57(1)-release (x86_64-apple-darwin18)
 - `direnv version (installed via Homebrew): 2.20.1`
 - `pipenv --version (installed via Homebrew): pipenv, version 2018.11.26`

**Additional context**
This is probably not the most common scenario since users probably create the Pipefile using pipenv which implicitly creates the virtualenv. However it is common enough (for my users at least) when checking out a project containing a Pipfile and .envrc.

metafeather:
For end users this issue can be worked around by using `! layout_pipenv` in the .envrc file instead. This causes `set -e` to ignore the error ([ref](https://stackoverflow.com/questions/11231937/bash-ignoring-error-for-a-particular-command))

The actual `error exit status 1` is from `layout_pipenv()` in stdlib.sh (#314):

`venv=$(pipenv --bare --venv 2>/dev/null)`

and this can also be suppressed by prefixing with ! with no other side-effects.

Users may alternatively  override `layout_pipenv()` in their own `.direnvrc` until fixed in a release.

JacobHayes:
@metafeather I was able to repro with the system bash 3.2, but it works fine with bash 5.0 from brew. Given `pipenv --bare --venv` returns 1 in both versions, it seems it _should_ be broken in bash 5.0.

Can you try out the version in https://github.com/direnv/direnv/pull/510 and see if that fixes your issue? You should be able to test with the following:

`/usr/bin/make build && mv /usr/local/bin/direnv /usr/local/bin/direnv.bak && mv direnv /usr/local/bin/direnv`

[...]

PR #510, opened 2019-07-02
Handle failing pipenv on missing venv and avoid an extra pipenv execution

JacobHayes:
Fixes `layout_pipenv`'s check for a venv path. For some reason, the error from `pipenv --bare --venv` when the venv doesn't exist does not exit the script in bash 5.0 (brew), but does in bash 3.2 (macos stock). Perhaps there needs to be some follow up investigation on if `set -e` is being enforced properly for bash 5.0.

This also removes an extra `pipenv --venv` lookup when it's already created, which will be every time except possibly the first, which should speed up loading a bit.

Resolves direnv/direnv#500

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -509,20 +509,18 @@ layout_anaconda() {
 # virtualenv from the Pipfile located in the same directory.
 #
 layout_pipenv() {
-  local venv
   PIPENV_PIPFILE="${PIPENV_PIPFILE:-Pipfile}"
   if [[ ! -f "$PIPENV_PIPFILE" ]]; then
     log_error "No Pipfile found.  Use \`pipenv\` to create a \`$PIPENV_PIPFILE\` first."
     exit 2
   fi
 
-  venv=$(pipenv --bare --venv 2>/dev/null)
+  VIRTUAL_ENV=$(pipenv --venv 2>/dev/null ; true)
 
-  if [[ -z $venv || ! -d $venv ]]; then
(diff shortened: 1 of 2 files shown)


stdlib.sh
```diff
@@ -509,21 +509,19 @@ layout_anaconda() {
...
 #
 layout_pipenv() {
-  local venv
   PIPENV_PIPFILE="${PIPENV_PIPFILE:-Pipfile}"
   if [[ ! -f "$PIPENV_PIPFILE" ]]; then
     log_error "No Pipfile found.  Use \`pipenv\` to create a \`$PIPENV_PIPFILE\` first."
     exit 2
   fi
 
-  venv=$(pipenv --bare --venv 2>/dev/null)
```
zimbatm: what does `--bare` do?

JacobHayes: Yeah, I hesitated removing that for any confusion. 😅

`--bare` prints ["Minimal output"](https://docs.pipenv.org/en/latest/#cmdoption-pipenv-bare), but `--venv` seems to not have any [conditional bare output](https://github.com/pypa/pipenv/blob/master/pipenv/cli/command.py#L147) (nor the code above) - either outputting only the path to stdout if exists or an error to stderr if it does not.

Since we previously only had `--bare` on the first `venv` definition (not the one that was actually exported) combined with the `[ ! d ...]` check, it seemed safe to remove. I'm happy to add it back though if you'd prefer.

zimbatm: Okay, thanks for the explanation!

PR #511, opened 2019-07-03
shell: fix elvish hook

dywedir:
TODO: for elvish 0.12+ we should check for `$nil`

### DIFF
--- a/shell_elvish.go
+++ b/shell_elvish.go
@@ -17,10 +17,10 @@ func (elvish) Hook() (string, error) {
 		if (> (count $m) 0) {
 			m = (explode $m)
 			keys $m | each [k]{
-				if (==s $k 'null') {
-					unset-env $k
-				} else {
+				if $m[$k] {
 					set-env $k $m[$k]
+				} else {
+					unset-env $k
 				}
 			}
 		}


zimbatm:
thanks, do you mind adding some tests to reproduce the issue?

dywedir:
This is more of a case of misleading code than a bug.
The only example that I can give is setting a variable named `null`.
Do you think it's worth adding?

zimbatm:
I don't know enough of evlish to really understand the problem. If there is a difference between two versions of elvish it's probably good to add some tests don't you think? What would happen in elvish 0.12?

shell_elvish.go
```diff
@@ -17,7 +17,7 @@ func (elvish) Hook() (string, error) {
 		if (> (count $m) 0) {
 			m = (explode $m)
 			keys $m | each [k]{
-				if (==s $k 'null') {
+				if (==s $m[$k] '') {
```
mmlb: I've changed my hook locally to  `if $m[$k]` which I think allows for an env var to be set to `''` which is what makes more sense I think. bash/elvish allow an env to be set to empty string and can tell the difference between set and empty vs not set. direnv also handles correctly.

mmlb:
@zimbatm : 
v12:
```
echo '{"foo":null}' | from-json
▶ [&foo='']
```
git master:
```
~> echo '{"foo":null}' | from-json
▶ [&foo=$nil]
```

dywedir:
@zimbatm sorry, should have given more information.

I see two problems in the current version:
- We use key, instead of value
- We check for `null` instead of an empty string

So, we `unset-env` only when we have to set env named `null`, which is definitely wrong.

zimbatm:
The shell is supposed to set empty environment variables. It's not a commonly used feature but it's not exactly the same thing as an unset environment variable.
```
$ export FOO=
$ env | grep FOO
FOO=
```

dywedir:
Thanks @mmlb @zimbatm!
I'll update the PR when I get back to my PC.

dywedir:
Changed to `if $m[$k]`

mmlb:
This works, but the issue is that direnv + elvish will only work on unyet released elvish versions. I'm not sure how @zimbatm wants to deal with that...

1. try it, see if anyone complains (it might be that since elvish  isn't stable, everyone uses master?)
2. handle non-nil/nil available in the hook by detecting absence of `$nil`?

zimbatm:
From my perspective, elvish is not a stable and widely used shell. I try really hard to keep direnv backward-compatible but don't mind relaxing the requirements for elvish because of that. It's also possible that you are both the only users for direnv+elvish.

With elvish 12 it's not possible to distinguish between empty and null strings. It means that it's not possible to implement the protocol correctly. Empty strings are a rare occurence but once a user depends on them it's going to be annoying to debug this weird corner-case.

I would suggest to change the hook to only support the correct implementation and break if nulls are not supported correct, with a message telling the user to use elvish v13+

dywedir:
Should we wait until `elvish` cuts a new release?

zimbatm:
:+1:  sounds like a good idea to me!

zimbatm:
should this PR get revisited now?

mmlb:
@zimbatm evlish 0.13 is available now so sounds like you can merge this now.

zimbatm:
alright :)

PR #514, opened 2019-07-19
shell: improve zsh hook

laggardkernel:
Prepend but not append `_direnv_hook` to make sure environment variables are changed before any other hooks, in case the vars are used by other hooks. This is what bash does with `PROMPT_COMMAND` in `shell_bash.go`.

Besides, add `_direnv_hook` into `chpwd` hook, which is triggered once working dir is changed by ZLE widget but not command `cd` in ZSH.

### DIFF
--- a/shell_zsh.go
+++ b/shell_zsh.go
@@ -11,7 +11,11 @@ _direnv_hook() {
 }
 typeset -ag precmd_functions;
 if [[ -z ${precmd_functions[(r)_direnv_hook]} ]]; then
-  precmd_functions+=_direnv_hook;
+  precmd_functions=( _direnv_hook ${precmd_functions[@]} )
+fi
+typeset -ag chpwd_functions;
+if [[ -z ${chpwd_functions[(r)_direnv_hook]} ]]; then
+  chpwd_functions=( _direnv_hook ${chpwd_functions[@]} )
 fi
 `
 


zimbatm:
The first part is legit.

The second part I don't know because I don't know enough about `chpwd`. Isn't the path always changed before showing a new prompt?

laggardkernel:
> Isn't the path always changed before showing a new prompt?

@zimbatm No. `precmd` is triggered on prompt rendering. `chpwd` is triggered once working directory is changed. And the `precmd` is NOT triggered when working dir is changed by a **ZLE widget**, like [this one](https://github.com/urbainvaes/fzf-marks/blob/master/fzf-marks.plugin.zsh#L159).

[9.3 Special Functions fronm zsh manual](http://zsh.sourceforge.net/Doc/Release/Functions.html)
> chpwd
>     Executed whenever the current working directory is changed.
> 
> precmd
>     Executed before each prompt. Note that precommand functions are not re-executed simply because the command line is redrawn, as happens, for example, when a notification about an exiting job is displayed.

In my opinion, `chpwd` is what we really need, `precmd` may be redundant in this situation. Cause `cd <path>` triggers `chpwd` as well. The reason I keep it hooked on `precmd` is that, the prompt function may use variables `export`ed by `direnv`, I need to make sure the variables are set before the prompt function by prepending `_direnv_hook` before it.

In my case, `PYENV_VERSION` is set in `.envrc`, and the prompt function `prompt_spaceship_precmd` uses this var to display the python environment.

In bash, `PROMPT_COMMAND` is equivalent to `precmd`. There's no `chpwd` equivalent in Bash. Fish's event `fish_prompt` is equivalent to zsh's `precmd`, there's no `chpwd` equivalent either.

zimbatm:
awesome, thanks for the explanation!

Yeah I think we need both for direnv to work as expected. For example if the `.envrc` is changed, the user should get a reload on the next prompt even if the current directory is not changed.

shell_zsh.go
```diff
@@ -11,7 +11,11 @@ _direnv_hook() {
 }
 typeset -ag precmd_functions;
 if [[ -z ${precmd_functions[(r)_direnv_hook]} ]]; then
```
zimbatm: just thinking about it; if _direnv_hook is already in the array, it should be removed and pushed back. That way it would increase the likelihood that _direnv_hook is in the front of the array.

laggardkernel: An easy solution would be defining the hook function with `typeset -agU precmd_functions`. This forces only the 1st occurrence of an item is kept in the array.

```zsh
typeset -ag precmd_functions
# re-prepend the item into the array
precmd_functions=( _direnv_hook ${precmd_functions[@]} )
```

But do we really need it? Most users init `direnv` with `eval "$(direnv hook zsh)"`.  I don't think a common user will meddle with `precmd` hook manually. The code here is enough to make sure `_direnv_hook` is hooked on `precmd` 1st.

zimbatm: It would be an improvement but not part of the original PR's intent so leaving it for now. Thanks!

vincentbernat:
I am only using chpwd and it seems it is enough to make environment variables available for the prompt. Maybe we could remove precmd (which is far more expensive)?

PR #524, opened 2019-10-07
Watch more than one file at a time.

allenap:
The rationale is that when watching a large number of files it can take a noticeable amount of time to add all the watches.

This also removes the code that replaces leading tildes with `$HOME` on the filename passed to watch_file. A leading tilde is typically replaced by the shell. By doing it a second time it complicates watching files that do actually begin with a tilde. But I'm happy to put it back in (and make it work for multiple files) if removing it is a blocker.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+UNRELEASED
+==========
+
+  * CHANGE: Watch multiple files in a single invocation of `watch_file` / `direnv watch`.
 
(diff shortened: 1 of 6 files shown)


zimbatm:
thanks, great work on this PR

zimbatm:
I also agree about the tilda replacement. Hopefully it won't break too many existing scripts.

allenap:
@zimbatm Thanks!

PR #525, opened 2019-10-16
layout_pyenv: support multiple python versions

eddies:
Adding to #505, this adds support for specifying multiple Python versions, e.g.:

```bash
layout pyenv 3.6.9 2.7.16
```

Pyenv supports specifying multiple pythons, e.g. `pyenv shell 3.6.9 2.7.16`, but `layout pyenv` didn't (till this PR). This also sets the PYENV_VERSION environment variable for multiple pythons as pyenv would (multiple versions separated by a colon, e.g. `3.6.9:2.7.16`).

This pretty much is just the previous layout_pyenv function run in a loop for each Python version specified.

The motivation for this was that I had Python 3 projects where I was using direnv to manage my Python 3 environment, but needed to run gcloud (Google Cloud SDK) which requires Python 2. 

Finally, this fixes a bug/edge case for virtualenv: previously, we were just invoking "virtualenv", which assumed it was available to the os/system/default python. But if virtualenv was only installed for a particular python, e.g. 2.7.16, which was *not* the default python, then you'd see something like:

```
pyenv: virtualenv: command not found

The `virtualenv' command exists in these Python versions:
  2.7.16
```

Now we invoke virtualenv in the same way we do venv--via the requested python version. Then, if virtualenv isn't available, it makes more sense: you need to `pip install virtualenv` for that python version and then direnv can happily use that.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -455,7 +455,7 @@ layout_python() {
         ;;
       "virtualenv")
         if [[ ! -d $VIRTUAL_ENV ]]; then
-          virtualenv "--python=$python" "$@" "$VIRTUAL_ENV"
+          $python -m virtualenv "$@" "$VIRTUAL_ENV"
         fi
         ;;
       *)
@@ -540,7 +540,7 @@ layout_pipenv() {
   export VIRTUAL_ENV
 }
 
-# Usage: layout pyenv <python version number>
+# Usage: layout pyenv <python version number> [<python version number> ...]
 #
 # Example:
 #
@@ -550,17 +550,27 @@ layout_pipenv() {
 # "$direnv_layout_dir/python-$python_version".
 #
 layout_pyenv() {
-  local python_version=$1
-  local pyenv_python
-  pyenv_python=$(pyenv root)/versions/${python_version}/bin/python
-  if [[ -x "$pyenv_python" ]]; then
-    if layout_python "$pyenv_python"; then
-      export PYENV_VERSION=$python_version
+  unset PYENV_VERSION
+  # layout_python prepends each python version to the PATH, so we add each
+  # version in reverse order so that the first listed version ends up
+  # first in the path
+  local i
+  for ((i = $#; i > 0; i--)); do
+    local python_version=${!i}
+    local pyenv_python
+    pyenv_python=$(pyenv root)/versions/${python_version}/bin/python
+    if [[ -x "$pyenv_python" ]]; then
+      if layout_python "$pyenv_python"; then
+        # e.g. Given "use pyenv 3.6.9 2.7.16", PYENV_VERSION becomes "3.6.9:2.7.16"
+        PYENV_VERSION=${python_version}${PYENV_VERSION:+:$PYENV_VERSION}
+      fi
+    else
+      log_error "pyenv: version '$python_version' not installed"
+      return 1
     fi
-  else
-    log_error "pyenv: version '$python_version' not installed"
-    return 1
-  fi
+  done
+
+  [[ -n "$PYENV_VERSION" ]] && export PYENV_VERSION
 }
 
 # Usage: layout ruby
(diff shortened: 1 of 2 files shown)


eddies:
One thing that tripped me up: in the little loop I added to run through python versions in reverse, I was using `i` (as one does) as my increment variable and it was getting stomped on in my invocation of `layout_python`. Turned out that was from the for loop in `path_add`. I considered 1) declaring `i` as local in `path_add`, 2) invoking `layout_python` in sub-shell, but settled on renaming my counter to `j` :P as the most straightforward.

zimbatm:
thanks, I fixed the `i` variable leak in c1470918081021a96aeedd6c14b8f29b04144ce6 and will be reviewing the PR next

stdlib.sh
```diff
@@ -551,17 +551,26 @@ layout_pipenv() {
 # "$direnv_layout_dir/python-$python_version".
 #
 layout_pyenv() {
-  local python_version=$1
-  local pyenv_python
-  pyenv_python=$(pyenv root)/versions/${python_version}/bin/python
-  if [[ -x "$pyenv_python" ]]; then
-    if layout_python "$pyenv_python"; then
-      export PYENV_VERSION=$python_version
+  unset PYENV_VERSION
```
zimbatm: can you add `local j` to avoid leaking the variable as well?

stdlib.sh
```diff
@@ -551,17 +551,26 @@ layout_pipenv() {
...
+  # layout_python prepends each python version to the PATH, so we add each
+  # version in reverse order so that the first listed version ends up
+  # first in the path
+  for ((j = $#; j >= 1; j--)); do
```
zimbatm: What is the reason to iterate the array backwards? Is it because of how PYENV_VERSION is constructed?

If not strictly required, I find this more readable:
```
local python_version
for python_version in "$@"; do
```

stdlib.sh
```diff
@@ -551,17 +551,26 @@ layout_pipenv() {
...
+    local python_version=${!j}
+    local pyenv_python
+    pyenv_python=$(pyenv root)/versions/${python_version}/bin/python
+    if [[ -x "$pyenv_python" ]]; then
+      if layout_python "$pyenv_python"; then
+        # e.g. Given "use pyenv 3.6.9 2.7.16", PYENV_VERSION becomes "3.6.9:2.7.16"
+        [[ -z "$PYENV_VERSION" ]] && PYENV_VERSION=$python_version || PYENV_VERSION="${python_version}:$PYENV_VERSION"
```
zimbatm: There is a shorter notation that can be used here:
```
PYENV_VERSION=${python_version}${PYENV_VERSION:+:$PYENV_VERSION}
```

stdlib.sh
eddies: If a user specifies multiple Python versions, e.g. `layout pyenv 3.6.9 2.7.16` and we iterate normally, layout_python will prepend (via path_add) python 3.6.9 and then prepend 2.7.16, such that 2.7.16 will be the first Python on the path, which is the reverse of what we want (and the reverse of what pyenv would do on its own).

I actually had originally written the loop as you did above before I noticed the issue.

stdlib.sh
eddies: Nice! I wasn't familiar with this, I had to look it up. For the curious: https://www.tldp.org/LDP/abs/html/parameter-substitution.html#PARAMALTV

stdlib.sh
eddies: done!

eddies:
rebased on master (c147091) and made the requested changes (`local j`, shorter way of constructing `PYENV_VERSION`) except for rewriting the for-loop (since I believe we need to iterate in reverse order).

stdlib.sh
zimbatm: okay, thanks for the explanation :)

PR #526, opened 2019-10-22
Pin nixpkgs to current NixOS 19.09 channel

curiousleo:
I suspect that people are now switching from 19.03 to 19.09. This change just means there's less to download when running `nix-shell` in the project directory initially because the diff between one's own system and the pinned version is smaller.


zimbatm:
thanks!

PR #534, opened 2019-10-27
Add use_asdf

vic:
This helps loading all environment variables as managed by [asdf](https://asdf-vm.com) plugins.

This patch will help people using asdf that have encountered problems with asdf shims. For example, some commands (afaik some npm executables) try to find themselves in PATH and
do some stuff with their location. However since asdf uses shims (that just delegate to the actual versioned command), said npm executables become very slow.

Using direnv and `use asdf` can help exporting the environment managed by asdf directly into the user shell, uloading and loading when needed. This will prevent shims from being used as PATH will be updated accordingly (and any other environment variable managed by the asdf plugin)


vic:
For context, check the README of https://github.com/asdf-community/asdf-direnv

PR #562, opened 2020-01-26
stdlib: add source_url function

zimbatm:
Provide an easy way to extend the stdlib with scripts from the Internet.
While still enforcing some level of security.
    
Eg:
    
    source_url "https://raw.githubusercontent.com/nmattia/sorri/bff9e44734b19d5c65534c9bfd10bdf3be90b623/sorri" "sha256-CMbjOcSXWtTqSW0jKBPWWPBMBiOvJsh2vnFJoQJ6yFM="
    
To find out the hash, get the immutable script URL, then run `direnv fetchurl <url>` to get back the hash.

### DIFF
--- a/man/direnv-fetchurl.1
+++ b/man/direnv-fetchurl.1
@@ -14,7 +14,8 @@ content of the retrieved file.
 
 .PP
 This has been introduced to avoid a dependency on \fB\fCcurl\fR or \fB\fCwget\fR, while also
-promoting a more secure way to fetch data from the Internet.
+promoting a more secure way to fetch data from the Internet. Use this instead
+of \fB\fCcurl <url> | sh\fR\&.
 
 .PP
 This command has two modes of operation:
@@ -29,7 +30,8 @@ implication of this design is that URLs must be stable and always return the
 same content.
 
 .PP
-Downloaded files are marked as read\-only and executable.
+Downloaded files are marked as read\-only and executable so it can also be used
+to fetch and execute static binaries.
 
 .SH OPTIONS
 .PP
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -108,6 +108,29 @@ Loads another \fB\fC\&.envrc\fR if found when searching from the parent director
 .PP
 NOTE: the other \fB\fC\&.envrc\fR is not checked by the security framework.
 
+.SS \fB\fCsource\_url <url> <integrity\-hash>\fR
+.PP
+Loads another script from the given \fB\fCurl\fR\&. Before loading it it will check the
+integrity using the provided \fB\fCintegrity\-hash\fR\&.
+
+.PP
+To find the value of the \fB\fCintegrity\-hash\fR, call \fB\fCdirenv fetchurl <url>\fR and
+extract the hash from the outputted message.
+
+.PP
+See also \fB\fCdirenv\-fetchurl(1)\fR for more details.
+
+.SS \fB\fCfetchurl <url> [<integrity\-hash>]\fR
+.PP
+Fetches the given \fB\fCurl\fR onto disk and outputs it's path location on stdout.
+
+.PP
+If the \fB\fCintegrity\-hash\fR argument is provided, it will also check the integrity
+of the script.
+
+.PP
+See also \fB\fCdirenv\-fetchurl(1)\fR for more details.
+
 .SS \fB\fCdirenv\_apply\_dump <file>\fR
 .PP
 Loads the output of \fB\fCdirenv dump\fR that was stored in a file.
--- a/man/direnv-stdlib.1.md
+++ b/man/direnv-stdlib.1.md
@@ -81,6 +81,25 @@ Loads another `.envrc` if found when searching from the parent directory up to /
 
 NOTE: the other `.envrc` is not checked by the security framework.
 
+### `source_url <url> <integrity-hash>`
+
+Loads another script from the given `url`. Before loading it it will check the
+integrity using the provided `integrity-hash`.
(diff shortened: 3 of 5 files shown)


cmd_fetchurl.go
```diff
@@ -0,0 +1,103 @@
...
+		}
+	}
+
+	tmpfile, err := ioutil.TempFile("", "fetchurl")
+	if err != nil {
+		return err
+	}
+	defer os.Remove(tmpfile.Name()) // clean up
+
+	resp, err := http.Get(url)
```
golangcibot: G107: Potential HTTP request made with variable url (from `gosec`)

sri/sri.go
```diff
@@ -0,0 +1,83 @@
+// Package sri implements helper functions to calulate SubResource Integrity
+// hashes.
+// https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity
+package sri
+
+import (
+	"crypto/sha256"
+	"crypto/sha512"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion
	"crypto/sha512"
	b64 "encoding/base64"
```

cmd_fetchurl.go
```diff
@@ -0,0 +1,103 @@
+package main
+
+import (
+	"fmt"
+	"io"
+	"io/ioutil"
+	"os"
```
golangcibot: File is not `gofmt`-ed with `-s` (from `gofmt`)
```suggestion

```

sri/sri.go
```diff
@@ -0,0 +1,83 @@
+// Package sri implements helper functions to calulate SubResource Integrity
```
golangcibot: `calulate` is a misspelling of `calculate` (from `misspell`)
```suggestion

```

sri/sri.go
```diff
@@ -0,0 +1,83 @@
...
+	default:
+		panic("unsupported SRI algo")
+	}
+
+	return Writer{w,algo,h}
+}
+
+func (w Writer) Write(b []byte) (int, error) {
+	// TODO: maybe only write to the hash if the copy to w.w succeeded
+	w.h.Write(b)
```
golangcibot: Error return value of `w.h.Write` is not checked (from `errcheck`)

Mic92:
Where does this cache the urls?

cmd_fetchurl.go
```diff
@@ -0,0 +1,102 @@
...
+	// validate the SRI hash if it exists
+	if len(args) > 2 {
+		sriHash = args[2]
+		algo, err = sri.GetAlgo(sriHash)
+		if err != nil {
+			return err
+		}
+
+		// shortcut if the cache already has the file
+		cacheFile = filepath.Join(config.CacheDir, "fetchurl", sriHash)
```
Mic92: I see.

Mic92:
This is a nice solution for projects. I would make `direnv fetchurl` also print copy'n'paste snippets for inclusion in .envrc

cmd_fetchurl.go
```diff
@@ -0,0 +1,102 @@
...
+	if _, err = io.Copy(f, r); err != nil {
+		return err
+	}
+
+	// Make the file read-only and executable
+	if err = f.Chmod(os.FileMode(0500)); err != nil {
+		return err
+	}
+	
+	_, err = fmt.Println(cacheFile)
```
Mic92: One could check if stdout is a pipe or a tty to decide wether to print the instructions on stdout or stderr.

zimbatm:
I have one remaining issue: the SRI encoding can contain `/` characters because of base64. I was hoping to re-use the same key for the filesystem cache but I need to find another encoding there. Which is a shame because now there isn't a 1:1 correspondence with the code anymore. Technically it works on POSIX but it's not very elegant.

cmd_fetchurl.go
Mic92: I would just escape `/` with `-` to avoid `/` in the the filename.

sri/sri.go
```diff
@@ -0,0 +1,83 @@
...
+)
+
+// Algo is a supported hashing algorithm
+type Algo string
+
+const (
+	// SHA256 algo
+	SHA256 = Algo("sha256")
+	// SHA384 algo
+	SHA384 = Algo("sha384")
```
Mic92: I have never seen sha384 in the wild. Any good reason to support it?

zimbatm: Mainly because the spec lists all the supported algos and it's in that list. I agree that it's not really common.

jeci-sarl:
I think there is a security issue with the cache. In the current implementation, the content is never check if the hash already exist in cache.

``` bash
direnv fetchurl https://hacker.com/bad/script.sh 
	=> #has123456789

#.envrc
source_url https://direnv.net/extension/script.sh #has123456789
```

I think we must always check hash when allowing script (in `direnv allow`), to prevent this kind of attack.

zimbatm:
It's a bit tricky to build a security model when it's assumed that the user is compromised. At that point pretty much anything can happen. In the scenario that is outlined, the attacker much coordinate both a call to `direnv fetchurl https://hacker.com/bad/script.sh`, and also push a compromised `source_url https://direnv.net/extension/script.sh #has123456789`. If these conditions are not met the attack fails.

jeci-sarl:
Yes you're right.

cmd_fetchurl.go
zimbatm: thanks, I used `_` instead so it can be distinguished from the `<algo>-<hash>` separator.

PR #568, opened 2020-01-28
Simplify direnv_load and make it work even when the command crashes.

### LINKED ISSUE #569: direnv_load blocks when command fails
**Describe the bug**
A call to, for example, `direnv_load true` would cause direnv to hang. This is because the command does not call `direnv dump`, hence nothing ever writes to the far end of the named pipe (made by `mkfifo`) that had been created to receive the output from the given command, `true`.

Another example:
```bash
direnv_load bash -c 'do_something_that_breaks && direnv dump'
```

In this instance we've remembered to call `direnv dump`, but it is never called because `do_something_that_breaks` returns non-zero.

**To Reproduce**
```bash
mkdir foo
echo 'direnv_load true' > foo/.envrc
direnv allow foo
cd foo  # --> hangs, use ctrl-\ to get out.
```

**Expected behavior**
It should show an error and not block.

**Environment**
 - OS: *nix
 - Shell: bash
 - Direnv version: 2.21.1

**Additional context**
(none)

allenap:
Fixes #569.

Previously a call to, for example, `direnv_load true` would cause direnv to hang. This is because the command didn't include a call to `direnv dump`, hence nothing ever wrote to the far end of the named pipe that had been created to receive the output from the given command, `true`.

Another example:

```
direnv_load bash -c 'do_something_that_breaks && direnv dump'
```

In this instance we've rememebered to call `direnv dump`, but it is never called because `do_something_that_breaks` returns non-zero.

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -261,49 +261,41 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"# process - cause that process to run \"direnv dump\" and then wrap\n" +
 	"# the results with direnv_load.\n" +
 	"#\n" +
+	"# shellcheck disable=SC1090\n" +
 	"direnv_load() {\n" +
 	"  # Backup watches in case of `nix-shell --pure`\n" +
 	"  local prev_watches=$DIRENV_WATCHES\n" +
-	"  local prev_dump_file_path=${DIRENV_DUMP_FILE_PATH:-}\n" +
+	"  local temp_dir output_file script_file exit_code\n" +
 	"\n" +
-	"  # Create pipe\n" +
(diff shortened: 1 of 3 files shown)


zimbatm:
thanks, I will try to go over thing over the weekend.

On top of the issue you mentioned, I have also seen users backup their environment to a cache and then use `direnv_load /path/to/cache` to restore it. It looks like I was too bold when I did the changes and will have to duplicate that function.

zimbatm:
this is brilliant

zimbatm:
I had to give it a try myself ( https://github.com/direnv/direnv/pull/588 ) before I could really appreciate your solution

Issue #566, opened 2020-01-27
Regression: fail if uninitialized variables are being used breaks a lot of scripts


rjcoelho:
Hi to all, https://github.com/direnv/direnv/commit/c70f7cddca922e830c6225833e4d736327d8bf7e breaks a lot of scripts. I'm kerl activating and it broke it. We should be aware of this when making such changes. 

Please add it as a direvn's op-in, since AFAIK that is the default behaviour for bash/zsh shell.

zimbatm:
Can you give me a list of things that are breaking so I can gauge how much of a problem it is?

Mellbourn:
@zimbatm No, we can't give you such a list, because this setting, `set -u`, affects the running of the scripts that we as _users_ of `direnv` have written for `direnv` to run.
That is why this is "breaks a lot of scripts"

halostatue:
@zimbatm The problem isn’t as much shell scripts that are part of the standard library, but the cascading effect that `set -u` has on shell scripts that are _sourced_ (e.g., kerl, kiex, and chruby). There is a workaround, but it requires people to update their `.direnvrc`. For example, you’d have to change the first `use_kerl` implementation in the [Erlang](https://github.com/direnv/direnv/wiki/Erlang) entry on the Wiki to be this:

```sh
use_kerl () {
    set +u
    . $KERL_DEFAULT_INSTALL_DIR/$1/activate
    set -u
}
```

The third-party tools here that require `source`ing instead of calling an external sub-shell tool are broken by the stdlib use of `set -u`.

halostatue:
This could be "fixed" if we added an `unsafe_source` function to the stdlib that basically does the following:

```sh
unsafe_src () {
    set +u
    source "$1"
    set -u
}
```

zimbatm:
Okay, thanks for sharing the use-cases with me.

A quick workaround is to add `set +u` in the top of the direnvrc while I think more about this.

My hope was to update everyone to use safer scripts by default but I understand that it can be quite disruptive to your workflow.

halostatue:
I’m adding this to my `.direnvrc`:

```sh
unsafe() {
  local res

  set +u

  "${@}"

  res=${?}
  set -u

  return "${res}"
}
```

It would be used as `unsafe source path-to-file`. It will work with any single-line command. I’m choosing to do this because `chruby` also executes in an “unsafe” manner, so I can also do `unsafe chruby ${1}`, as an example.

zimbatm:
[...]

PR #572, opened 2020-01-30
Add strict_env and unstrict_env

halostatue:
Allows the gradual introduction of `set -euo pipefail` into envrc processing,
as discussed in #566, as well as an escape hatch for commands that must be run
in the envrc context that do not meet `set -euo pipefail` requirements.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -12,7 +12,6 @@ shopt -s gnu_errfmt
 shopt -s nullglob
 shopt -s extglob
 
-
 # NOTE: don't touch the RHS, it gets replaced at runtime
 direnv="$(command -v direnv)"
 
@@ -26,6 +25,90 @@ direnv_config_dir="${DIRENV_CONFIG:-${XDG_CONFIG_HOME:-$HOME/.config}/direnv}"
 # of a .envrc evaluation context. It is ignored by the direnv diffing
 # algorithm and so it won't be re-exported.
 export DIRENV_IN_ENVRC=1
+
+__env_strictness() {
+  local mode tmpfile old_shell_options
+  local -i res
+
+  tmpfile=$(mktemp)
+  res=0
+  mode="$1"
+  shift
+
+  set +o | grep 'pipefail\|nounset\|errexit' > "$tmpfile"
+  old_shell_options=$(< "$tmpfile")
+  rm -f tmpfile
+
+  case "$mode" in
+  strict)
+    set -o errexit -o nounset -o pipefail
+    ;;
+  unstrict)
+    set +o errexit +o nounset +o pipefail
+    ;;
+  *)
+    log_error "Unknown strictness mode '${mode}'."
+    exit 1
+    ;;
+  esac
+
+  if (($#)); then
+    "${@}"
+    res=$?
+    eval "$old_shell_options"
+  fi
+
+  # Force failure if the inner script has failed and the mode is strict
+  if [[ $mode = strict && $res -gt 0 ]]; then
+    exit 1
+  fi
+
+  return $res
+}
+
+# Usage: strict_env [<command> ...]
+#
+# Turns on shell execution strictness. This will force the .envrc
+# evaluation context to exit immediately if:
+#
+# - any command in a pipeline returns a non-zero exit status that is
+#   not otherwise handled as part of `if`, `while`, or `until` tests,
+#   return value negation (`!`), or part of a boolean (`&&` or `||`)
+#   chain.
+# - any variable that has not explicitly been set or declared (with
+#   either `declare` or `local`) is referenced.
+#
+# If followed by a command-line, the strictness applies for the duration
+# of the command.
+#
+# Example:
+#
+#    strict_env
+#    has curl
+#
+#    strict_env has curl
+strict_env() {
+  __env_strictness strict "$@"
+}
+
+# Usage: unstrict_env [<command> ...]
+#
+# Turns off shell execution strictness. If followed by a command-line, the
+# strictness applies for the duration of the command.
+#
+# Example:
+#
+#    unstrict_env
+#    has curl
+#
+#    unstrict_env has curl
+unstrict_env() {
+  if (($#)); then
+    __env_strictness unstrict "$@"
+  else
+    set +o errexit +o nounset +o pipefail
+  fi
+}
 
 # Usage: direnv_layout_dir
 #
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -474,6 +474,78 @@ Checks that the direnv version is at least old as \fB\fCversion\_at\_least\fR\&.
 be useful when sharing an \fB\fC\&.envrc\fR and to make sure that the users are up to
 date.
 
+.SS \fB\fCstrict\_env [<command> ...]\fR
+.PP
+Turns on shell execution strictness. This will force the .envrc
+evaluation context to exit immediately if:
+.IP \(bu 2
+any command in a pipeline returns a non\-zero exit status that is not
+otherwise handled as part of \fB\fCif\fR, \fB\fCwhile\fR, or \fB\fCuntil\fR tests,
+return value negation (\fB\fC!\fR), or part of a boolean (\fB\fC\&\&\fR or \fB\fC||\fR)
+chain.
+.IP \(bu 2
+any variable that has not explicitly been set or declared (with
+either \fB\fCdeclare\fR or \fB\fClocal\fR) is referenced.
+
+.PP
+If followed by a command\-line, the strictness applies for the duration
+of the command.
+
+.PP
+Example (Whole Script):
+
+.PP
+.RS
+
+.nf
+strict\_env
+has curl
+
+.fi
+.RE
+
+.PP
+Example (Command):
+
+.PP
+.RS
+
+.nf
+strict\_env has curl
+
+.fi
+.RE
+
+.SS \fB\fCunstrict\_env [<command> ...]\fR
+.PP
+Turns off shell execution strictness. If followed by a command\-line, the
+strictness applies for the duration of the command.
+
+.PP
+Example (Whole Script):
+
+.PP
+.RS
+
+.nf
+unstrict\_env
+has curl
+
+.fi
+.RE
+
+.PP
+Example (Command):
+
+.PP
+.RS
+
+.nf
+unstrict\_env has curl
+
+.fi
+.RE
+
 .SH COPYRIGHT
 .PP
 MIT licence \- Copyright (C) 2019 @zimbatm and contributors
--- a/man/direnv-stdlib.1.md
+++ b/man/direnv-stdlib.1.md
@@ -328,6 +328,45 @@ Checks that the direnv version is at least old as `version_at_least`. This can
 be useful when sharing an `.envrc` and to make sure that the users are up to
 date.
 
+### `strict_env [<command> ...]`
+
+Turns on shell execution strictness. This will force the .envrc
+evaluation context to exit immediately if:
+
+- any command in a pipeline returns a non-zero exit status that is not
+  otherwise handled as part of `if`, `while`, or `until` tests,
+  return value negation (`!`), or part of a boolean (`&&` or `||`)
+  chain.
+- any variable that has not explicitly been set or declared (with
+  either `declare` or `local`) is referenced.
+
+If followed by a command-line, the strictness applies for the duration
+of the command.
+
+Example (Whole Script):
+
+    strict_env
+    has curl
+
+Example (Command):
+
+    strict_env has curl
+
+#### `unstrict_env [<command> ...]`
+
+Turns off shell execution strictness. If followed by a command-line, the
+strictness applies for the duration of the command.
+
+Example (Whole Script):
+
+    unstrict_env
+    has curl
+
+Example (Command):
+
+    unstrict_env has curl
+
+
 COPYRIGHT
 ---------
 
--- a/stdlib.go
+++ b/stdlib.go
@@ -15,7 +15,6 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"shopt -s nullglob\n" +
 	"shopt -s extglob\n" +
 	"\n" +
-	"\n" +
 	"# NOTE: don't touch the RHS, it gets replaced at runtime\n" +
 	"direnv=\"$(command -v direnv)\"\n" +
 	"\n" +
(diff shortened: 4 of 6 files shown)


halostatue:
I haven’t yet tested these, but they are based on the functions discussed in #566, which I _have_ tested.  I’ve gone ahead and introduced `strict_env` as you mentioned, and I’ve made it so that things can be introduced gradually on a command-by-command basis or globally.

These _should_ probably be made a little more robust such that the following doesn’t do weird things:

```sh
strict_env
strict_env command
```

This will have the counter-intuitive result of `set +euo pipefail` being called. This can be fixed with a bit of checking by calling `set -o` and comparing the values `pipefail`, `errexit`, and `nounset`…but that’s a little harder to do and won’t work _quickly_ with Bash 3 because of the lack of associative arrays.

I can also make them _not_ symmetrical so that `strict_env` always sets `set +euo pipefail` and `unstrict_env` only works with a provided command (it does nothing if `$#` is 0).

stdlib.sh
```diff
@@ -26,6 +26,68 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
...
+  set -euo pipefail
+
+  if (( $# )); then
+    local res
+
+    "${@}"
+
+    res=$?
+
+    set +euo pipefail
```
zimbatm: Do you know if it's possible to detect and backup the original state of `eu` and `pipefail` so they can be restored here?

zimbatm:
https://github.com/direnv/direnv/pull/573 will be complementary to this PR.

stdlib.sh
yermulnik: Wouldn't this help (`help set`)?
```
The current set of flags may be found in $-.
```

halostatue: `$-` doesn’t contain the `set -o pipefail` flag. That can be obtained (and parsed) via `set -o`.

halostatue: @zimbatm Yes, it is. If I were to do that, I’d probably want to use `set -o errexit -o nounset -o pipefail` to set, because that’s the corresponding values from `set -o`.

halostatue: OK. The latest version does this mostly correctly. I have tested this in an interactive bash shell (the output of `set -o` has been trimmed to the relevant settings):

```sh
bash-3.2$ source stdlib.sh
bash-3.2$ set -o
errexit        	off
nounset        	off
pipefail       	off

bash-3.2$ strict_env
bash-3.2$ set -o
errexit        	on
nounset        	on
pipefail       	on

bash-3.2$ strict_env echo HELLO
HELLO
bash-3.2$ set -o
errexit        	off
nounset        	on
pipefail       	on

bash-3.2$ unstrict_env echo HELLO
HELLO

bash-3.2$ set -o
errexit        	off
nounset        	on
pipefail       	on

bash-3.2$ unstrict_env
bash-3.2$ set -o
errexit        	off
nounset        	off
pipefail       	off
```

The one piece that’s _broken_ is `set -o errexit` and that‘s because the execution in a subshell doesn’t preserve `set -e`.

halostatue:
I’m pushing one more commit, but…I don’t like it. It works _perfectly_, but only by using temp files. I don’t really see a good way to do what we want to do otherwise (and `set -euo pipefail` causes some of the more clever mechanisms to fail, anyway).

zimbatm:
ouch I didn't realize it would be this painful to backup the original actions.

What do you think of creating a new direnv sub-command that backups the options?

Example usage:
```sh
local original
original=$(set -o | direnv options-to-bash)
set -euo pipefail
("$@")
eval "$original"
```
Where options-to-bash would parse each line and output a single line with corresponding `set -o` or `set +o` flags.

If you're not comfortable with Go I could help out with that.

zimbatm:
Actually, did you see the output of `set +o`? It's already doing exactly that (except that it's multi-line).

halostatue:
I had not seen the output of `set +o`. That’s useful, and we can just _grep_ the result to store…except that it runs into the same problem.

```sh
bash-3.2$ set -euo pipefail
bash-3.2$ set +o | grep 'errexit\|nounset\|pipefail'
set -o errexit
set -o nounset
set -o pipefail
bash-3.2$ echo $(set +o | grep 'errexit\|nounset\|pipefail')
set +o errexit set -o nounset set -o pipefail
```

I show the last as a subshell to `echo` because it approximates `local saved=$(set -o | grep '…')`. This is the problem. However, with the output to `set +o`, we can simplify the state saving to a single temporary file:

```diff
diff --git i/stdlib.go w/stdlib.go
index 56db8b0..16fb239 100644
--- i/stdlib.go
+++ w/stdlib.go
@@ -30,31 +30,15 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"export DIRENV_IN_ENVRC=1\n" +
 	"\n" +
 	"__store_env_strictness() {\n" +
-	"  local flag n1 n2\n" +
-	"\n" +
-	"  n1=/tmp/strictness.seto.$$.$RANDOM\n" +
-	"  n2=/tmp/strictness.stored.$$.$RANDOM\n" +
-	"\n" +
-	"  set -o > \"$n1\"\n" +
-	"\n" +
-	"  while read -ra flag; do\n" +
-	"    case \"${flag[0]}\" in\n" +
-	"      pipefail | nounset | errexit)\n" +
-	"        {\n" +
-	"          [[ \"${flag[1]}\" == off ]] &&\n" +
-	"            printf -- \"+o %s \" \"${flag[0]}\" ||\n" +
-	"            printf -- \"-o %s \" \"${flag[0]}\"\n" +
-	"        } >> \"$n2\"\n" +
-	"        ;;\n" +
-	"      *)\n" +
-	"        ;;\n" +
-	"    esac\n" +
-	"  done < \"$n1\"\n" +
-	"\n" +
-	"  # shellcheck disable=SC2207\n" +
-	"  strictness=($(cat \"$n2\"))\n" +
+	"  set +o | grep 'pipefail|nounset|errexit' > \"/tmp/env_strictness.$$.$1\"\n" +
+	"}\n" +
 	"\n" +
-	"  rm -f \"$n1\" \"$n2\"\n" +
+	"__restore_env_strictness() {\n" +
+	"  local file\n" +
+	"  file=\"/tmp/env_strictness.$$.$1\"\n" +
+	"  if [ -f \"$file\" ]; then\n" +
+	"    source \"$file\"\n" +
+	"    rm -f \"$file\"\n" +
 	"}\n" +
 	"\n" +
 	"# Usage: strict_env [COMMAND...]\n" +
@@ -79,8 +63,9 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"#\n" +
 	"#    strict_env has curl\n" +
 	"strict_env() {\n" +
-	"  local -a strictness\n" +
-	"  __store_env_strictness\n" +
+	"  local strictness\n" +
+	"  strictness=$RANDOM\n" +
+	"  __store_env_strictness \"$strictness\"\n" +
 	"\n" +
 	"  set -o errexit -o nounset -o pipefail\n" +
 	"\n" +
@@ -91,7 +76,7 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"\n" +
 	"    res=$?\n" +
 	"\n" +
-	"    set \"${strictness[@]}\"\n" +
+	"    __restore_env_strictness \"$strictness\"\n" +
 	"\n" +
 	"    return $res\n" +
 	"  fi\n" +
@@ -109,8 +94,9 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"#\n" +
 	"#    unstrict_env has curl\n" +
 	"unstrict_env() {\n" +
-	"  local -a strictness\n" +
-	"  __store_env_strictness\n" +
+	"  local strictness\n" +
+	"  strictness=$RANDOM\n" +
+	"  __store_env_strictness \"$strictness\"\n" +
 	"\n" +
 	"  set +o errexit +o nounset +o pipefail\n" +
 	"\n" +
@@ -121,7 +107,7 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"\n" +
 	"    res=$?\n" +
 	"\n" +
-	"    set \"${strictness[@]}\"\n" +
+	"    __restore_env_strictness \"$strictness\"\n" +
 	"\n" +
 	"    return $res\n" +
 	"  fi\n" +
diff --git i/stdlib.sh w/stdlib.sh
index 251b29a..dc0f36f 100755
--- i/stdlib.sh
+++ w/stdlib.sh
@@ -27,31 +27,16 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
 export DIRENV_IN_ENVRC=1
 
 __store_env_strictness() {
-  local flag n1 n2
-
-  n1=/tmp/strictness.seto.$$.$RANDOM
-  n2=/tmp/strictness.stored.$$.$RANDOM
-
-  set -o > "$n1"
-
-  while read -ra flag; do
-    case "${flag[0]}" in
-      pipefail | nounset | errexit)
-        {
-          [[ "${flag[1]}" == off ]] &&
-            printf -- "+o %s " "${flag[0]}" ||
-            printf -- "-o %s " "${flag[0]}"
-        } >> "$n2"
-        ;;
-      *)
-        ;;
-    esac
-  done < "$n1"
-
-  # shellcheck disable=SC2207
-  strictness=($(cat "$n2"))
+  set +o | grep 'pipefail|nounset|errexit' > "/tmp/env_strictness.$$.$1"
+}
 
-  rm -f "$n1" "$n2"
+__restore_env_strictness() {
+  local file
+  file="/tmp/env_strictness.$$.$1"
+  if [ -f "$file" ]; then
+    source "$file"
+    rm -f "$file"
+  fi
 }
 
 # Usage: strict_env [COMMAND...]
@@ -76,8 +61,9 @@ __store_env_strictness() {
 #
 #    strict_env has curl
 strict_env() {
-  local -a strictness
-  __store_env_strictness
+  local strictness
+  strictness=$RANDOM
+  __store_env_strictness "$strictness"
 
   set -o errexit -o nounset -o pipefail
 
@@ -88,7 +74,7 @@ strict_env() {
 
     res=$?
 
-    set "${strictness[@]}"
+    __restore_env_strictness "$strictness"
 
     return $res
   fi
@@ -106,8 +92,9 @@ strict_env() {
 #
 #    unstrict_env has curl
 unstrict_env() {
-  local -a strictness
-  __store_env_strictness
+  local strictness
+  strictness=$RANDOM
+  __store_env_strictness "$strictness"
 
   set +o errexit +o nounset +o pipefail
 
@@ -118,7 +105,7 @@ unstrict_env() {
 
     res=$?
 
-    set "${strictness[@]}"
+    __restore_env_strictness "$strictness"
 
     return $res
   fi
```

zimbatm:
ah the joys of bash programming :)

I think you could also backup and get the result from the temp file in a single block like that:
```bash
my_function() {
  # backup the old strictness
  local tmpfile old_strictness
  tmpfile=$(mktemp)
  set +o | grep 'pipefail\|nounset\|errexit' > "$tmpfile"
  old_strictness=$(< "$tmpfile")
  rm "$tmpfile"

  # ... do the things
  set -euo pipefail
  "$@"

  # and restore
  eval "$old_strictness"
}
```

halostatue:
Yes. I could do that, but that would end up duplicating the behaviour in both `strict_env` and `unstrict_env`, or we create two functions each `*_env` and `*_env_do` where `*_env_do` does the temporary setting. I can’t call out to a subshell at all in the upper level functions without messing up the `errexit` status.

I might be able to do this as something like:

```bash
__do_and_restore() {
  # this is a cheat. If __do_and_restore is called outside of a function that
  # declares $tempfile, set -euo pipefail will cause it to fail.

  local -i res
  res=0

  if (( $# )); then
    "${@}"
    res=$?
    source "$tempfile"
  fi

  rm -f "$tempfile"
  return $res
}

strict_env() {
  local tempfile
  tempfile=$(mktemp)
   __save_env_strictness "$tempfile"

  set -euo pipefail

  __do_and_restore "${@}"
}
```

That will reduce duplication.

halostatue:
Actually, the latest commit (just pushed) takes your idea and simplifies it substantially. I am happier with this and the tests that I added still pass.

stdlib.sh
```diff
@@ -26,6 +25,92 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
...
+
+    res=$?
+
+    __restore_env_strictness "$strictness"
+
+    return $res
+  fi
+}
+
+# Uuage: unstrict_env [COMMAND...]
```
zimbatm:
```suggestion
# Usage: unstrict_env [COMMAND...]
```

stdlib.sh
```diff
@@ -26,6 +25,92 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
 # algorithm and so it won't be re-exported.
 export DIRENV_IN_ENVRC=1
 
+__store_env_strictness() {
+  set +o | grep 'pipefail\|nounset\|errexit' > "/tmp/env_strictness.$$.$1"
+}
+
+__restore_env_strictness() {
+  local file
+  file="/tmp/env_strictness.$$.$1"
```
zimbatm: Some environments use TMPDIR to set the temp dir location.
```suggestion
  file="${TMPDIR:-/tmp}/env_strictness.$$.$1"
```

stdlib.sh
```diff
@@ -26,6 +25,92 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
 # algorithm and so it won't be re-exported.
 export DIRENV_IN_ENVRC=1
 
+__store_env_strictness() {
+  set +o | grep 'pipefail\|nounset\|errexit' > "/tmp/env_strictness.$$.$1"
```
zimbatm: Some environments use TMPDIR to set the temp dir location.
```suggestion
  set +o | grep 'pipefail\|nounset\|errexit' > "${TMPDIR:-/tmp}/env_strictness.$$.$1"
```

stdlib.sh
```diff
@@ -26,6 +25,92 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
...
+# If followed by a command-line, the strictness applies for the duration
+# of the command.
+#
+# Example:
+#
+#    strict_env
+#    has curl
+#
+#    strict_env has curl
+strict_env() {
```
zimbatm: Since there are two uses here, I recommend to early-exit when there are zero arguments.
```suggestion
strict_env() {
  if [[ $# -eq 0 ]]; then
    set -euo pipefail
    return
  fi
```
With the current implementation the __store_env_strictness is called in that first case and the temp file left around on the filesystem.

This would duplicate the `set -o errexit -o nounset -o pipefail` line but I don't think we can do better.

stdlib.sh
```diff
@@ -26,6 +25,89 @@ direnv_config_dir=${XDG_CONFIG_DIR:-$HOME/.config}/direnv
...
+  unstrict)
+    set +o errexit +o nounset +o pipefail
+    ;;
+  *)
+    echo "Unknown strictness mode '${mode}'."
+    exit 1
+    ;;
+  esac
+
+  if (($#)); then
```
halostatue: I _could_ remove this particular test because of the early exit in `strict_env` and `unstrict_env`, but this still seems safer.

zimbatm:
@halostatue the last part that's missing is to also update doc/direnv-stdlib.1.md

halostatue:
OK. I’ll get to that later.

halostatue:
Should be ready now.

zimbatm:
sorry for the wait. I rebased the PR and make a few cosmetic changes.

PR #575, opened 2020-01-31
config: fix the configuration file selection

ytoku:
`direnv.toml` was not read since `config.TomlPath` was overwritten.


zimbatm:
doh. thanks!

Issue #478, opened 2019-03-21
[fish] UTF in env variables breaks direnv


Profpatsch:
This might be a regression, because the changelog states that UTF is supported.

### Steps to reproduce

`.envrc`:
```
export UTFSTUFF=♀♂♀♂
```

`direnv allow`
```
direnv: loading .envrc
direnv: export +UTFSTUFF
- (line 1): Invalid token '''\xe2''\x99''\x80''\xe2''\x99''\x82''\xe2''\x99''\x80''\xe2''\x99''\x82'''
begin; set -x -g 'UTFSTUFF' ''\xe2''\x99''\x80''\xe2''\x99''\x82''\xe2''\x99''\x80''\xe2''\x99''\x82'';set -x -g 'DIRENV_DIFF' 'eJw0y02umkAAAOC7zFoC5L0HlKQLeICAgAqIwoYgHf4UZhzRgTEmTW_h9TxJF0033-57AAzUx3MBBqA-gOGEZpDkhhMCFXB8g3rI46Y9t5g_oZGniJz4kcKi5s-IkJYnEBMEFv_fXou_bTMCKoDudGXEWlOmbbVAz5WQibIdKEfiM4eYnX5bl0tpvHnkUuj1MMkcRZvcED6G3N9zF6_f0mvZV-l5JBe9q_DVuUO05koy6uaYejOuneGAuYLtdZhBWWR9NjuHxv6yhZL540dVRXZ_xJwCf9DO5WaySmBayCtpCqJCdo3vrf21q71M9qvwBsM5yopwatgmNuzOTVpfkaDoiyvBa9JTRwe2LIRho7UBu396VUUtTTPyXEp-RejzJ1iAXWxF8c6ygArer9_v159_gufzbwAAAP__vGN5vQ==';set -x -g 'DIRENV_WATCHES' 'eJxszrFOwzAQANB_8Rz17HN8brMzIrEjBuOcG6tuLrqaBgnx7-woP_D03n_MW-qLmQwscmfYltrqBjfpsIveoO-crtBEtYLypgInXp-azWBeZe71zmZyIXhH5H0czMt3ffSHmbp-8e9wjJ-yrKVeYa7K6xNSa7JDCQH5UgLZ7MfRueRySZaRxhzPTDHjJViM86e1M1K0LhYkjwnzGa0nPAiNzv4LffwFAAD__6VdSo4='; 
                            ^
from sourcing file -
[...]

PR #584, opened 2020-02-06
Update shell_fish.go to use \X for UTF encoding

nkoch-ops:
This should fix issue #478.

### DIFF
--- a/shell_fish.go
+++ b/shell_fish.go
@@ -60,7 +60,7 @@ func (sh fish) escape(str string) string {
 	l := len(in)
 
 	hex := func(char byte) {
-		out += fmt.Sprintf("'\\x%02x'", char)
+		out += fmt.Sprintf("'\\X%02x'", char)
 	}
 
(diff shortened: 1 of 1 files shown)


zimbatm:
I added some tests in https://github.com/direnv/direnv/pull/585 to test UTF-8. If you uncomment the fish test you will see if it fixes it.

nkoch-ops:
Alright, thanks for that.
I already checked manually with a test case that originally caused that error.
The test unfortunately errors out with:
```
Error:
 - expected: ♀♂
 -      got: ♀♂
```
So it looks like the fix seems to work, but the comparison fails for some reason?

zimbatm:
The joys of Unicode. To fix this you will have to see if the hex code of both outputs is the same because Unicode characters can be encoded in multiple ways. In the Go code, you might want to try and iterate over each character instead of bytes by using the `for char := range str {}` iterator instead.

zimbatm:
finally!

PR #596, opened 2020-02-20
stdlib: Cache use_guix results

ajgrf:
This commit caches the result of `use_guix`, which is normally unusably slow. After this change, subsequent loads are now nearly instantaneous.

However, it is slightly more confusing to use now, because users will need to remove the cache to make changes.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -787,9 +787,16 @@ use_nix() {
 # `--ad-hoc` flag is used `use guix --ad-hoc hello`. Other options
 # include `--load` which allows loading an environment from a
 # file. For a full list of options, consult the documentation for the
-# `guix environment` command.
+# `guix environment` command. The results will be cached in a profile
+# under ./.guix-profile to speed up subsequent loading. Remember to
+# remove it when you make changes!
 use_guix() {
-  eval "$(guix environment "$@" --search-paths)"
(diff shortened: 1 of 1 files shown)


stdlib.sh
```diff
@@ -787,9 +787,16 @@ use_nix() {
 # `--ad-hoc` flag is used `use guix --ad-hoc hello`. Other options
 # include `--load` which allows loading an environment from a
 # file. For a full list of options, consult the documentation for the
-# `guix environment` command.
+# `guix environment` command. The results will be cached in a profile
+# under ./.guix-profile to speed up subsequent loading. Remember to
+# remove it when you make changes!
 use_guix() {
-  eval "$(guix environment "$@" --search-paths)"
+  if [[ -e ./.guix-profile/etc/profile ]]; then
```
zimbatm: We have a standard location for project-specific dependencies to store this:
```suggestion
  if [[ -e "$(direnv_layout_dir)/guix-profile/etc/profile ]]; then
```

zimbatm:
There is a similiar issue with `use_nix`. The problem is that like you said, it requires the user to know about the implementation to invalidate the cache manually.

I think this is best added to the wiki with an explanation so that users learn about the implementation details before adding it to their `direnvrc` file

ajgrf:
Thanks, I just added it to the wiki and will close this PR.

PR #607, opened 2020-03-07
Fix issues caught during debian package update to v2.22.2

punitagrawal:
These are issues in direnv sources that were encountered when packaging for debian. As they might affect other users as well, please consider merging.

### DIFF
--- a/man/direnv.1.md
+++ b/man/direnv.1.md
@@ -36,7 +36,7 @@ $ cd ~/my_project
 $ echo ${FOO-nope}
 nope
 $ echo export FOO=foo > .envrc
-.envrc is not allowed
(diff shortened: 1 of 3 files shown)


zimbatm:
thanks!
I just had to push another change to re-generate the .1 and stdlig.go file: 8b4b2687c9146ac1e315f69e0e372ded582beab3

punitagrawal:
Thanks for merging the patches (and fixing the generated files).
In Debian, we are triggering re-generation for every build so I missed the mismatch.

zimbatm:
yeah I know :)

Thanks again for the patches. It's nice to get well crafted patches with explanations in them.

PR #609, opened 2020-03-11
Permissions check for data directory

iRyanBell:
Without access to the `$USER/.local/share` directory, `cmdAllowAction` throws `".envrc file not found"` -- this message doesn't accurately describe other potential issues. RCFromPath() errors inside FindRC() are often masked.

As a solution, `RCFromPath` could be modified to return a tuple (along with all functions referencing FindRC), or `LoadConfig` could be modified to perform this check. As a starting point, I brought in `golang.org/x/sys/unix`, adding this check and appending it to the status report.


config.go
```diff
@@ -182,6 +189,26 @@ func (config *Config) LoadedRC() *RC {
...
+
+	if err != nil {
+		return false, fmt.Errorf("data home directory %s doesn't exist", dataHomeDir)
+	}
+
+	if info.Mode().Perm()&(1<<(uint(6))) == 0 {
+		return false, fmt.Errorf("can't read directory %s", dataHomeDir)
+	}
+
+	if unix.Access(dataHomeDir, unix.W_OK) != nil {
```
zimbatm: Is it not possible to use `os.Stat` here and avoid the sys/unix package dependency?

iRyanBell: This was the thread that turned me down the sys/unix dep path: [StackOverflow: How to tell if folder exists and is writable? (GoLang)](https://stackoverflow.com/questions/20026320/how-to-tell-if-folder-exists-and-is-writable)

Through `os.Stat`, I can use `.Mode().Perm()` to check the directory permissions (validating `+x`), but I didn't see a clear way to get the permissions relative to the user through .Stat alone to validate `~/.local/share` access (eg. if owned by root.)

This line would provide the clue, but it's called too late:
rc.go:
`os.MkdirAll(filepath.Dir(rc.allowPath), 0755)`

Without access to data_home, FindRC fails (and throws the wrong error)

Invalid error for insufficient DATA_HOME access

```
rc := FindRC(rcPath, config)
if rc == nil {
    return fmt.Errorf(".envrc file not found")
}
return rc.Allow()
```

Perhaps the attempt to make `config.AllowDir()` could occur sooner, although it's nice to have that at the end of the function call rather than during the config check.

zimbatm: Okay, thanks for the explanation!

I agree, it would be much better to get the error directly from `MkdirAll` and not essentially duplicate what it does, with some potential margin of error.

Do you mind finding a way to do that?

iRyanBell: It looks like with error tuples added to a few return statements, the correct messages can be passed to the user. I've added a new PR with these changes.

Here's what the correct error looks like with this patch:

Screen Shot 2020-03-17 at 4 04 21 PM

PR #612, opened 2020-03-20
Source hash

### LINKED ISSUE #448: Feature request: Make direnv ask for verification of changes to watched files
## Issue description

When sourcing or running external files from .envrc, direnv will execute changed code in those files without verification. This is clearly documented, but I still think it would still be great if the `direnv allow` process could be applied to such files as well.

Since direnv already knows how to watch other files and reload when they change, the same list of files would be a great candidate for files to also protect from unverified changes.

*Note: This is a very humble suggestion without any insight into how complex the implentation would be. Since the documentation mentions the lack of this feature, I guess it has been considered but deemed non-trivial.*

### Steps to reproduce

1. Source and watch an external file from .envrc
2. Call `direnv allow`
3. Change the external file

Direnv will reload and source the changed external file without verification.

jeci-sarl:
I implement an alternative solution to resolve #448 . I create a `source_hash` command (a wrapper of `source_env`) that run the external file only if it match the hash given.

Before going further, I want to ask you if this is a good idea?

I think we can add the same mechanism on `use_nix` to optionally validate `shell.nix` file.

Sample usage:

``` bash
$ cat .envrc 
source_hash envrc.test 228f40df92e538a261071baefe4f48926060f130
source_hash notfound.sh
source_hash ../base 8989705c25c4a9f3f0c7617a762279bf016a269f

$ cat subdir/.envrc 
source_env .. 8773e41fb8788695e71cc6a373fbe312517ff18f

$ cd subdir 
direnv: loading ~/git/direnv/test/scenarios/hash/subdir/.envrc
direnv: loading ../.envrc
direnv: loading envrc.test
direnv: referenced notfound.sh does not exist
direnv: loading ../base/.envrc
direnv: export +HELLO +TEST

$ echo "#Add Comment" >> ../envrc.test 
direnv: loading ~/git/direnv/test/scenarios/hash/subdir/.envrc                                                                                           
direnv: loading ../.envrc
direnv: referenced envrc.test has change, please rehash
direnv: referenced notfound.sh does not exist
direnv: loading ../base/.envrc
direnv: export +HELLO

$ sed -i '/source_hash/ s,envrc.test.*,envrc.test '"$(shasum -t "../envrc.test" | cut -d \  -f 1)"',' ../.envrc
direnv: loading ~/git/direnv/test/scenarios/hash/subdir/.envrc                                                                                           
direnv: referenced ../.envrc has change, please rehash

$ sed -i '/source_hash/ s,\.\..*,\.\. '"$(shasum -t "../.envrc" | cut -d \  -f 1)"',' .envrc
direnv: error /home/jlesage/git/direnv/test/scenarios/hash/subdir/.envrc is blocked. Run `direnv allow` to approve its content       
                    
$ direnv allow
direnv: loading ~/git/direnv/test/scenarios/hash/subdir/.envrc                                                                                           
direnv: loading ../.envrc
direnv: loading envrc.test
direnv: referenced notfound.sh does not exist
direnv: loading ../base/.envrc
direnv: export +HELLO +TEST
```

I made a bash script to automatically update hashes (instead of using manual sed). It is not fully functional but before correct all leading bugs I want to integrate this functionality in direnv. My first idea is to create new command on direnv

* `direnv hash --update` : update all `source_hash` reference, like my script `direnv_update_hash` do.
* `direnv hash <file>` : add or update a `source_hash` reference

I also think to add `--hash` parameter on `direnv allow` to automatically hash and allow. Or on `direnv edit` to manually show/edit `source_hash` reference.

What do you think about this solution ?

### DIFF
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -101,6 +101,13 @@ Loads another \fB\fC\&.envrc\fR either by specifying its path or filename.
 .PP
 NOTE: the other \fB\fC\&.envrc\fR is not checked by the security framework.
 
+.SS \fB\fCsource\_hash <file\_or\_dir\_path> <shasum>\fR
+.PP
+Loads another \fB\fC\&.envrc\fR either by specifying its path or filename.
+
+.PP
+The other ".envrc" is validated using shasum check.
+
 .SS \fB\fCsource\_up [<filename>]\fR
 .PP
 Loads another \fB\fC\&.envrc\fR if found when searching from the parent directory up to /.
--- a/man/direnv-stdlib.1.md
+++ b/man/direnv-stdlib.1.md
@@ -75,6 +75,13 @@ Loads another `.envrc` either by specifying its path or filename.
 
 NOTE: the other `.envrc` is not checked by the security framework.
 
+
+### `source_hash <file_or_dir_path> <shasum>`
+
+Loads another `.envrc` either by specifying its path or filename.
+
+The other ".envrc" is validated using shasum check.
+
 ### `source_up [<filename>]`
 
 Loads another `.envrc` if found when searching from the parent directory up to /.
--- a/script/direnv_update_hash
+++ b/script/direnv_update_hash
@@ -1,0 +1,25 @@
+#!/usr/bin/env bash
+
+update_hash() {
+  while IFD=' ' read -r _ _path hash
+  do
+    if [[ -d "$_path" ]]; then
+      rcpath="$_path/.envrc"
+    else
+      rcpath="$_path"
+    fi
+    rchash=$(shasum -t "$rcpath" | cut -d \  -f 1)
+
+    if [[ "$hash" != "$rchash" ]]; then
+      echo "referenced $rcpath has change, rehash ?"
+      echo "------------------"
+      cat "$rcpath"
+      echo "------------------"
+
+      sed -i '/source_hash/ s,'"$_path"'.*,'"$_path"' '"$rchash"',' .envrc
+    fi
+
+  done <<< "$(grep -E '^source_hash' .envrc)"
+}
+
+update_hash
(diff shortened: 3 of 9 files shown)


zimbatm:
Interesting idea! It's tangentially related to this other WIP PR: https://github.com/direnv/direnv/pull/562 which downloads and evals sources from the internet. I wonder if both could be included in one.

Fundamentally, bash is hard to sandbox, it would require to override the `source` and `eval` builtins at minimum. But it doesn't mean that we can't try to make things better.

To update the hashes you probably want a special marker because bash code is not structured enough to make the replacement reliable. It's might be easier to do a find-and-replace on the raw text stream.

I don't know yet if this is worth the extra complexity but am willing to play with the idea. For example, what happens when there is a chain of sources files, will `direnv hash --update` follow all the files and update them recursively?

jeci-sarl:
I'm agree, your PR is related. One difference, the hash is mandatory in my solution. If the hash is absent, the target file is ignored and a warn invite user to update the hash.

I can rewrite my solution into `fetchurl`. I will add a check if the url refer to local url (`file://` or no protocol) and add interpretation (`.`, `..`, `$HOME`, etc.).

To update hashes in `.envrc` files I use a regex on `/^source_hash/` to replace all target references. I will re-write my script in go and update it to match `/^source_url/`.

We can try recursive update, but we must protect the code for loop call. What would be great is to make a recursive and interactive `direnv allow`.

Thanks for your feedback.

zimbatm:
I still think that this PR is useful. Now that direnv has support for SRI hashes, one way forward would be to add a new `direnv hash-file` command. And then implement `source_hash` with that.

AkechiShiro:
Any way to push this PR forward ? @zimbatm 
Should the current code be discarded, because too old/no longer relevant ?
Has the `direnv hash-file` command not been implemented ? (at least not merged in Master from what I see)

PR #615, opened 2020-03-27
stdlib: add PATH_rm <pattern> [<pattern> ...]

bdupras:
When working on a machine with a managed development environment, it's common for the managed `PATH` to have many entries. `PATH_rm` allows me to selectively remove entries that conflict with specific projects.

Example:

    echo $PATH
    # output: /opt/my_employer/thing1/bin:/opt/my_employer/thing2/bin:/usr/local/bin/:...
    PATH_rm '/opt/my_employer/*'
    echo $PATH
    # output: /usr/local/bin/:...

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -10,6 +10,7 @@
 # SC2059: Don't use variables in the printf format string. Use printf "..%s.." "$foo".
 shopt -s gnu_errfmt
 shopt -s nullglob
+shopt -s extglob
 
 
 # NOTE: don't touch the RHS, it gets replaced at runtime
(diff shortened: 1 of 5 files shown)


stdlib.sh
```diff
@@ -369,6 +370,63 @@ MANPATH_add() {
...
+  # iterate over path entries and test for pattern match
+  for path in "${path_array[@]}"; do
+    match=false
+    for pattern in "${patterns[@]}"; do
+      if [[ "$path" == +($pattern) ]]; then
+        match=true
+        break
+      fi
+    done
+    if ! $match; then
```
zimbatm: nit: rename the variable to `discard` so the logic become obvious. or invest the boolean logic to `if $keep; then`

bdupras: 👍 ah yes - "$discard" is much clearer. submitted.

bdupras:
Thanks @zimbatm! Cool project.

zimbatm:
thanks!

PR #619, opened 2020-04-07
direnv-stdlib.1: add layout php

glensc:
I noticed that I have `layout php` in one project and `layout_php` in another and wondered what is the canonical one that should be used.

looked up in documentation and did notice `layout php` missing:
- https://direnv.net/man/direnv-stdlib.1.html#codelayout-lttypegtcode


glensc:
seems some others are missing, are they really all manually copied from stdlib source?

zimbatm:
thanks :)

yes, they are right now, and people submit PRs without updating the doc. It would be good to write a script that extracts those from the comments in the stdlib.

PR #622, opened 2020-04-15
stdlib: `layout go` adds layout dir to GOPATH

kayhide:
Update `layout go` so that the layout dir is added to GOPATH instead of PWD.

Since go does not want to see `go.mod` in GOPATH, PWD is not available sometimes.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -474,10 +474,11 @@ layout() {
 
 # Usage: layout go
 #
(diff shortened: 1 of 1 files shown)


zimbatm:
ok why not

zimbatm:
hopefully this won't uncover yet another edge-case

Mic92:
Makes sense. Might be changelog worthy.

PR #624, opened 2020-04-24
Remove ">" typo in Fish hook instructions

KingMob:
For some reason, the Fish example was rendering with a ">" at the beginning, even though ">" wasn't present in the code itself. This is invalid, no ">" is needed. I changed the backtick language from `fish` to `sh` like the others, and it seemed to stop rendering a ">".


zimbatm:
ah, that's where it was coming from. thanks!

PR #625, opened 2020-04-25
Update direnv-test.elv

xiaq:
The "path" field of (src) has been removed. Use "name" to get the path of file sources.

### DIFF
--- a/test/direnv-test.elv
+++ b/test/direnv-test.elv
@@ -1,6 +1,6 @@
 #!/usr/bin/env elvish
 
-E:TEST_DIR = (path-dir (src)[path])
+E:TEST_DIR = (path-dir (src)[name])
 set-env XDG_CONFIG_HOME $E:TEST_DIR/config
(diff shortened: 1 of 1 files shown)


zimbatm:
For which version of elvish does that apply?

xiaq:
Good question. I thought I introduced this in 0.13 but forgot to document it, but at it turns out it was changed after 0.13. The first release that contains this change will be 0.14, not yet released. Feel free to close this and delay the change until 0.14 is released.

zimbatm:
Is there a way to make the change backward-compatible? So that elvish users are not forced to update both direnv and elvish at the same time.

xiaq:
Hmm, to ease transition I can reintroduce `path` as a valid field for indexing in 0.14, and remove it later in 0.15.

xiaq:
The `path` field now exists: https://github.com/elves/elvish/commit/4d54943ffb4ee3e161a811affbbb295c8de86e6a

This change should still be made after Elvish's 0.14.0 release, since it will be removed afterwards.

Issue #594, opened 2020-02-17
Cannot export variable error while running on Windows 10


sph3rex:
**Describe the bug**
Error triggered everytime on Windows 10 (VirtualBox 6.0.12_SUSEr132055) : "direnv: error write 3: The handle is invalid."
I would assume is regarded to environment variables handling and not related to vbox env.

**To Reproduce**
Download latest version of direnv (v2.21.2  - direnv.windows-amd64), copy it to 

> /usr/bin

 make a test dir called "test", add an .envrc file with 

> export P=1
 and execute 

> direnv allow
direnv: error write 3: The handle is invalid.

**Expected behavior**
P variable should be exported, and aforementioned error should not be present.

**Environment**
 - OS: Windows 10 Pro 1909 - 18363.657 (VirtualBox 6.0.12_SUSEr132055)
 - Shell: gitbash (mintty 3.1.0) - ran both as administrator or regular user
 - Direnv version v2.21.2

zimbatm:
I am pretty sure that this an issue with how Windows handles file descriptors (ie: not). In a4632773637ee1a6b08fa81043cacd24ea941489 the stdout file descriptor is backed-up to FD 3.

sph3rex:
Thanks for the reply. Is there any solution that comes easily or I should just fork, dive and try to make it work? I tried on a physical win10 machine and it has the same behaviour. Meanwhile I think this should be stated clearly on readme, that support for Windows is not here (yet?!) even if there are releases for it.

zimbatm:
Hey,

Have you been using direnv on Windows for long? My understanding is that there are other issues such as PATH handling which made it unusable but maybe I am wrong. I would very much like direnv to support Windows but didn't have the time to implement it myself.

If you want to dig into it, the line that triggers the issue is here: https://github.com/direnv/direnv/blob/83c861eadc48de96b2905e7ab011785dd49bcb07/stdlib.sh#L803-L804 . I don't really have an idea on how to fix while guaranteeing that direnv dumps and outputs don't get intermingled.

I agree that Windows shouldn't be advertised as ready. It's currently not tested by CI.

sph3rex:
I'm not using it for long on Windows, because my main is Linux, but I sometimes need a Windows vm or otherwise for various tasks. Having it running equally well on both OSes would make my life easier, as direnv already does on Linux. Path handling can be somehow tamed, however not being able to export anything makes it unusable. I will try to have a look on it on Windows when I'll get some spare time.
[...]

PR #630, opened 2020-05-06
Draft Windows support (fix for #594 and related)

pjeby:
* Fix for #594 (write error to fd 3 on Windows)
* Fix wrong SelfPath for 'direnv hook' on Windows
* Fix `$PATH` corruption across process boundaries

The PATH corruption issue is fixed by adding an extra command line argument, `UNIX_PATH`, to `direnv export`, `direnv dump`, and stdlib's `__main__`:

If it is non-empty and *different* from the value of `$PATH`, it is used in place of `$PATH` for purposes of generating the exported environment.  Shells supply this argument from their hook function to bypass the conversion between Windows and Unix-style paths that takes place when changing between shells and Go programs on Windows, thereby preserving symlinks, case, etc. that would otherwise be lost in translation.

Internally, a DIRENV_PLATFORM_PATH key is added to the logical environment processed by `direnv export`, `direnv dump`, and `direnv exec`, and its changes are tracked in DIRENV_DIFF, but it is never actually exposed as an environment variable to parent or child processes, nor is it ever read from the OS environment.

(It's tracked in DIRENV_DIFF so that `direnv exec` can know the previous OS path, though at the moment that's not very useful because syscall.Exec doesn't work on Windows and I don't know what a proper replacement would be.  So for the moment, `direnv exec` still doesn't work on Windows.)


pjeby:
FYI, I'm in the middle of trying to get the full test suite to actually *run* on Windows, some parts of which are harder than others.  :wink:

(One interesting note is that Windows cannot distinguish between empty and unset environment variables, so those tests would need to be skipped.)

Apart from the empty vs. unset tests, I actually have the bash tests successfully running on Cygwin in my current work directory, but there are a couple of hairy bits having to do with DIRENV_DUMP_FILE_PATH needing to be a Windows path but it being set from bash (and thus needing conversion).  There are some hurdles still to go with getting this working on msys/Git bash, as well.

So, this is definitely not ready to merge just yet, but I thought I should inquire as to whether you think this is a viable approach or if you have any objections to the direction this is taking.  Thanks!

cmd_export.go
```diff
@@ -52,6 +52,16 @@ func exportCommand(ctx context.Context, env Env, args []string, config *Config)
 		target = args[1]
 	}
 
+	// PATH is the "logical" path manipulated by stdlib and .envrc
+	// DIRENV_PLATFORM_PATH is the OS-specific path needed for invoking programs.
+	// Anywhere but Windows, these are both the same.
+	env[DIRENV_PLATFORM_PATH] = env["PATH"]
```
zimbatm, line 58: The current implementation seems quite tricky to get right. The developer has to know all the execution paths and make sure that the env is added or removed at the right places.

Maybe it would work to add it to env.CleanContext() and in the IgnoredKeys?

pjeby: No.  I thought I could do that, but it means that `direnv exec` wouldn't work.  The platform path has to be diffed, so it can't be ignored.  (Otherwise, `direnv exec doesn't` know what the platform path should be when the current .envrc is unloaded.)

So, what you see now is currently the simplest logic that would still let direnv exec function, while also being (relatively) straightforward to reason about.

Specifically, the invariant is that `DIRENV_PLATFORM_PATH` is never directly visible as a variable *outside* of the direnv process, but must always be present *inside* of it. So `ToGoEnv` replaces `PATH` with it and removes it.  And in a handful of other places where it would be visible to the parent or child process, it's deleted.  But, it must *not* be removed at any other time.  As soon as an env-affecting command starts, it sets it from PATH, and there it stays until there's a chance of it being exposed to another program.

TBH, the hairiness of where those deletes are has more to do with existing organizational issues in the code, than to do with the changes themselves.  There is a lot of semi-invisible state coupling taking place in export, not counting the indirect state coupling *across* different commands!

I almost think it'd be easier to replace direnv dump with some bash code; an awful lot of the hairyness for Windows has to do with it in one way or another.  Dumping anything but the gzenv format would be pretty easy to do from bash in the stdlib, and it would fix the various piping, tempfile paths, etc. issues I'm having with the test suite.  But of course there are other problems with that approach...

zimbatm: > TBH, the hairiness of where those deletes are has more to do with existing organizational issues in the code, than to do with the changes themselves. There is a lot of semi-invisible state coupling taking place in export, not counting the indirect state coupling across different commands!

Ouch. This rings very true :) Even as the main author I need to re-load all of this complexity every time I touch the main engine.

pjeby: Yeah.  I'm finding this is harder than it looks; notably, I haven't yet encountered this issue at runtime use of this version, but I'm pretty sure that watches on anything other than the original .envrc file aren't going to work, because the stdlib will pass a unix path to direnv watch, which being a Windows program, it won't understand.

(Ironically, going the other way is not necessarily a problem; Cygwin and MSYS programs can use Windows paths just fine, as long as they don't try to *manipulate* them in any way, which is why the initial call to `source_env` works fine despite direnv supplying a Windows path.)

Likewise, the stdlib's call to `direnv dotenv` would also need to convert the path it sends in.

I'm starting to think it makes sense to play around more with doing .direnvrc additions to patch the stdlib for cygwin/msys.

I think what I should do at this point is first do a PR *just* to fix the incorrect path in `direnv hook`, which I think we can agree is an uncontroversial change, since it makes it possible to *try* running direnv on Windows.  :wink:

Next, a PR for some changes to the stdlib to detect a cygwin or msys environment, so as to only pass Windows paths to direnv {dotenv,watch,dump,...}.

At that point, we can revisit how to address PATH, empty environment variables, and anything else that comes up.  The only real difficulty I see is that the path conversion changes to the stdlib can't really be tested by the test suite yet, except insofar as testing that they don't break non-Windows platforms.  ;-)

What do you think of that?  Or I suppose I could just do one PR with both the direnv hook fix and the stdlib changes for Windows path conversion?

(I'm thinking the path conversion change for the stdlib could just be something like defining a direnv function in the stdlib that runs "$direnv" "$@" on non-Windows, but on Windows rewrites any paths in the arguments and in `DIRENV_DUMP_FILE_PATH`.  Then replace all the current `$direnv` invocations with calls to that function instead.)

Alternately, I guess I could go ahead and do that in *this* PR, and see if that gets the test suite running, apart from empty vars.

Your thoughts?

pjeby: Another thought: `expand_path` can be portably implemented in pure bash, which would fix one subcommand right up, and allow dropping the `direnv expand_path` command.

The `realpath.absolute` function from [realpaths](https://github.com/bashup/realpaths) implements an all-bash substitute for the go `expandPath` function, just with arguments ordered the other way around (i.e. "[relto] path" rather then "path [relto]").  That would let us just drop `cmd_expand_path.go` entirely, incidentally making things faster on Windows since forks are expensive there.  (And the current way of using expand_path requires two forks and an exec, vs. just one fork and no execs with a bash version.)

In other words, by adding the `realpath.absolute` and `realpath.dirname` functions from realpaths (about 12 lines of code) to the stdlib, the new version of expand_path would look like this:

```shell
expand_path() {
  local REPLY; realpath.absolute ${2+"$2"} ${1+"$1"}; echo "$REPLY"
}
```

And then that's one less thing to figure out Windows path translations for.  ;-)

So in principle, this could be split to another prerequisite PR, i.e. replacing cmd_expand_path.go with the above version of expand_path.

pjeby:
So, with some extra changes not shown here, I've gotten the tests to run for bash, fish, and zsh on Cygwin.  (Don't have elvish or tcsh installed at the moment.)

So next, I tried it out on msys (git bash), only to find out (after much trial and tribulation), that git bash does [this bullcrap](http://www.mingw.org/wiki/Posix_path_conversion).

Yeah that's right.  When running any non-msys executable, it *silently translates command arguments based on guesswork as to whether something *looks* like a path!  Which means it takes the extra "$PATH" argument and goes, "oh, it looks like you're trying to pass unix paths to a Windows program.  Here, let me fix that for you..."  BOOM.  Headshot.  There goes the entire functioning of this PR.

And it translates environment variables in this way as well.

So the good news is that this means that all msys/git bash needs to work is the stdlib dump piping fix I did here.  (Using redirection instead of passing an fd.)  Yay?  (i.e. the actual fix for #594, and the addition of #632, are really the only thing stopping direnv working under msys/git bash.)

The bad news is that the rest of the work I've done for cygwin is precisely counterproductive for msys, because the matching path value throws this PR off and ends up with it thinking it's working with a pure windows shell like Elvish, and then having stuff break in the stdlib.

It *is* possible to force msys to *not* translate an argument by adding it to `MSYS2_ARG_CONV_EXCL`, but then that would become part of the direnv state while simultaneously preventing the variable from being controllable *by* direnv.  So that's probably best left alone, I guess.

Hm, now that I think of it, probably lots of things can go badly with direnv on git bash/msys due to the conversion, because it *also* converts environment variables, and anything that looks like an absolute path is going to get mangled on the way back into the enclosing shell.   And the only way around it is to pipe stuff in and out to bypass the mangling.

So, it's really more like, "msys/git bash is an easy fix if the mangling doesn't break anything you care about".  Otherwise, it requires the full-on "direnv never trusts the OS environment as a diffing source or target on windows and instead expects to be piped variables from the calling shell" approach I laid out above.

Anyway, I need to give this some more thought, to see if there's any way to get a Windows version that doesn't suck for either kind of shell environment (msys or cygwin), and at the same time doesn't require a complete reworking of direnv internals.  :frowning_face:

cmd_export.go
zimbatm: I think it's a good idea to take baby steps. Make the minimal changes in this PR so that things mostly work on Windows. Disable most of the tests for Windows and test in CI. Then expand.

The worst outcome, in my opinion, would be to aim for something perfect and run out of energy before it's finished. As long as it doesn't break backward-compatibility then it's fine.

+1 on replacing expand_path with a pure-bash implementation as long as it works in Bash 3.0 (thanks macOS...)

zimbatm:
I really admire your persistence :)

> So next, I tried it out on msys (git bash), only to find out (after much trial and tribulation), that git bash does this bullcrap.

This is the kind of difference that led me to believe that the golang path package should take an encoding argument. Either at runtime or at compilation time.

Do you think it would help if direnv wasn't perceived as a Windows program? I get the sense that on Windows you have to treat cygwin, msys/git, ... each of those as essentially different environments. And have a version of direnv that is compiled specifically for that target.

Another related idea I had was to embed a statically-compiled version of Bash into direnv itself. This would allow to also patch bash to change its behavior.

zimbatm:
Let's get the FD fix in as a first PR.

cmd_export.go
pjeby: PR #631 is bash 3.2 compatible.  The original raison d'etre for bashup/realpaths, as it happens, was to provide a realpath alternative for OS X without homebrew.  So it had to not only be pure bash, but pure bash 3.2.  I'm actually pretty hyper-aware of the differences between certain versions of bash, such as the horrible interaction of `set -u` and array variables that happens in bash 4.3, but not 3.2-4.2 or 4.4+.  :wink:

PR #631, opened 2020-05-08
Replace `direnv expand_path` with pure bash

pjeby:
This makes there be one less command to figure out how to port to Windows.  :wink:  It also makes expand_path faster, as there is no need to fork and exec direnv to do the bash equivalent of filepath.Join and filepath.Clean.


zimbatm:
thanks! Just pushed a basic test for the sake of it.

pjeby:
LGTM.  realpaths has an extremely exhaustive suite, and I copied the code verbatim except for function names, so a smoke test like that one is really all we need here, I think.

zimbatm:
exactly

PR #635, opened 2020-05-08
ci: add basic windows CI

zimbatm:


zimbatm:
@pjeby it looks like direnv has more Unixism (see rc_test.go). What is the recommended way to deal with paths in a cross-platform manner?

pjeby:
Well, for that test you could just delete the tests and the function they're testing, which is what I did in my WIP because the rootDir function is no longer used anywhere AFAICT.  :wink:

zimbatm:
good point. any ideas why the rest of the tests are failing?

pjeby:
The eachDir test is failing because `/` yields `E:\` and `E:` as parents.  (Well, at least on my machine, where the tests are running on E:.)  It's a more-or-less reasonable outcome (though the meaning of `E:` and `E:\` are subtly different).  I've basically just been skipping that test while figuring out more important stuff.

I'll go ahead and tell you that there are a bunch of shallow failures of this sort elsewhere, like the use of `cd -P` making paths look wrong when you compare them even though they're actually equivalent.  I've got workarounds for some of them, and a few commented out that just aren't going to pass (the empty-var ones) and I also have the symlink-changed test commented out for now as well.

These changes are all in my work tree for the attempt at fixing PATH, which I need to clean out and rebase now that so many other things are in.

zimbatm:
Ok, it might make sense to pick my commits into your branch then. I will close this PR.

PR #636, opened 2020-05-08
Streamline core algorithm of export and exec

pjeby:
The previous logic flow for export was highly non-linear and tightly state-coupled across multiple functions with side-effects, making it extremely hard to follow what was happening.

This refactoring linearizes the core logic of export, replacing struct members with simple local variables, and makes all the internal functions fully functional and side-effect free.  In addition, some duplicated high-level code and miscellaneous housekeeping from exec and export was moved into rc and config, further clarifying how both exec and export work.

You can now see how the various env values are reverted, diffed, patched, etc. in one screenful of code, without needing to search through multiple functions trying to figure out who was doing what to whom in which order.  This should make it easier to plan out the changes for better Windows support, and hopefully make other maintenance easier as well.

Last, but not least, there is a slight performance improvement in that export no longer does redundant timestamp processing or content hashing when neither the existing path nor timestamps have changed.  (This duplication was easy to spot and fix after the main refactoring was done, suggesting that maintenance has already become a little easier. :wink:)

(By the way, given the amount of deleted code and re-organization, you will probably find the diffs for cmd_exec and cmd_export *much* easier to review in side-by-side mode rather than unified.  Or better yet, just read the new versions directly!)

### DIFF
--- a/rc.go
+++ b/rc.go
@@ -8,6 +8,7 @@ import (
 	"io/ioutil"
 	"os"
 	"os/exec"
+	"os/signal"
 	"path/filepath"
 	"strings"
 	"time"
@@ -132,14 +133,12 @@ const notAllowed = "%s is blocked. Run `direnv allow` to approve its content"
 // Load evaluates the RC file and returns the new Env or error.
 //
 // This functions is key to the implementation of direnv.
-func (rc *RC) Load(ctx context.Context, config *Config, env Env) (newEnv Env, err error) {
+func (rc *RC) Load(previousEnv Env) (newEnv Env, err error) {
+	config := rc.config
 	wd := config.WorkDir
 	direnv := config.SelfPath
-	newEnv = env.Copy()
+	newEnv = previousEnv.Copy()
 	newEnv[DIRENV_WATCHES] = rc.times.Marshal()
-	defer func() {
-		rc.RecordState(env, newEnv)
-	}()
 
 	if !rc.Allowed() {
 		err = fmt.Errorf(notAllowed, rc.Path())
@@ -157,6 +156,16 @@ func (rc *RC) Load(ctx context.Context, config *Config, env Env) (newEnv Env, er
 		direnv,
 		rc.Path(),
 	)
+
+	// Allow RC loads to be canceled with SIGINT
+	ctx, cancel := context.WithCancel(context.Background())
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
+	go func() {
+		<-c
+		cancel()
+	}()
+
 	cmd := exec.CommandContext(ctx, config.BashPath, "--noprofile", "--norc", "-c", arg)
 	cmd.Dir = wd
 	cmd.Env = newEnv.ToGoEnv()
@@ -188,13 +197,9 @@ func (rc *RC) Load(ctx context.Context, config *Config, env Env) (newEnv Env, er
 		newEnv = newEnv2
 	}
 
-	return
-}
-
-// RecordState just applies the new environment
-func (rc *RC) RecordState(env Env, newEnv Env) {
 	newEnv[DIRENV_DIR] = "-" + filepath.Dir(rc.path)
-	newEnv[DIRENV_DIFF] = env.Diff(newEnv).Serialize()
+	newEnv[DIRENV_DIFF] = previousEnv.Diff(newEnv).Serialize()
+	return
 }
 
 /// Utils
--- a/rc.go
+++ b/rc.go
@@ -139,6 +139,11 @@ func (rc *RC) Load(previousEnv Env) (newEnv Env, err error) {
 	direnv := config.SelfPath
 	newEnv = previousEnv.Copy()
 	newEnv[DIRENV_WATCHES] = rc.times.Marshal()
+	defer func() {
+		// Record directory changes even if load is disallowed or fails
+		newEnv[DIRENV_DIR] = "-" + filepath.Dir(rc.path)
+		newEnv[DIRENV_DIFF] = previousEnv.Diff(newEnv).Serialize()
+	}()
 
 	if !rc.Allowed() {
 		err = fmt.Errorf(notAllowed, rc.Path())
@@ -180,25 +185,17 @@ func (rc *RC) Load(previousEnv Env) (newEnv Env, err error) {
 		cmd.Stdin = os.Stdin
 	}
 
-	out, err := cmd.Output()
-	if err != nil {
-		return
-	}
-
-	if len(out) > 0 {
+	if out, err := cmd.Output(); err == nil && len(out) > 0 {
 		var newEnv2 Env
 		newEnv2, err = LoadEnvJSON(out)
-		if err != nil {
-			return
+		if err == nil {
+			newEnv = newEnv2
 		}
 		if newEnv2["PS1"] != "" {
 			logError("PS1 cannot be exported. For more information see https://github.com/direnv/direnv/wiki/PS1")
 		}
-		newEnv = newEnv2
 	}
 
-	newEnv[DIRENV_DIR] = "-" + filepath.Dir(rc.path)
-	newEnv[DIRENV_DIFF] = previousEnv.Diff(newEnv).Serialize()
 	return
 }
 
--- a/cmd_exec.go
+++ b/cmd_exec.go
@@ -53,8 +53,7 @@ func cmdExecAction(env Env, args []string, config *Config) (err error) {
 
 	// Load the rc
 	if toLoad := findUp(rcPath, ".envrc"); toLoad != "" {
-		newEnv, err = config.EnvFromRC(toLoad, previousEnv)
-		if err != nil {
+		if newEnv, err = config.EnvFromRC(toLoad, previousEnv); err != nil {
 			return
 		}
 	} else {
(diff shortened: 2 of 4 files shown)


zimbatm:
I need to do some manual testing on this one. Let me know once the CI has turned green.

pjeby:
Ugh.  The linter is really picky.  Or really, I'm just not fluent enough in Go to write things the way it likes.  Guess I should install a local copy; I've been running local tests without it.

(Huh.  Looks like I've just installed a newer version that's even pickier; it's complaining about your code, too.  :wink:)

Anyway, it's green now.  Enjoy!

zimbatm:
Thanks!

I noticed that direnv doesn't emit the diff after a failed state. This is needed so that direnv doesn't retry the failure after every prompt.

To reproduce, use a fresh shell with no hooks installed. Create an `.envrc` with content `exit 1` and then run `direnv allow` and `direnv export bash`. On master, DIRENV_DIFF and DIRENV_WATCHES get emitted.

pjeby:
Ah, that explains the bits that didn't make sense to me, because it seemed like they would ignore errors or never happen.  :wink:  I'll go ahead and put them back; can you add a test to master for that case, and I'll rebase to match?

zimbatm:
:+1: waiting on CI for merging https://github.com/direnv/direnv/pull/637

pjeby:
I've rebased, so it's ready when you are.

rc.go
```diff
@@ -157,6 +161,16 @@ func (rc *RC) Load(ctx context.Context, config *Config, env Env) (newEnv Env, er
 		direnv,
 		rc.Path(),
 	)
+
+	// Allow RC loads to be canceled with SIGINT
+	ctx, cancel := context.WithCancel(context.Background())
+	c := make(chan os.Signal, 1)
+	signal.Notify(c, os.Interrupt)
```
zimbatm, line 168: When `signal.Notify` is called, it changes the runtime. Once the function is exited, the runtime is still configured to send signals to that channel.

I think the original design was a bit cleaner in that regard since both were decoupled. Not a big issue in practice since rc.Load is always called once.

pjeby:
However you want to handle it is fine by me; I mainly moved it there so that:

1. It would always be active for load (direnv exec didn't support it)
2. The export command (that didn't do anything with the context) didn't have to carry it around just to pass it down several calling layers.

I didn't realize any cleanup was required, since I was just refactoring what was already there: the original code didn't turn off the notification, either.  :wink:

By the way, I think I've figured out how to get direnv to work properly on Windows, whether it's msys, cygwin, git bash, mingw, whatever:

By rewriting it in bash.  :wink:

I know, I know, speed, right?  Well, the thing is, the only part of direnv that's speed critical is the "find the .envrc and check the watched files" part of direnv export, and I've come up with a draft version of that which I did some speed testing on (multiple platforms), to see if it was a viable approach:

```shell
#!/usr/bin/env bash
realpath.dirname() { REPLY=.; ! [[ $1 =~ /+[^/]+/*$|^//$ ]] || REPLY="${1%${BASH_REMATCH[0]}}"; REPLY=${REPLY:-/}; }

direnv::has-changed() {
	local REPLY=$PWD/.; local watch wf ENVRC_TO_EXPORT=
	while [[ $REPLY != / ]] && realpath.dirname "$REPLY"; do
		[[ -f "${REPLY%/}/.envrc" ]] || continue; ENVRC_TO_LOAD=$REPLY; break
	done
	[[ $ENVRC_TO_EXPORT == "${DIRENV_DIR#-}" ]] || return 0
	for watch in "${!__direnv_watch_@}"; do
		wf=$DIRENV_WATCH_DIR/$watch
		[[ -e $wf && -e ${!watch} && ! "${wf}" -ot "${!watch}" && ! "${wf}" -nt "${!watch}" ]] ||
			return 0
	done
	return 1  # no change, everything matches
}

if [[ $0 == "$BASH_SOURCE" && "${DIRENV_WATCH_DIR-}" && ${1-} == export ]]; then
	direnv::has-changed || exit 0
fi
```

Basically, this is the top of a hypothetical bash version of direnv, that checks to see if any of the watched files have changed timestamps or if the target directory is different, and then aborts early if there's no change.  On Linux, the "nothing changed" time is within a few milliseconds of `direnv export`, and same for OS x.

On windows, there's about 10-20 milliseconds difference...  and sometimes it's *in favor of the bash version*!

That is, the variation in timing between runs is such that the golang version's average time is higher than the bash version's best time.  And if I cheat and pretend the `direnv::has-changed` function is included in `direnv hook bash`, then the times on all platforms smoke the golang version all to heck, at least when used in bash.  :grinning:

Yes, some other parts of it might be slower.  But I'm not really sure what, since direnv has to launch bash for most of the complex bits, and launch itself again to pipe a script to it, whereas a bash version of direnv would already have bash and the stdlib in memory when it comes time to do the actual envrc loading.  Also, there's the nice benefit of not having to distribute different binaries for different platforms.  :wink:  (And the DIRENV_DUMP_FILE_PATH thing can be replaced with an fd redirect again, since Windows bashes can do that just fine within their *nix emulations; the golang version can't because Windows doesn't have real fds.)

The hard parts of this would appear to be environment diffs, escaping for other shells, etc.  But I did some sketches of that stuff too, and it's quite doable even in bash 3.2.  I've previously written extremely fast JSON escapers in bash; making a fish or tcsh one doesn't seem that hard, while bash itself can do sh/bash/zsh escaping.  environment diffs are also really easy and doable in pure bash without external program use.

Compressing and base64 encoding can be done with external programs, though TBH the bash version would just be to include `__direnv_old_X` and or `__direnv_new_X` vars for each variable removed/changed, or added, respectively.  (As that is the natural format of an EnvDiff implemented in bash.)  Meanwhile, file watches can be implemented as touching files in a direnv data directory with generated tempnames of `__direnv_watch_XXXXXXXXXXX`, and a matching env var of the same name whose value is the file path being watched.

(Every supported platform has `touch -r`, and bash has built-in timestamp testing for the comparison, neatly avoiding the utter hell that is the incompatibility of the `stat` command on OS X, let alone the part where using `[[ ]]` is just a syscall away and actually using stat means forking and exec-ing.)

Anywho, it's something to think about for the future.  Doing this PR helped a lot in working it out, because until it was reorganized I couldn't be sure I understood the whole thing well enough to see how a bash version could actually do what needed to be done.

zimbatm:
Oh my :)

One of the historical reasons for writing direnv is that at the time, [RVM](http://rvm.io/) was the most popular ruby version switcher. That project was literally thousands of bash code piled onto each-other and impossible to understand. Version switchers became more popular and started existing for a number of other languages and I thought that it would be good to come up with a generic solution.

For direnv 3 I think I'd rather use an embedded scripting language and remove most of the bash code actually.

zimbatm:
Since we are talking out of the box, another idea would be to ship direnv with a static build of bash. And add all the sub-commands as extra builtins (or bash plugins). That would save one extra process invocation. The main advantage is that the version of bash could then be controlled by direnv.

pjeby:
Except that then you'd never get it to work properly on Windows, unless you bundled both an msys and cygwin version.  And you'd still have to solve the cross-system environment munging problem: PATH on cygwin, and *all* env vars containing anything resembling paths on msys/git bash.  (Assuming you could link msys or cygwin to Go code at all, which I have yet to find a single example on the internet of anybody successfully doing.)

The status of my previous experiment in getting direnv to work on Windows was that the "pass an extra argument" trick will get it to basically work on cygwin with some adjustment to the tests (and an inability to handle empty strings), but that on msys/git bash that trick breaks horribly.  AFAICT, the *only* way to get uncorrupted vars on msys (and empty vars on Windows, *period*) is to *pipe* the cygwin/msys environment into direnv.  That is, direnv can never trust the contents of the actual process environment outside of its own (non-empty, and obfuscated so msys doesn't munge them) state variables.

Basically you need the shell-side hook to be something like

* call some direnv command that's the first half of export, to see if the dir or watchlist indicate a change
* run, say, `bash -c 'declare -x' | direnv export-part-2 myshelltype` to run the second half of export, parsing the bash declarations to get a valid environment

I mean, technically you could do that by having direnv export simply output that second command when it detects a change, but you still have to have the *parent* shell output the enviroment or it's no good.

That realization is what prompted the, "well hell, might as well just write it in bash to begin with, then it's one executable, portable everywhere, and a lot fewer process swaps, no Go API worries, etc.".  Plus, direnv is already 25% bash by volume.  :wink:  (That's line count; it's >30% by character count.)

But if you prefer to use the more convoluted way to get windows to work, or to give up on windows altogether, those are certainly options that exist.  I'm just looking at it from the POV of, "what value is Go adding here?"  And AFAICT the answer is, "not much".  And for Windows support, it's an active detriment.

There is apparently an msys go compiler based on gccgo, but it [doesn't do cygwin](https://stackoverflow.com/a/61522751).  I also don't know if said compiler actually produces msys-linked programs, or if it's just meant to build mingw programs (which aren't the same thing).  That is, if it builds Go programs that think their target runtime is Windows, then it's useless.  And even if it didn't, then cygwin would still not be doable.

zimbatm:
@pjeby do you want to have a live chat to hash this out? https://calendly.com/zimbatm/

Let's get this PR merged for now.

pjeby:
Sure, but it doesn't look like our regularly available hours overlap nearly at all.  I'm usually going to bed when you're getting up and vice versa it looks like.

PR #641, opened 2020-05-26
Change XDG_CONFIG_DIR to XDG_CONFIG_HOME

### LINKED ISSUE #640: Migrating of data outputs incorrect XDG env var
**Describe the bug**
The output when migrating the direnv data says:

```sh
Migrating the allow data to the new location

The allowed .envrc permissions used to be stored in the XDG_CONFIG_DIR. It's
better to keep that folder for user-editable configuration so the data is
being moved to XDG_DATA_HOME.

moving /home/userA/.config/direnv/allow to /home/userA/.local/share/direnv/allow
creating a symlink back from /home/userA/.local/share/direnv/allow to /home/userA/.config/direnv/allow for back-compat.

All done, have a nice day!
```

The env var is not `XDG_CONFIG_DIR` but rather `XDG_CONFIG_HOME`.

thernstig:
It was incorrect in the output when migrating direnv data.

Fixes #640

### DIFF
--- a/cmd_allow.go
+++ b/cmd_allow.go
@@ -17,7 +17,7 @@ var CmdAllow = &Cmd{
 var migrationMessage = `
 Migrating the allow data to the new location
 
-The allowed .envrc permissions used to be stored in the XDG_CONFIG_DIR. It's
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks!

PR #664, opened 2020-08-31
man: regenerate direnv-stdlib.1 and direnv.toml.1

fredrikekre:
Missed this in https://github.com/direnv/direnv/pull/661

Is there a reason these auto-generated files are included in the repo at all? Seems like the `man/*.1` pages, and `stdlib.go` could just be `.gitignored`?

### DIFF
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -225,6 +225,10 @@ A semantic dispatch used to describe common project layouts.
 .PP
 Sets the GOPATH environment variable to the current directory.
 
+.SS \fB\fClayout julia\fR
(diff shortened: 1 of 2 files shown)


zimbatm:
This is a requirement for users that `go get` the repository. I agree that it's not great and should probably add a CI check for it.

I guess this is not needed since #665 has been merged?

fredrikekre:
> This is a requirement for users that go get the repository. I agree that it's not great and should probably add a CI check for it.

Okay, yea that is a good idea!

> I guess this is not needed since #665 has been merged?

Correct!

PR #665, opened 2020-08-31
stdlib: semver_search

fredrikekre:
`stdlib`: new function `semver_search` that search a directory for the highest version on SemVer format. This function is factored out from the implementation of `use_node` with the intention to be used by other `use_*` implementations.

Edit: 2abfb8660dd1830b4364962dff41a957504efc7f is the commit to review, the others are just autogenerated changes.


zimbatm:
thanks!

PR #666, opened 2020-08-31
stdlib: use_julia <version>

fredrikekre:
Note: This is pending https://github.com/direnv/direnv/pull/665 which I factored out from this branch to simplify reviewing.

`stdlib`: new function `use_julia` that loads the specified julia version. The implementation is similar to `use_node`. In particular it requires the user to specify the directory with installations with `$JULIA_VERSIONS` and for partial versions a `semver_search` is performed.

Edit: 1cfc446a9a6e9cc18d448a8167524e841f98e787 is the commit to review for this PR, the rest are autogenerated changes or changes to review in https://github.com/direnv/direnv/pull/665.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -767,6 +767,50 @@ use() {
   "use_$cmd" "$@"
 }
 
+# Usage: use julia [<version>]
+# Loads specified Julia version.
+#
+# Environment Variables:
+#
+# - $JULIA_VERSIONS (required)
+#   You must specify a path to your installed Julia versions with the `$JULIA_VERSIONS` variable.
+#
+# - $JULIA_VERSION_PREFIX (optional) [default="julia-"]
+#   Overrides the default version prefix.
+#
+use_julia() {
+  local version=${1:-}
+  local julia_version_prefix=${JULIA_VERSION_PREFIX:-julia-}
+  local search_version
+  local julia_prefix
+
+  if [[ -z ${JULIA_VERSIONS:-} || -z $version ]]; then
+    log_error "Must specify the \$JULIA_VERSIONS environment variable and a Julia version!"
+    return 1
+  fi
+
+  julia_prefix="${JULIA_VERSIONS}/${julia_version_prefix}${version}"
+
+  if [[ ! -d ${julia_prefix} ]]; then
+    search_version=$(semver_search "${JULIA_VERSIONS}" "${julia_version_prefix}" "${version}")
+    julia_prefix="${JULIA_VERSIONS}/${julia_version_prefix}${search_version}"
+  fi
+
+  if [[ ! -d $julia_prefix ]]; then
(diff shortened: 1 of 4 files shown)


fredrikekre:
CC @garrison and @tshort who took part in the discussion in https://github.com/direnv/direnv/pull/389.

stdlib.sh
```diff
@@ -733,6 +767,51 @@ use() {
...
+    return 1
+  fi
+
+  if [[ ! -x $julia_prefix/bin/julia ]]; then
+    log_error "Unable to load Julia binary (julia) for version ($version) in ($JULIA_VERSIONS)!"
+    return 1
+  fi
+
+  load_prefix "$julia_prefix"
+  # PATH_add "$julia_prefix/bin"
```
zimbatm, line 810: left-over comment
```suggestion
```

zimbatm (approved):
Brilliant. Clear use-case, documentation, and tests. What else could I ask for?

garrison:
I like it.

I am a bit curious what `$JULIA_VERSIONS/julia-master` looks like for you.  For me, it is a source checkout, which means I really need `$julia_prefix` to be `$JULIA_VERSIONS/julia-master/usr`.  Stated more directly, I think it would be awesome if the julia version in the directory could either be a compiled source repo or downloaded official binaries, and direnv could handle this transparently by looking for the julia binary in both possible locations (with and without `usr/`).

fredrikekre:
I was thinking that `julia-master` would be a prefix where you have `make install`ed to. Loading `usr` as a prefix wholesale is bad since it includes a bunch of other libraries and binaries that might interfere badly with the rest of your system (try `ls usr/bin` and `ls usr/lib`).

I also discussed with @staticfloat and he said that in Julia 1.6 and above it should be safe to load `usr` as a prefix, since the way Julia builds and install it's own dependencies will change. Perhaps we can add some autodetection like you suggest once Julia 1.6 is out.

If you only care about pointing `julia` to the compiled binary you can just add the following to `.envrc`, (rather than adding `use julia master`)

```
PATH_add /path/to/repo
```
since there is a symlink in the root of the repo to `usr/bin/julia`.

zimbatm:
ping me once the conversation has settled :)

fredrikekre:
@zimbatm I don't intend to change anything more. I might revisit https://github.com/direnv/direnv/pull/666#issuecomment-684072331 when Julia 1.6 has been released.

zimbatm:
alright!

fredrikekre:
Thanks for the fast review and merge! I noticed that you also tagged a new version, thanks! Out of curiousity, what is the timeline for uploading binaries to https://github.com/direnv/direnv/releases after a tag?

zimbatm:
The process is manual since the switch to GitHub Actions. There is a script in `script/dist-push` that does most of the work and should be included as part of `.github/workflows/release.yml` at some point.

fredrikekre:
Thanks again!

PR #667, opened 2020-09-02
stdlib/use_julia: fix a bug in parameter substitution

fredrikekre:
The first commit fixes some complaints when running with `-u` as is done in the tests:
```
stdlib.sh:354: !1: unbound variable
```
and
```
stdlib.sh:359: path_array[@]: unbound variable
```
The fix for the array expansion is taken from https://stackoverflow.com/a/61551944/5087136 which suggests this fix as the "only safe idiom". I am not 100% sure about this first commit, so please review carefully!

The second commit is just a stupid bug fix -- the empty version prefix should of course be used, just like https://github.com/direnv/direnv/blob/98033c305ef2fdd3eb7ad5a83974c24da8becd15/stdlib.sh#L861. Also adds tests to `use julia` (which is what prompted the first commit, since `path_add` errors otherwise).

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -780,7 +780,7 @@ use() {
 #
 use_julia() {
   local version=${1:-}
-  local julia_version_prefix=${JULIA_VERSION_PREFIX:-julia-}
+  local julia_version_prefix=${JULIA_VERSION_PREFIX-julia-}
   local search_version
   local julia_prefix
 
--- a/stdlib.go
+++ b/stdlib.go
@@ -783,7 +783,7 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"#\n" +
 	"use_julia() {\n" +
 	"  local version=${1:-}\n" +
-	"  local julia_version_prefix=${JULIA_VERSION_PREFIX:-julia-}\n" +
+	"  local julia_version_prefix=${JULIA_VERSION_PREFIX-julia-}\n" +
 	"  local search_version\n" +
 	"  local julia_prefix\n" +
 	"\n" +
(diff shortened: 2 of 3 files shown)


fredrikekre:
Friendly bump @zimbatm :)

zimbatm:
sorry, I have been mulling over this PR and failed to report back.

stdlib.sh
```diff
@@ -351,12 +351,12 @@ path_add() {
...
   # split existing paths into an array
   declare -a path_array
-  IFS=: read -ra path_array <<<"${!1}"
+  IFS=: read -ra path_array <<<"${!1-}"
   shift
 
   # prepend the passed paths in the right order
   for ((i = $#; i > 0; i--)); do
-    path_array=("$(expand_path "${!i}")" "${path_array[@]}")
+    path_array=("$(expand_path "${!i}")" ${path_array[@]+"${path_array[@]}"})
```
zimbatm, line 359: I'm not sure about this solution. It seems like it's for bash 3.0 who is on macOS.

fredrikekre: Did you take a look at the SO post in the OP? That seems to suggest that this works on Bash 3, 4, 5. I have only tested with Bash 4 though, and that seems to work fine.

Would you rather have me revert this change and just disable `set -u` for the tests that I added?

fredrikekre: > Would you rather have me revert this change and just disable set -u for the tests that I added?

I did this for now.

zimbatm: It looks like https://github.com/direnv/direnv/issues/675 is having a similar issue.

zimbatm:
thanks!

fredrikekre:
@zimbatm would it be possible to release 2.22.1? Would be nice to have this patch and #676 in a patch release. Thanks!

zimbatm:
Thanks for the reminder! https://github.com/direnv/direnv/releases/tag/v2.22.1

Issue #669, opened 2020-09-08
layout_go: Potential regression introduced with v2.21


ealves-pt:
**Describe the bug**

The release of v2.21 introduced a regression to setup using `layout go`. Specifically https://github.com/direnv/direnv/pull/622.

Now having a `.direnv/` folder became mandatory but when one exists the `$GOPATH/bin` is not properly added to the $PATH.

**To Reproduce**

> You need Go installed. I did this using go1.14.8

Setup
```
$ mkdir -p ~/Regression/go
$ echo "layout go" > ~/Regression/go
$ cd ~/Regression/go
$ direnv allow
$ go get -u github.com/direnv/direnv
```

direnv will be installed into `~/Regression/go/.direnv/go/bin` but I cannot run it because the bin path is not properly set in the $PATH.

**Expected behavior**

Since we did this changes to now force the usage of a `.direnv` and having the $GOPATH pointing there then we should make sure that we correctly add the bin to the $PATH as well.

So, doing `echo $PATH` one would find `/home/foobar/Regression/go/.direnv/go/bin` instead of finding `/home/foobar/Regression/go/bin`.

**Environment**
 - OS: Ubuntu 18.04
 - Shell: bash
 - Direnv version: 2.22.0

**Additional context**

Please let me know if I'm missing something here.

But I believe the PR tried to solve a subset of problems while causing other problems as a side-effect. I personally don't see the need for requiring `.direnv/` and then forcing the $GOPATH into `.direnv/go` as I never ran into any issues but if we are keeping this approach then I would say we should keep the consistency and set the the correct $PATH.

But preferences aside, I'm happy to submit a PR to change [this function](https://github.com/direnv/direnv/blob/master/stdlib.sh#L529) into:
```
layout_go() {
  path_add GOPATH "$(direnv_layout_dir)/go"
  PATH_add "$(direnv_layout_dir)/go/bin"
}
```

PR #670, opened 2020-09-08
stdlib: adds GOPATH/bin to PATH

ealves-pt:
As reported in #669

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -528,7 +528,7 @@ layout() {
 #
 layout_go() {
   path_add GOPATH "$(direnv_layout_dir)/go"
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks!

stdlib.sh
```diff
@@ -528,7 +528,7 @@ layout() {
 #
 layout_go() {
   path_add GOPATH "$(direnv_layout_dir)/go"
-  PATH_add bin
+  PATH_add "$(direnv_layout_dir)/go/bin"
```
fredrikekre, line 531: Perhaps the docs on L527 just above should be adjusted for this too?

PR #676, opened 2020-09-28
stdlib: use Bash 3.0-compatible array expansion

### LINKED ISSUE #675: layout go and strict_env results in a failure and doesn't set GOPATH
**Describe the bug**
When loading an `.envrc` with `strict_env = true` and `layout go` the `GOPATH` is not set and bash shows an error:

```
> direnv allow
direnv: loading ~/projects/gmailfilters/.envrc
/bin/bash:353: !1: unbound variable
direnv: export +IFS
```

**To Reproduce**
1. Set `strict_env = true` in `~/.config/direnv/direnv.toml`
2. Create an `.envrc` with `layout go`
3. `direnv allow`

**Expected behavior**

When I unset `strict_env`: 

```
❯ direnv allow
direnv: loading ~/projects/gmailfilters/.envrc
direnv: loading ~/projects/.envrc
direnv: export +GOPATH +PROJECT_MARKER_FILE ~PATH
```

**Environment**
 - OS: macOS Catalina
 - Shell: fish
 - Direnv version 2.20.0

**Additional context**

zimbatm:
When `set -u` is being used, Bash 3.0 tracks empty array as unused
variables.

There is a big table of compatible versions, with each their own errors:
https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u/61551944#61551944

Thanks to @fredrikekre for finding the root cause.

Fixes #675

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -351,12 +351,12 @@ path_add() {
   local path i var_name="$1"
   # split existing paths into an array
   declare -a path_array
-  IFS=: read -ra path_array <<<"${!1}"
+  IFS=: read -ra path_array <<<"${!1-}"
   shift
 
   # prepend the passed paths in the right order
   for ((i = $#; i > 0; i--)); do
-    path_array=("$(expand_path "${!i}")" "${path_array[@]}")
+    path_array=("$(expand_path "${!i}")" ${path_array[@]+"${path_array[@]}"})
   done
 
   # join back all the paths
@@ -419,9 +419,9 @@ path_rm() {
   results=()
 
   # iterate over path entries, discard entries that match any of the patterns
-  for path in "${path_array[@]}"; do
+  for path in ${path_array[@]+"${path_array[@]}"}; do
     discard=false
-    for pattern in "${patterns[@]}"; do
+    for pattern in ${patterns[@]+"${patterns[@]}"}; do
       if [[ "$path" == +($pattern) ]]; then
         discard=true
         break
--- a/stdlib.go
+++ b/stdlib.go
@@ -354,12 +354,12 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"  local path i var_name=\"$1\"\n" +
 	"  # split existing paths into an array\n" +
 	"  declare -a path_array\n" +
-	"  IFS=: read -ra path_array <<<\"${!1}\"\n" +
+	"  IFS=: read -ra path_array <<<\"${!1-}\"\n" +
 	"  shift\n" +
 	"\n" +
 	"  # prepend the passed paths in the right order\n" +
 	"  for ((i = $#; i > 0; i--)); do\n" +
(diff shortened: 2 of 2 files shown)


stdlib.sh
```diff
@@ -356,7 +356,7 @@ path_add() {
 
   # prepend the passed paths in the right order
   for ((i = $#; i > 0; i--)); do
-    path_array=("$(expand_path "${!i}")" "${path_array[@]}")
+    path_array=("$(expand_path "${!i}")" "${path_array[@]+${path_array[@]}}")
```
yermulnik, line 359: > There is a big table of compatible versions, with each their own errors:
> https://stackoverflow.com/questions/7577052/bash-empty-array-expansion-with-set-u/61551944#61551944

JFYI: following the link there's an explicit note re the syntax:
```
Quoting the full expansion instead of the nested array ("${arr[@]+${arr[@]}}"), which I would have expected to be roughly equivalent, is similarly unsafe in 4.2.
```

fredrikekre: In https://github.com/direnv/direnv/pull/667 I had this:
```suggestion
    path_array=("$(expand_path "${!i}")" ${path_array[@]+"${path_array[@]}"})
```
before dropping that commit.

zimbatm: oops, good catch

fredrikekre:
I can still see the error from #675 with this patch, even when including this which I had in #667 
```diff
$ git diff stdlib.sh
diff --git a/stdlib.sh b/stdlib.sh
index f63f82b..b21efe7 100755
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -351,7 +351,7 @@ path_add() {
   local path i var_name="$1"
   # split existing paths into an array
   declare -a path_array
-  IFS=: read -ra path_array <<<"${!1}"
+  IFS=: read -ra path_array <<<"${!1-}"
   shift
 
   # prepend the passed paths in the right order
```

zimbatm:
I think I finally got it right. Ideally, I would have bash 3.0 in my test environment but that's already quite good.

fredrikekre:
I am still getting
```
$ direnv allow .
direnv: loading ~/dev/direnv/.envrc
/bin/bash:353: !1: unbound variable
```
with `layout go` or `use julia ...` in `.envrc`.

```
$ bash --version
GNU bash, version 4.3.48(1)-release (x86_64-pc-linux-gnu)
Copyright (C) 2013 Free Software Foundation, Inc.
License GPLv3+: GNU GPL version 3 or later <http://gnu.org/licenses/gpl.html>

This is free software; you are free to change and redistribute it.
There is NO WARRANTY, to the extent permitted by law.
```

zimbatm:
@fredrikekre can you try again after running `eval "$(direnv hook bash)"`? typically, direnv will install the hook with an absolute path to direnv so it might not pick up the changes.

fredrikekre:
Yea that works, looks like I forgot to reload the shell when testing. Sorry for the noise!

zimbatm:
thanks!

PR #678, opened 2020-09-29
Clarify path to direnv.toml

nmattia:
From the documentation it wasn't quite clear where the `direnv.toml` was
being sourced from.


zimbatm:
merci!

PR #685, opened 2020-10-07
shell: Update Elvish hook to replace deprecated `explode`

Cottser:
https://elv.sh/blog/0.14.0-release-notes.html

I haven't manually tested this on Elvish 0.12 but was at least able to confirm that the `all` builtin exists in that version.

Without this change, opening a new shell results in a deprecation warning like the following:

```
deprecation: the "explode" command is deprecated; use "all" instead
/Users/USERNAME/.elvish/lib/direnv.elv, line 6: 			m = (explode $m)
```


zimbatm:
thanks!

PR #696, opened 2020-10-21
fix: handle links on Mac when using `allow`

### LINKED ISSUE #694: Pass a file path to the allow command on macOS doesn't handle correctly
Basically as the title says, I don't have a machine with macOS installed, so I can't reproduce this easily.
But, the following CI results may help you to reason:

- [Use direnv 2.23.0 for CI](https://github.com/asdf-community/asdf-direnv/commit/ab0dcb99cf6027fb7a6c3adabb54807e198e3457) - [CI result](https://github.com/asdf-community/asdf-direnv/runs/1282378776#step:6:21)
- [Use a temporary workaround to avoid macOS test fail](https://github.com/asdf-community/asdf-direnv/commit/7536047b9215c0015da65efb0de73937b8002380) - [CI result](https://github.com/asdf-community/asdf-direnv/runs/1282500065#step:6:18)

jason0x43:
At least on macOS, paths passed to `direnv allow <dir>` weren't being fully resolved during the `allow` process. For example, `direnv allow /tmp/dir` would store an authorization for /tmp/dir/.envrc. On Macs, /tmp is an alias for /private/tmp, and when the user actually cd-ed into /tmp/dir, the path provided by os.Getwd would be /private/tmp/dir, which resulted in a different hash.

Aliases aren't symlinks, but go's `EvalSymlinks` resolves them anyway.

Both the `Abs` and `EvalSymlinks` calls are required, as `Abs` doesn't resolve links and `EvalSymlinks` doesn't resolve relative paths.

resolves #694

### DIFF
--- a/cmd_allow.go
+++ b/cmd_allow.go
@@ -25,7 +25,12 @@ being moved to XDG_DATA_HOME.
 func cmdAllowAction(env Env, args []string, config *Config) (err error) {
 	var rcPath string
 	if len(args) > 1 {
-		rcPath = args[1]
+		if rcPath, err = filepath.Abs(args[1]); err != nil {
+			return err
+		}
+		if rcPath, err = filepath.EvalSymlinks(rcPath); err != nil {
+			return err
+		}
 	} else {
 		if rcPath, err = os.Getwd(); err != nil {
 			return err


zimbatm:
thanks a lot!

I suspect that this issue might crop up in other places inside of direnv, but that fixes the immediate problem.

smorimoto:
Thanks, @jason0x43. You worked on and fixed it much faster than I did :)

PR #697, opened 2020-10-21
add watch_dir command

evalexpr:
Adds a command named `watch_dir` that watches the directory passed to it
recursively

### DIFF
--- a/cmd_watch_dir.go
+++ b/cmd_watch_dir.go
@@ -1,0 +1,61 @@
+package main
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// CmdWatchDir is `direnv watch-dir SHELL PATH`
+var CmdWatchDir = &Cmd{
+	Name:    "watch-dir",
+	Desc:    "Recursively adds a directory to the list that direnv watches for changes",
+	Args:    []string{"SHELL", "DIR"},
+	Private: true,
+	Action:  actionSimple(watchDirCommand),
+}
+
+func watchDirCommand(env Env, args []string) (err error) {
+	if len(args) < 3 {
+		return fmt.Errorf("a directory is required to add to the list of watches")
+	}
+
+	shellName := args[1]
+	dir := args[2]
+
+	shell := DetectShell(shellName)
+
+	if shell == nil {
+		return fmt.Errorf("unknown target shell '%s'", shellName)
+	}
+
+	if _, err := os.Stat(dir); os.IsNotExist(err) {
+		return fmt.Errorf("dir '%s' does not exist", dir)
+	}
+
+	watches := NewFileTimes()
+	watchString, ok := env[DIRENV_WATCHES]
+	if ok {
+		err = watches.Unmarshal(watchString)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
+		return watches.NewTime(path, info.ModTime().Unix(), true)
+	})
+	if err != nil {
+		return fmt.Errorf("failed to recursively watch dir '%s': %v", dir, err)
+	}
+
+	e := make(ShellExport)
+	e.Add(DIRENV_WATCHES, watches.Marshal())
+
+	os.Stdout.WriteString(shell.Export(e))
+
+	return
+}
--- a/test/direnv-test-common.sh
+++ b/test/direnv-test-common.sh
@@ -92,6 +92,8 @@ test_start base
   direnv_eval
   echo "${HELLO}"
   test -z "${HELLO}"
+
+  unset WATCHES
 test_stop
 
 test_start inherit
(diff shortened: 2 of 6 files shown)


evalexpr:
The only way I could find to test this was to use `~/.config/direnv/direnvrc`

zimbatm:
:+1: for the idea, but the implementation should be recursive.

zimbatm:
I think something like this would do it, but it requires to also add `shopt -s globstar` to the script:  https://github.com/koalaman/shellcheck/wiki/Sc2044#if-find-is-just-matching-globs-recursively

evalexpr:
Hmm, it seems that the macos tests are using bash < 4.0 which doesn't include the shopt globstar. I think macos ships with zsh by default now but it'd be nice to make it work with that older version of bash too - maybe we need to switch to a `find` approach.

I've also tested this on a directory with 7000 files in, and it seemed to hang - are we concerned about users watching a 7000 file directory?

zimbatm:
Good points. The downside with the `find` approach is that it adds an external dependency, which is why I proposed globstar.

Alternatively, `watch_dir` could be implemented in Go directly. Go has a [path/filepath.Walk](https://golang.org/pkg/path/filepath/#Walk) function in the stdlib that should make the implementation relatively straight-forward. And it should be more efficient than spawning 7000 watch_file.

evalexpr:
Definitely faster

```
❯ time direnv reload
direnv reload  0.00s user 0.00s system 110% cpu 0.003 total

❯ ./direnv status | wc -l
7241
```

Any tips on a more thorough test would be appreciated :)

cmd_watch_dir.go
```diff
@@ -0,0 +1,68 @@
...
+		err = watches.Unmarshal(watchString)
+		if err != nil {
+			return err
+		}
+	}
+
+	err = filepath.Walk(dir, func(path string, info os.FileInfo, err error) error {
+		if err != nil {
+			return err
+		}
```
zimbatm, line 55:
```suggestion
if info.Isdir() {
  return nil
}
```
We should only be watching files as folders don't have meaningful mtimes.

cmd_watch_dir.go
```diff
@@ -0,0 +1,68 @@
...
+	"fmt"
+	"os"
+	"path/filepath"
+)
+
+// CmdWatchDir is `direnv watch-dir SHELL PATH`
+var CmdWatchDir = &Cmd{
+	Name:    "watch-dir",
+	Desc:    "Recursively adds a directory to the list that direnv watches for changes",
+	Args:    []string{"[SHELL]", "[DIR]"},
```
zimbatm, line 13:
```suggestion
	Args:    []string{"SHELL", "DIR"},
```
Let's make both arguments required to simplify the logic. Potentially we want to add a variable number of directories in the future to watch multiple folders at once.

zimbatm:
Almost ready!

If you want to unpack the gzenv format, direnv comes with a tool builtin: `direnv show_dump "$DIRENV_WATCHES"`

cmd_watch_dir.go
evalexpr: Is it not worth keeping directories as part of this? meaning that we'll reload `direnv` whenever we add/remove a file from the directory (or a subdirectory). I'm not sure how mtimes work so I'm probably wrong here :)

evalexpr: I have added this suggestion for now though

test/direnv-test-common.sh
```diff
@@ -201,6 +203,21 @@ test_start "failure"
...
+test_start "watch-dir"
+    echo "No watches by default"
+    test_eq "${DIRENV_WATCHES}" "${WATCHES}"
+
+    direnv_eval
+
+    if ! direnv show_dump "${DIRENV_WATCHES}" | grep -q "testfile"; then
+        echo "FAILED: testfile not added to DIRENV_WATCHES"
+        exit 1
+    fi
```
evalexpr, line 215: I wasn't sure about adding this here, is it worth it?

cmd_watch_dir.go
zimbatm: Actually, I don't know.

I assumed that the directory didn't have a meaningful mtime. But now that I am testing it locally, it looks like it *does* update whenever a file gets created. So I might be wrong. It looks like it works on Linux, but I don't know if it's portable across POSIX.

test/direnv-test-common.sh
zimbatm: Yeah, sure! Thanks for adding tests.

test/direnv-test-common.sh
```diff
@@ -92,6 +92,8 @@ test_start base
   direnv_eval
   echo "${HELLO}"
   test -z "${HELLO}"
+
+  unset WATCHES
```
zimbatm, line 96:
```suggestion
  unset DIRENV_WATCHES
```
Is that what you meant?

cmd_watch_dir.go
yermulnik: Not only it does but it also has to: https://serverfault.com/a/388100 (and yep looks like POSIX is what IEEE 1003 is).

test/direnv-test-common.sh
evalexpr: Nope, we set `WATCHES` on line 77:

```
WATCHES=$DIRENV_WATCHES
```

Meaning that we need to unset it at the end of this test I think. Sorry I've not had much time to look at this PR

zimbatm:
@evalexpr do you mind checking why the checks are failing?

evalexpr:
@zimbatm I've left a comment on the previous review comment: https://github.com/direnv/direnv/pull/697#discussion_r537421210 as to what I _think_ the reason is.

I will have time to look into it further this weekend if necessary - after that we can decide whether we want to support watching the directory itself too. Do we still think that is a bad idea?

Having the directory as part of the watches seemed to work for me locally (again I can try testing on macOS this weekend - maybe we can merge without that and create a follow-up PR)

evalexpr:
@zimbatm I have opted to remove the `isDir` check due to the fact that it looks like we _can_ rely on mtimes for directories. I have tested this on both NixOS and macOS by creating a file and directory inside both the top level of a directory tracked via `watch_dir` as well as deep in a subdirectory, all of these scenarios cause direnv to reload which is a desirable outcome I think.

This should be good to go now :)

zimbatm:
thanks for doing this, I already reached the same conclusion and was going to remove IsDir as well :)

PR #702, opened 2020-10-30
Add on_git_branch command to detect whether a specific git branch is checked out

brianrodri:
When working on GitHub projects, I've found it helpful to have env variables set-up when I'm inside a specific branch. For example, sometimes I want to change the merge-base my editor uses to display signifiers. I can accomplish this with the new command like so:

```bash
if on_git_branch child_changes; then
  export MERGE_BASE_BRANCH=parent_changes
fi
```

To detect movement between branches, the `.git/HEAD` file is watched with `watch_file`. This is only done when a branch name is specified, otherwise no calls to `watch_file` are made.


stdlib.sh
```diff
@@ -978,6 +978,37 @@ direnv_version() {
...
+#    if on_git_branch develop; then
+#      echo "Remember to merge with upstream regularly!"
+#    fi
+#
+#    if on_git_branch; then
+#      echo "Thanks for contributing to a GitHub project!"
+#    fi
+#
+on_git_branch() {
+  if ! has git; then return 1; fi
```
zimbatm, line 1002: how about also adding a message like `log_error "on_git_branch needs git, which was not found on your system"`

stdlib.sh
```diff
@@ -978,6 +978,37 @@ direnv_version() {
...
+  local git_dir
+  if git_dir=$(git rev-parse --absolute-git-dir 2> /dev/null); then
+    [ -n "$1" ] && watch_file "$git_dir/HEAD"
```
zimbatm, line 1005: This line return 1 if "$1" is empty. Is that what you want?

stdlib.sh
```diff
@@ -978,6 +978,37 @@ direnv_version() {
...
+  else
```
zimbatm, line 1006: maybe add an error message as well here explaining that a .git directory was not found?

zimbatm (commented):
sounds good!

Can you also document the function in man/direnv-stdlib.1.md ?

stdlib.sh
brianrodri: Done.

stdlib.sh
brianrodri: Done.

stdlib.sh
brianrodri: Hmm nope, I've changed it to an if branch. Thanks!

brianrodri (commented):
Addressed comments + added entry to `man/direnv-stdlib.1.md`, PTAL!

brianrodri:
@zimbatm note that I've run `make fmt` on my changes, let me know if you'd prefer I revert that.

zimbatm:
It looks like the branch is ready?

brianrodri:
Yup, I'm happy with it if you are.

zimbatm:
thanks!

PR #723, opened 2020-11-26
Fix proposal for  #707, broken direnv compatibility under Windows

### LINKED ISSUE #707: direnv: referenced .envrc does not exist while executing direnv allow .
**Describe the bug**

I'm not able to pass the [quick-demo](https://github.com/direnv/direnv#quick-demo). After the `direnv allow .` step I got the following error:
```
direnv: referenced C:\Users\goost\direnv-test\.envrc does not exist
```

**To Reproduce**

* Install latest version by `scoop install direnv`
* Execute the `direnv hook bash | clip` command
* Append the clipboard content to the `~/.bashrc`
* Execute the commands:
```
mkdir direnv-test && pushd direnv-test
echo export FOO=foo > .envrc
direnv allow .
```

**Actual behaviour**
```
direnv: referenced C:\Users\goost\direnv-test\.envrc does not exist
```

**Expected behaviour**
Direnv hook should load the created `.envrc` file from the current directory and execute it.

**Environment**
 - OS: `Windows 10.0.19042 N/A Build 19042`
 - Shell: `bash`
 - Direnv version `2.23.1`

**Additional context**

image

pjos:
Tests are passing under Linux/Ubuntu as far as I could see. 
Resolves #707

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -302,13 +302,9 @@ find_up() {
 #
 # NOTE: the other ".envrc" is not checked by the security framework.
 source_env() {
-  local rcpath
-
-  if [[ "$1" =~ ^[a-zA-Z]:.* ]]; then
-    # Go will pass native windows path when executed under Git Bash/Git for Windows
-    rcpath=$(type cygpath > /dev/null 2>&1 && cygpath -u "${1/#\~/$HOME}" 2> /dev/null)
-  else
-    rcpath=${1/#\~/$HOME}
+  local rcpath=${1/#\~/$HOME}
+  if has cygpath ; then
+    rcpath=$(cygpath -u "$rcpath")
   fi
 
   local REPLY
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -302,7 +302,15 @@ find_up() {
 #
 # NOTE: the other ".envrc" is not checked by the security framework.
 source_env() {
-  local rcpath=${1/#\~/$HOME}
+  local rcpath
+
+  if [[ "$1" =~ ^[a-zA-Z]:.* ]]; then
+    # Go will pass native windows path when executed under Git Bash/Git for Windows
+    rcpath=$(type cygpath > /dev/null 2>&1 && cygpath -u "${1/#\~/$HOME}" 2> /dev/null)
+  else
+    rcpath=${1/#\~/$HOME}
+  fi
+
   local REPLY
   if [[ -d $rcpath ]]; then
     rcpath=$rcpath/.envrc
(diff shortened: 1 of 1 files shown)


stdlib.sh
```diff
@@ -302,7 +302,15 @@ find_up() {
 #
 # NOTE: the other ".envrc" is not checked by the security framework.
 source_env() {
-  local rcpath=${1/#\~/$HOME}
+  local rcpath
+
+  if [[ "$1" =~ ^[a-zA-Z]:.* ]]; then
+    # Go will pass native windows path when executed under Git Bash/Git for Windows
+    rcpath=$(type cygpath > /dev/null 2>&1 && cygpath -u "${1/#\~/$HOME}" 2> /dev/null)
```
zimbatm, line 309: What does the `cygpath -u` command do?

If `type cygpath` returns false then the rcpath will be empty isn't it?

pjos: https://cygwin.com/cygwin-ug-net/cygpath.html, converts local Windows path to corresponding  unix path. 
`cygpath` is available on cygwin like "bash" environment distributions for Windows. I don't want to assume it's existence, checked the most poppular https://gitforwindows.org/ & https://cmder.net/ full distribution. 

Yes, the `rcpath` path will be empty if `cygpath` doesn't exists, I could write out some information with `log_status` if that happens. If this shows to be an issue in the future, the idea is to replace `cygwin` path conversion with a similar construction like the current one for *nix systems 
https://github.com/direnv/direnv/blob/0c2afcc9fb9aec55c714844ce9760d5b43eb960d/stdlib.sh#L205-L220

zimbatm: What do you think of always running it through cygpath, but only if it exists?

```
local rcpath=${1/#\~/$HOME}
if has cygpath ; then
  rcpath=$(cygpath -u "$rcpath")
fi
```

pjos: `Command 'has' not found,` 
on ubuntu and msys ... needs probably be something like this 
```
if type cygpath > /dev/null 2>&1; then 
  rcpath=$(cygpath -u "$rcpath")
fi
```
Otherwise I like the construction, i need to run some scenarios and update this PR if it works as expected.

zimbatm: [has](https://github.com/direnv/direnv/blob/0c2afcc9fb9aec55c714844ce9760d5b43eb960d/stdlib.sh#L166-L179) is part of the stdlib :)

goostleek:
It does not fix the problem. Please see [the comment](https://github.com/direnv/direnv/issues/707#issuecomment-742446890) in the original issue.

PR #725, opened 2020-11-27
Fixing fish shell hook

### LINKED ISSUE #583: fish: direnv is no longer triggered on shell creation
**Describe the bug**
direnv does not execute on creation of a new fish shell (or tmux pane or window, etc.) -- I have to hit 'enter' once to trigger it. Triggers fine when changing directories.

**To Reproduce**
1. `bash`
2. `cd` into a directory with a whitelisted `.envrc` file
3. `fish`

Steps to reproduce the behaviour via tmux:
1. Use fish
2. Use tmux, with either of the [recommended configurations from the wiki](https://github.com/direnv/direnv/wiki/Tmux)
3. `cd` into a directory with a whitelisted `.envrc` file
4. Create a new tmux pane or window

**Expected behavior**
direnv should immediately trigger and load the environment in the new fish shell / tmux pane/window. It does not.

**Environment**
 - OS: nixos 20.03.git.a21c2fa (Markhor)
 - Shell: fish 3.0.2
 - Direnv version: 2.21.2

**Additional context**
This issue was essentially caused by #512, which solved a smaller issue (direnv not un/loading when changing directories using alt-left/right), and then later this was **not** fixed by #549, which solved a larger issue cased by #512 (a general 1-command "lag" on direnv activation when using fish).

Incidentally: that first issue is *also* back after #549; direnv no longer un/loads when changing directories in fish using alt-left/right. I feel like you need regression tests, although unhelpfully I have no clue how you'd instrument them for something like this :P.

if `fish_preexec` is the only thing that fires on changing directories with alt-left/right, and `fish_prompt` is the only thing that fires on creation of a new shell, then it seems like perhaps *both* need to be hooked?

### LINKED ISSUE #614: direnv reload on new shell
If I go to a directory for the first time, direnv properly loads my `.envrc` content. But if I open a new shell directly with e.g. `fish`, while inside the directory I just went to, it does not re-load my `.envrc` content. The problem with this is that if I have paths prepends done in `.config/fish/config.fish`, they they take precedence over my project specific `PATH_add`.

Wouldn it be a good default for direnv to reload the env if launching a new shell from current shell?

I could add a `direnv reload` to the bottom of my `.config/fish/config.fish`, but I feel the above default might make more sense. I might be missing something.

dimitarnestorov:
Closes #583 
Closes #614

### DIFF
--- a/docs/hook.md
+++ b/docs/hook.md
@@ -29,7 +29,8 @@ eval "$(direnv hook zsh)"
 Add the following line at the end of the `~/.config/fish/config.fish` file:
 
 ```sh
-eval (direnv hook fish)
+direnv hook fish | source
+direnv export fish | source
 ```
 
 ## TCSH


zimbatm:
hi, can you explain a little bit what the change does?

If `direnv hook fish` is needed on initialization, it could be incorporated into the [hook code](https://github.com/direnv/direnv/blob/0c2afcc9fb9aec55c714844ce9760d5b43eb960d/shell_fish.go#L13-L17)

zimbatm:
closed by #732

PR #727, opened 2020-11-30
Added use_nodenv to stdlib

jamesboehmer:
Similar to `layout_pyenv`, this relies on the existence of `nodenv` and its `versions` directory, e.g. `$(nodenv root)/versions`.

Unlike `layout_pyenv`, this takes only one argument and adds a single version to the `PATH`.

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -692,32 +692,6 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"#\n" +
 	"# Uses use_node and layout_node to add the chosen node version and \n" +
 	"# \"$PWD/node_modules/.bin\" to the PATH\n" +
-	"#\n" +
-	"# Usage: layout nodenv <node version number>\n" +
-	"#\n" +
-	"# Example:\n" +
-	"#\n" +
-	"#    layout nodenv 15.2.1\n" +
-	"#\n" +
-	"# Uses use_node and layout_node to add the chosen node version and \n" +
-	"# \"$PWD/node_modules/.bin\" to the PATH\n" +
-	"#\n" +
-	"layout_nodenv() {\n" +
-	"  local node_version=\"${1}\"\n" +
-	"  local node_versions\n" +
-	"  local nodenv_version\n" +
-	"  node_versions=\"$(nodenv root)/versions\"\n" +
-	"  nodenv_version=${node_versions}/${node_version}\n" +
-	"  if [[ -e \"$nodenv_version\" ]]; then\n" +
-	"      # Put the selected node version in the PATH\n" +
-	"      NODE_VERSIONS=\"${node_versions}\" NODE_VERSION_PREFIX=\"\" use_node \"${node_version}\"\n" +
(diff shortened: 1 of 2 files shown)


stdlib.sh
```diff
@@ -687,6 +687,32 @@ layout_node() {
...
+  local nodenv_version
+  node_versions="$(nodenv root)/versions"
+  nodenv_version=${node_versions}/${node_version}
+  if [[ -e "$nodenv_version" ]]; then
+      # Put the selected node version in the PATH
+      NODE_VERSIONS="${node_versions}" NODE_VERSION_PREFIX="" use_node "${node_version}"
+      # Add $PWD/node_modules/.bin to the PATH
+      layout_node
+  else
+    log_error "nodenv: version '$node_version' not installed"
```
zimbatm, line 711: What do you think of adding the command needed to install the version as well?

stdlib.sh
```diff
@@ -687,6 +687,32 @@ layout_node() {
...
+#
+# Uses use_node and layout_node to add the chosen node version and 
+# "$PWD/node_modules/.bin" to the PATH
+#
+layout_nodenv() {
+  local node_version="${1}"
+  local node_versions
+  local nodenv_version
+  node_versions="$(nodenv root)/versions"
+  nodenv_version=${node_versions}/${node_version}
```
zimbatm, line 704:
```suggestion
  node_versions_dir="$(nodenv root)/versions"
  nodenv_version=${node_versions_dir}/${node_version}
```

zimbatm (commented):
Sounds good!

It should be named `use_nodenv`. The `use` prefix is for loading dependencies, whereas the `layout` prefix is to setup a typical project layout. `layout_pipenv` was a mistake...

stdlib.sh
jamesboehmer: Good suggestion.  Added instructions at https://github.com/jamesboehmer/direnv/blob/2151074144f76d226e61adec835e7474de625413/stdlib.sh#L1098

stdlib.sh
jamesboehmer: I renamed the variable and also quoted it properly https://github.com/jamesboehmer/direnv/blob/2151074144f76d226e61adec835e7474de625413/stdlib.sh#L1088-L1094

stdlib.go
```diff
@@ -684,6 +684,15 @@ const StdLib = "#!/usr/bin/env bash\n" +
...
 	"\n" +
+	"# Usage: layout nodenv <node version number>\n" +
+	"#\n" +
+	"# Example:\n" +
+	"#\n" +
+	"#    layout nodenv 15.2.1\n" +
+	"#\n" +
+	"# Uses use_node and layout_node to add the chosen node version and \n" +
+	"# \"$PWD/node_modules/.bin\" to the PATH\n" +
+	"\n" +
```
zimbatm, line 695: Last thing, can you re-generate the stdlib.go?

jamesboehmer: My mistake.  This is cleaned up now.

PR #732, opened 2020-12-14
updated fish hook support issue

### LINKED ISSUE #583: fish: direnv is no longer triggered on shell creation

wderezin:
closes #583 

I updated the `fish hook` to handle the 3 use cases that @pjeby mentioned through an environment variable

```fish
set -g direnv_fish_mode eval_on_arrow    # trigger direnv at prompt, and on every arrow-based directory change (default)
set -g direnv_fish_mode eval_after_arrow # trigger direnv at prompt, and only after arrow-based directory changes before executing command
set -g direnv_fish_mode disable_arrow    # trigger direnv at prompt only, this is similar functionality to the original behavior
```

The `disable_arrow` option is similar to the original functionality with no support for the arrow directory change.  However, the hook is on `fish_prompt` instead of `fish_postexec` which for most use cases should provide the same functionality.

Also, update the docs to use recommended `direnv hook fish | source` format when loading the hook.

### DIFF
--- a/man/direnv.1
+++ b/man/direnv.1
@@ -89,11 +89,27 @@ eval "$(direnv hook zsh)"
 .PP
 Add the following line at the end of the \fB\fC\~/.config/fish/config.fish\fR file:
 
+
 .PP
 .RS
 
 .nf
 eval (direnv hook fish)
+
+.fi
+.RE
+
+.PP
+Fish supports 3 modes you can set with with the global environment variable \fB\fCdirenv_fish_mode\fR :
+
+
+.PP
+.RS
+
+.nf
+set -g direnv_fish_mode eval_on_arrow    # trigger direnv at prompt, and on every arrow-based directory change (default)
+set -g direnv_fish_mode eval_after_arrow # trigger direnv at prompt, and only after arrow-based directory changes before executing command
+set -g direnv_fish_mode disable_arrow    # trigger direnv at prompt only, this is similar functionality to the original behavior
 
 .fi
 .RE
--- a/man/direnv.1.md
+++ b/man/direnv.1.md
@@ -80,8 +80,17 @@ eval "$(direnv hook zsh)"
 Add the following line at the end of the `~/.config/fish/config.fish` file:
 
 ```fish
-eval (direnv hook fish)
+direnv hook fish | source
 ```
+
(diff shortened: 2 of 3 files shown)


zimbatm:
Sounds good. Anyone else wants to test this before I merge?

zimbatm:
Talking about testing, is there any way to test this programmatically?

wderezin:
If someone can point me to how to automate the arrow change directory I could probably create some test cases.

zimbatm:
It would require a full fake-shell test driver, which is too much to ask for this PR. Is there a subset that can be reasonably tested? Otherwise, if you think that this PR is ready, I will just merge it.

zimbatm:
I'm going to make a release so it's a good time to merge this

PR #735, opened 2020-12-21
fix: create temp dir in current working dir for one test

### LINKED ISSUE #729: test/stdlib.bash fails if /tmp is noexec
**Describe the bug**
* Running `./test/stdlib.bash` fails with exit status 1, with no output.

```sh
$ ./test/stdlib.bash
--- find_up
--- source_up
direnv: loading ~/direnv/test/.envrc
--- direnv_apply_dump
--- PATH_rm
--- path_rm
--- expand_path
--- semver_search
--- use_julia
$ echo $?
1
```

**To Reproduce**
* `sudo mount -o remount,noexec /tmp`
* `./test/stdlib.bash`

**Expected behavior**
* At least display something, rather than exiting with no output.

**Environment**
 - OS: Arch Linux
 - Shell: zsh
 - Direnv version: 2.25.0

**Additional context**
* /tmp is mounted with noexec option
    > `$ mount | grep -w \/tmp`
    > tmpfs on /tmp type tmpfs (rw,nosuid,noexec,noatime,inode64)
* workaround: set $TMPDIR as described in [mktemp.1](https://jlk.fjfi.cvut.cz/arch/manpages/man/core/coreutils/mktemp.1.en)
    > `$ TMPDIR=$XDG_RUNTIME_DIR ./test/stdlib.bash`
* To see an actual error message to figure out why it was failing, I had to run `bash -x ./test/stdlib.bash`

etcusrvar:
Fixes #729

### DIFF
--- a/test/stdlib.bash
+++ b/test/stdlib.bash
@@ -105,7 +105,7 @@ test_name semver_search
 test_name use_julia
 (
   load_stdlib
-  JULIA_VERSIONS=$(mktemp -d)
+  JULIA_VERSIONS=$(mktemp -d -p .)
   trap 'rm -rf $JULIA_VERSIONS' EXIT
 
   test_julia() {
--- a/test/stdlib.bash
+++ b/test/stdlib.bash
@@ -105,7 +105,7 @@ test_name semver_search
 test_name use_julia
 (
   load_stdlib
-  JULIA_VERSIONS=$(mktemp -d -p .)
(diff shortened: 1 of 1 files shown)


zimbatm (commented):
thanks.

It looks like it broke the MacOS test. I guess the `-p` option is not available in the BSD `mktemp` that is being shipped with it.

zimbatm:
thanks!

PR #754, opened 2021-01-27
Add/update fish tests

etcusrvar:
I went through direnv-test-common.sh and added the same tests to direnv-test.fish.

I used the same naming and conventions as the tests and utility functions currently there.
I also ran fish_indent to standardize the formatting ([4 spaces](https://github.com/fish-shell/fish-shell/blob/ca3d22665925421f7f96e05890b294d39a0b064a/src/fish_indent.cpp#L51-L53)).
The begin's and end's are not strictly needed, but help with the formatting and look and feel.

I also looked at the currently commented out 'utf-8' test, but wasn't able to figure out why it's failing after some investigation, so I left it commented out.

### DIFF
--- a/test/direnv-test.fish
+++ b/test/direnv-test.fish
@@ -11,6 +11,22 @@ function test_eq --argument-names a b
     end
 end
 
+function test_neq --argument-names a b
+    if not test (count $argv) = 2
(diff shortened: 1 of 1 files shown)


zimbatm:
awesome. thanks a lot!

PR #759, opened 2021-02-12
man/direnv.1.md: add FILES section (fix #758)

### LINKED ISSUE #758: Please mention `XDG_DATA_HOME` location in direnv(1)
**Describe the bug**
Some files sensibly were moved from `XDG_CONFIG_HOME` to `XDG_DATA_HOME`; it would be good if you could mention this location in a `FILES` section in direnv(1), so that it's easy for users to find.

Thanks for direnv! Been meaning to try it for ages, why did I wait so long???

rrthomas:
Also fix some other references that ought to be to `$XDG_CONFIG_HOME` rather than `~/.config`.

### DIFF
--- a/man/direnv.1
+++ b/man/direnv.1
@@ -1,3 +1,4 @@
+.nh
 .TH DIRENV 1 "2019" direnv "User Manuals"
 .SH NAME
 .PP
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks!

PR #767, opened 2021-02-21
test: Fix errors for elvish test

punitagrawal:
In recent versions of elvish the builtin path-dir has been moved to a
module. Update the test script to use the module to avoid a warning
about deprecated usage. The move of "path" to the module also seems to
have updated the key "path" to "name" - reflect this in the test
script.

Similarly, temporary assignment now requires using the "set"
keyword. Fix this while at it.


zimbatm:
thanks!

PR #770, opened 2021-03-01
pipenv --python $PYTHON_VERSION

yssource:
`⠙ Creating virtual environment...FileNotFoundError: [Errno 2] No such file or directory: '/shims/python'`

Above could happen, if it is to use python from pyenv, instead of the one from system.

### DIFF
--- a/stdlib.go
+++ b/stdlib.go
@@ -867,9 +867,13 @@ const StdLib = "#!/usr/bin/env bash\n" +
 	"    exit 2\n" +
 	"  fi\n" +
 	"\n" +
+	"  PYTHON_VERSION=$(python --version | cut -d\" \" -f2)\n" +
 	"  VIRTUAL_ENV=$(pipenv --venv 2>/dev/null ; true)\n" +
 	"\n" +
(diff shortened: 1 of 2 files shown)


zimbatm:
What does `pipenv --python $python_version` do exactly? It seems a bit racey if it switches the python version on the system or user-level.

yssource:
according to the pipenv document
1. before installing dependent packages, create a project.
```
Usage Examples:
   Create a new project using Python 3.7, specifically:
   $ pipenv --python 3.7

```
2. https://pipenv.pypa.io/en/latest/basics/#specifying-versions-of-python
3. https://pipenv.pypa.io/en/latest/diagnose/#pipenv-does-not-respect-pyenvs-global-and-local-python-versions

I think many people could install different python versions using pyenv. And actually the python path was set to be
`/home/User/.pyenv/shims/python`, instead of the system path `/usr/bin/python`.

If the pyenv is installed, and python version is not specified, `pipenv install` could get the default python version from `/shims/python`, which cause `⠙ Creating virtual environment...FileNotFoundError: [Errno 2] No such file or directory: '/shims/python`'

PR #783, opened 2021-03-15
Fixed extra quotes for lower alpha characters

### LINKED ISSUE #769: Parentheses not properly escaped under tcsh
**Describe the bug**
Parentheses are not being escaped for tcsh.  

**To Reproduce**
The following will fail for tcsh:

```
export FOO="aaa(bbb)ccc"
```

error:

```
direnv: export +FOO
Badly placed ()'s.
```

**Expected behavior**
Don't remove the quote or escape the ( ) characters.

**Environment**
 - OS: macos 10.15.7  (homebrew)
 - Shell: tcsh
 - Direnv version: 2.27.0

**Additional context**
I don't get this error under bash.  

Thanks

ytrinh:
PR 788 (https://github.com/direnv/direnv/pull/778) to fix  #769 introduced extra quotes around lower case characters.  This PR should fix that issue.  Also, fixed a const misnomer 'LOWERCASE_Z', ascii 90 is uppercase `Z`.

### DIFF
--- a/shell_bash.go
+++ b/shell_bash.go
@@ -71,12 +71,13 @@ const (
 	PLUS          = 43
 	NINE          = 57
 	QUESTION      = 63
-	LOWERCASE_Z   = 90
+	UPPERCASE_Z   = 90
 	OPEN_BRACKET  = 91
 	BACKSLASH     = 92
 	UNDERSCORE    = 95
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks, good catch!

PR #800, opened 2021-05-20
Add -r flag for matching Git branches with a regexp

brianrodri:
I've found it to be very helpful to be able to match a regexp rather than an exact GitHub branch. In my use case, I do this to choose between a Python 2 layout vs Python 3, for example (included in documentation):

```bash
if on_git_branch -r '.*py2'; then
  layout python2
else
  layout python
fi
```

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1212,7 +1212,7 @@ on_git_branch() {
   elif [ -z "$1" ]; then
     return 0
   elif [ "$1" = "-r" ] && [ -z "$2" ]; then
-    log_error "missing regexp pattern after `-r` flag"
+    log_error "missing regexp pattern after \`-r\` flag"
     return 1
   fi
   watch_file "$git_dir/HEAD"
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1218,7 +1218,11 @@ on_git_branch() {
   watch_file "$git_dir/HEAD"
   local git_branch
   git_branch=$(git branch --show-current)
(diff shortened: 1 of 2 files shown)


stdlib.sh
```diff
@@ -1202,9 +1211,14 @@ on_git_branch() {
     return 1
   elif [ -z "$1" ]; then
     return 0
+  elif [ "$1" = "-r" ] && [ -z "$2" ]; then
```
zimbatm, line 1214:
```suggestion
  elif [[ "$1" = "-r" && -z "$2" ]]; then
```
Let's use bash tests everywhere.

stdlib.sh
```diff
@@ -1202,9 +1211,14 @@ on_git_branch() {
...
+    log_error "missing regexp pattern after \`-r\` flag"
+    return 1
   fi
   watch_file "$git_dir/HEAD"
-  [ "$(git branch --show-current)" = "$1" ]
+  local git_branch
+  git_branch=$(git branch --show-current)
+  [ "$1" = "$git_branch" ] || [ "$1" = '-r' ] && [[ "$git_branch" =~ $2 ]]
```
zimbatm, line 1221: The logic would work unless `"$1" = "-r"`. This is unlikely but it's better to use `if [[ "$1 = '-r' ]]; then [[ $git_branch" =~ $2 ]] else [[ "$1" = "$git_branch" ]]; fi`

zimbatm (commented):
sounds good!

stdlib.sh
brianrodri: Done!

zimbatm:
thanks!

PR #808, opened 2021-06-25
Updated conditional for zsh hook to be more forgiving

### LINKED ISSUE #803: Zsh direnv hook errors on "paramter not set"
**Describe the bug**
The zsh hook fails

**To Reproduce**

```sh
$ eval "$(direnv hook zsh)" 
zsh: precmd_functions[(r)_direnv_hook]: parameter not set
$ echo $?
1
```

**Expected behavior**
The hook should succeed

**Environment**
 - OS: Fedora 33
 - Shell: zsh
 - Direnv version: 2.27.0

**Additional context**
I tried the bash hook on my environment and it works

Risto-Stevcev:
closes #803 

The issue is when something like `set -e` or `set -euo pipefail` is set, it causes the conditional check to fail and then the entire script for the hook exits.

This updates the conditional so that it doesn't error out in this scenario. It's adapted from the POSIX-compliant test here:
https://stackoverflow.com/questions/42655304/how-do-i-check-if-a-variable-is-set-in-zsh

### DIFF
--- a/shell_zsh.go
+++ b/shell_zsh.go
@@ -13,11 +13,11 @@ _direnv_hook() {
   trap - SIGINT;
 }
 typeset -ag precmd_functions;
-if [[ -z ${precmd_functions[(r)_direnv_hook]} ]]; then
+if [[ -z "${precmd_functions[(r)_direnv_hook]+1}" ]]; then
   precmd_functions=( _direnv_hook ${precmd_functions[@]} )
 fi
 typeset -ag chpwd_functions;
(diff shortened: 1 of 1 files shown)


zimbatm:
Thanks. Do you mind using the `if (( ${+SOME_VARIABLE} )); then` notation since it's zsh?

Risto-Stevcev:
That syntax also fails in the same way, the entire script for the hook exits on failure. The POSIX-compliant one was the only one that worked for me.

zimbatm:
Oh well, let me stop being nitpicky. Tested locally and it works as expected.

PR #845, opened 2021-11-26
Add automatic `.env` load

felipecrs:
This PR is a continuation of #438, made by @aude.


felipecrs:
This should be ready to review.

Kaligule:
I am not a maintainer but I read throught the changes (especially the tests) and it looks very reasonable to me.

zimbatm:
Thanks for your PR. I should have left a note before; before I can merge this, I want to give it a try and see how it feels.

felipecrs:
Whoa, what a good day to live!

hacfi:
As much as I like this addition I would like to disable this by default if there is a way to do. Is there?

felipecrs:
You can disable it by creating an .envrc file right beside your .env file. Even if it's empty.

awarrenlove:
Can this be disabled without requiring a .envrc file? I'm now getting prompts to allow loading any .env files on my machine, even in places that are completely unrelated to anything I want running under direnv.

I tried creating an empty .envrc file in my user directory just to see if that would override it in all subdirectories, but it still wants to load any .env files in the current directory, which is understandable given the file search behavior, but it is in this case unfortunate.

mfriedenhagen:
@felipecrs: but even if you create an empty `.envrc` and allow it, the variables defined in a parent directory are removed.

I use `.envrc` to import specific password-manager commands to expose a token on demand for accessing a gitlab instance with [lab](https://zaquestion.github.io/lab/) and parsing `.env` completely breaks this :-(

felipecrs:
@awarrenlove

> Can this be disabled without requiring a .envrc file?

Not with the current implementation. But I also agree that it would be nice to be configurable.

> I tried creating an empty .envrc file in my user directory just to see if that would override it in all subdirectories, but it still wants to load any .env files in the current directory, which is understandable given the file search behavior, but it is in this case unfortunate.

Indeed, this is the current behavior. That's why I said: _You can disable it by creating an .envrc file **right beside** your .env file_.


@mfriedenhagen
> but even if you create an empty .envrc and allow it, the variables defined in a parent directory are removed.

Variables removed? That's a weird, I don't think I understood, but you may try to create an `.envrc` (empty or not) in every folder that has a `.env` which you don't want it to be loaded.

awarrenlove:
> Indeed, this is the current behavior. That's why I said: _You can disable it by creating an .envrc file **right beside** your .env file_.

Yes, I understand that. However, that now requires me to litter my machine with empty .envrc files anywhere I have a .env file. And in fact, I have more .env files on my machine that I don't want direnv to load than those that I do (currently using an explicit dotenv_if_exists).

felipecrs:
Right, so let's focus on making it configurable then. I would suggest you to first open a new issue so it can be discussed with the maintainers.

mfriedenhagen:
Hello @felipecrs,

* I have a file `.envrc` like:
```
❯ cat ~/ghq/git.example.com/.envrc
LC_HOST=git.example.com
export LAB_CORE_HOST=https://$LC_HOST
export LAB_CORE_USER=theuser
export LAB_CORE_TOKEN_GENERATOR="security find-internet-password -s $LC_HOST -a $LAB_CORE_USER -g -w"
export GIT_AUTHOR_NAME="Name"
export GIT_AUTHOR_EMAIL=email@example.com
export GIT_COMMITTER_EMAIL=email@example.com
```

resp:

```
❯ cat ~/ghq/github.com/.envrc
export GITHUB_TOKEN_GENERATOR="security find-internet-password -s github.com -a mfriedenhagen -g -w"
export GIT_AUTHOR_NAME="Mirko Friedenhagen"
export GIT_AUTHOR_EMAIL=mfriedenhagen-at-...
export GIT_COMMITTER_EMAIL=mfriedenhagen-at-...
```

* All projects from are checked out beneath these directory (using [ghq](https://github.com/x-motemen/ghq) BTW).
* I may now use aforementioned `lab` tool or [gh](https://cli.github.com) using aliases like: 
```lab='env LAB_CORE_TOKEN=`eval $LAB_CORE_TOKEN_GENERATOR` VISUAL=vi /usr/local/bin/lab'```


And I now have quite some projects where I created files named `.env` for other reasons (secrets for local docker-compose setups, e.g.) but would like to still use my settings for `lab` resp. `gh`.

felipecrs:
@mfriedenhagen if I understood well, all you have to do is to create an empty .envrc right beside your .env files, as I suggested first.

mfriedenhagen:
Hi @felipecrs, that I tried firstly. I touched .envrc and allowed the empty file. Changing out of the project directory immediately exported e.g. LAB_CORE_HOST. But when I cd'd to the project directory with now both .env and .envrc again, the variable disappears from the environment as before.

felipecrs:
Wow, okay... that was not supposed to be how direnv works. Unless I'm wrong, direnv keeps collecting the environment for all the .envrc files up to your home, so the blank .envrc file should not make it stop doing that.

But I'm not as experienced in direnv as some other guys here, so... can someone confirm if what I said correct?

/cc @zimbatm

lucaspiller:
I've found this feature has completely broken direnv for the way I use it. A few projects I work on have different components (in sub-folders, e.g. frontend and backend) that each need their own component-specific environmental variables.

I use direnv to populate project-wide settings that I need in the shell (e.g. AWS credentials) and also setup the development environment (e.g. initialising nvm).

Then I use .env files with dotenv (or something similar) to automatically load them in the development environment when the app is booted - but I don't care about accessing the contents of the .env files in the shell.

felipecrs:
Ok guys, I created https://github.com/direnv/direnv/issues/880. Would that address your concerns?

Issue #768, opened 2021-02-22
direnv allow and direnv exec treat paths with symlinks differently causing frustration


krupan:
**Describe the bug**
`direnv allow` resolves a path with symlinks and records a canonical path with no symlinks.  `direnv exec` does not.  That means that if you `allow` and `exec` the same path, `exec` might still report that the .envrc at that path is not allowed.

**To Reproduce**
```
mkdir -p /tmp/long/path
ln -s /tmp/long/path /tmp/path
echo "export FOO=foo" > /tmp/path/.envrc
direnv allow /tmp/path
direnv exec /tmp/path sh -c 'echo $FOO'
```
You get an error:
```
direnv: error /tmp/path/.envrc is blocked. Run `direnv allow` to approve its content
```
**Expected behavior**
no errors, "foo" is printed to the shell

**Environment**
 - OS: centos 7
 - Shell: bash
 - Direnv version 2.27.0

**Additional context**
I know, I know, stupid symlinks.

The workaround for this, by the way, is to do this:
```
mkdir -p /tmp/long/path
ln -s /tmp/long/path /tmp/path
echo "export FOO=foo" > /tmp/path/.envrc
direnv allow /tmp/path
direnv exec $(readlink -f /tmp/path) sh -c 'echo $FOO'
```

PR #851, opened 2021-12-01
Resolve symlinks during `direnv deny`

### LINKED ISSUE #850: direnv allow resolves symlinks; direnv deny does not
**Describe the bug**
i'd assumed that `direnv allow some/path` followed by `direnv deny some/path` blocks that .envrc again, but it does not if part of the path is a symlink.

**To Reproduce**
```
mkdir foo
ln -s foo bar
touch foo/.envrc
direnv allow bar
direnv deny bar
```

the deny call results in an error:
> direnv: error remove /home/mkhl/.local/share/direnv/allow/...: no such file or directory

**Expected behavior**
i expect `direnv deny bar` to block the file that `direnv allow bar` allowed.
instead `direnv deny foo` is required.

**Environment**
 - OS: Fedora Linux 35
 - Shell: fish
 - Direnv version 2.28.0

**Additional context**
this might be the same issue as #768, but i think the fact that `direnv allow` and `direnv deny` should work identically should give it more priority.

this problem popped up while working on the vscode extension. because of it, the extension currently cannot block the .envrc of a workspace with a symlink in its path.

mkhl:
- add test for denying envrc in symlinked path
- deny: fix for paths with symlinks

fixes #850


mkhl:
this does *not* fix #768 since resolving symlinks is done per command. it might make sense to extract a function to rc that preprocesses paths to envrc files

PR #868, opened 2021-12-28
Fix version check command (fixes #867)

### LINKED ISSUE #867: `direnv version <ver>` always returns older
When running `direnv version <ver>`, it always thinks it's an older version.

```bash
$ direnv version 2.10
direnv: error current version v2.30.1
 is older than the desired version v2.10
```

**To Reproduce**
Run `direnv version <v>` with any older version, newer version, or same version

**Expected behavior**
Should exit 0 and no message for older versions

**Environment**
 - OS: macOS Monterey
 - Shell: zsh
 - Direnv version 2.30.1 (installed from homebrew)

**Additional context**
This happens with a fresh checkout as well.
I think the telltale is the newline after the old version. The version is read from a file, and I think the newline from the file is there too.

zohar-cyc:


eval:
@zohar-cyc I ran into a related issue (ie `direnv version` printing an extra newline) and submitted PR #869 that fixes both issues.

zimbatm:
thanks, #869 now contains a similar fix

PR #869, opened 2021-12-28
Strip newline from version.txt

### LINKED ISSUE #867: `direnv version <ver>` always returns older

eval:
This fixes the version command showing an extra newline:
```
$ direnv version
2.30.1

$ 
```
And fixes #867.

### DIFF
--- a/main.go
+++ b/main.go
@@ -4,6 +4,7 @@ import (
 	_ "embed"
 	"github.com/direnv/direnv/v2/internal/cmd"
 	"os"
+	"strings"
 )
 
 var (
@@ -20,7 +21,7 @@ func main() {
 		env  = cmd.GetEnv()
 		args = os.Args
 	)
-	err := cmd.Main(env, args, bashPath, stdlib, version)
+	err := cmd.Main(env, args, bashPath, stdlib, strings.TrimSpace(version))
 	if err != nil {
 		os.Exit(1)
 	}


zohar-cyc:
I'm worried that's fragile, someone might re-add the newline and not notice, and it will break again.

zimbatm:
thanks for the PR, I updated it to trim the spaces at runtime to avoid this issue in the future as well.

eval:
@zimbatm I just pushed [this commit](https://github.com/eval/direnv/commit/57ae9b7852728ad41319ae19453cdb91be0ca224.patch) after the merge. This would check for the root cause, ie if version.txt contains a semantic version. The semver-library compare will consider it older than any other semantic version.  
Let me know if you need a new PR for it.

zimbatm:
good catch, yes please :)

PR #870, opened 2021-12-28
source_env: show full path

eval:
Paths passed to `source_env(_if_exists)` are logged as-is, e.g. `.envrc.local`.
This PR ensures the full path is logged, like with `source_up`.
```bash
$ cat foo/.envrc
source_up
source_env .envrc.local

# CURRENT
$ direnv exec foo true
direnv: loading ~/src/direnv/foo/.envrc
direnv: loading ~/src/direnv/.envrc
direnv: loading .envrc.local

# PR
$ direnv exec foo true
direnv: loading ~/src/direnv/foo/.envrc
direnv: loading ~/src/direnv/.envrc
direnv: loading ~/src/direnv/foo/.envrc.local
```

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -346,7 +346,7 @@ source_env() {
   pushd "$(pwd 2>/dev/null)" >/dev/null || return 1
   pushd "$rcpath_dir" >/dev/null || return 1
   if [[ -f ./$rcpath_base ]]; then
-    log_status "loading $rcfile"
+    log_status "loading $(user_rel_path "$(expand_path "$rcpath")")"
     # shellcheck disable=SC1090
     . "./$rcpath_base"
   else
--- a/test/stdlib.bash
+++ b/test/stdlib.bash
@@ -190,6 +190,11 @@ test_name source_env_if_exists
   echo "export FOO=bar" > existing_file
   source_env_if_exists existing_file
   [[ $FOO = bar ]]
+
+  # Expect correct path being logged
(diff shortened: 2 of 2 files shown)


CrystalMethod:
We have already stumbled across this several times.  If you have an ".envrc.local" in both directories, you get the same logging twice, which often leads to the assumption that the same file is sourced twice.

```
direnv: loading ~/src/direnv/foo/.envrc
direnv: loading ~/src/direnv/.envrc
direnv: loading .envrc.local
direnv: loading .envrc.local
```

stdlib.sh
```diff
@@ -346,7 +346,7 @@ source_env() {
   pushd "$(pwd 2>/dev/null)" >/dev/null || return 1
   pushd "$rcpath_dir" >/dev/null || return 1
   if [[ -f ./$rcpath_base ]]; then
-    log_status "loading $rcfile"
+    log_status "loading $(expand_path "$rcpath")"
```
yermulnik, line 349: Should the `expand_path` be amended to show `~` instead of user home path?
Kind of:
```suggestion
    log_status "loading $(expand_path "$rcpath" | sed 's|^'${HOME:=none}'|~|')"
```

eval: @yermulnik thanks for catching the regression! `user_rel_path` did it eventually.

eval:
@zimbatm thoughts on this one? 🙏🏻

zimbatm:
thanks!

PR #871, opened 2021-12-28
Test whether version.txt contains semantic version

eval:
semver.Compare considers all non-semantic versions as older; the root
cause of #867.


zimbatm:
thanks!

PR #872, opened 2021-12-29
stdlib: add `env_vars_required`

### LINKED ISSUE #843: Check presence of env-vars

I often use the following setup when working with teams:
- `.envrc.local` git-ignored (originating from `.envrc.local.example`) and `.envrc` in git containing:
```
# .envrc
export DATABASE_URL=postgres://localhost:5432/project_development

source_env_if_exists .envrc.local
# Beyond this point we need SOME_TOKEN to be present
```
So DATABASE_URL is a variable that is straightforward: we can provide a sane default value and if needed it can be locally overridden.

There are also variables that our project requires but 1) for which we cannot provide plausible defaults (vary too much per developer) or 2) that are secrets which we won't include in `.envrc`.

For these types of variables it would be nice to have the possibility to check their presence and warn if absent:
```
# at end of .envrc
env_vars_required GITHUB_NAME, SOME_TOKEN
```

eval:
Check presence of env vars:

    env_vars_required SOME_TOKEN

Fixes #843

### DIFF
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -126,6 +126,27 @@ source_env_if_exists .envrc.private
 .fi
 .RE
 
+.SS \fB\fCenv_vars_required <varname> [<varname> ...]\fR
+.PP
+Logs error for every variable not present in the environment or having an empty value.
+.br
+Typically this is used in combination with source_env and source_env_if_exists.
+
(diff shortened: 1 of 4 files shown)


zimbatm (approved):
sweet, thanks for the PR

PR #878, opened 2022-01-04
Allow skipping `.env` autoload

awarrenlove:
The feature to automatically load .env files in https://github.com/direnv/direnv/pull/845 is nice, but it's a little too greedy in certain situations. Allow skipping that behavior by default (going back to explicit `dotenv` calls in `.envrc` files) in the configuration toml file.

### DIFF
--- a/internal/cmd/cmd_exec.go
+++ b/internal/cmd/cmd_exec.go
@@ -52,7 +52,7 @@ func cmdExecAction(env Env, args []string, config *Config) (err error) {
 	previousEnv.CleanContext()
 
 	// Load the rc
-	if toLoad := findEnvUp(rcPath); toLoad != "" {
+	if toLoad := findEnvUp(rcPath, config.SkipDotenv); toLoad != "" {
 		if newEnv, err = config.EnvFromRC(toLoad, previousEnv); err != nil {
 			return
 		}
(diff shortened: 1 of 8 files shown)


zimbatm:
thanks for the contrib!

zimbatm:
I might switch the option around tomorrow to make it opt-in instead since it breaks back-compat.

awarrenlove:
Personally I would prefer that, and thought about going that route with this PR, but didn't want to switch default behavior.

aneeshusa:
Also would really like to make the `.env` functionality opt-in instead of opt-out!

shartep:
my version is 2.30.3:
```
$ direnv --version
2.30.3
```
but it still loads `.env` files by default, what should I do to change this?

arulrajnet:
@shartep As per the [doc](https://direnv.net/man/direnv.toml.1.html) you have to create direnv.toml file in the config directory

```
mkdir -p ${HOME}/.config/direnv
tee ${HOME}/.config/direnv/direnv.toml <<"EOF"
[global]
skip_dotenv = true
EOF
```

This works for me.

zimbatm:
Sorry for the back and forth, #911 now inverts the config so that feature isn't enable by default anymore.

PR #573, opened 2020-01-30
config: add strict_env option

zimbatm:
When set to true, the `.envrc` will be loaded with the `set -euo
pipefail` flags.

PR #881, opened 2022-01-05
Re-add accidentally deleted comment line

aphedges:
I was looking through [`config.go`](https://github.com/direnv/direnv/blob/f68f741895754e9551059c8e0132a796ad1c3c3a/internal/cmd/config.go), and I found a comment that was incomplete. It turned out that the final line was accidentally deleted in #573.

I can't find any development guide on things like commit message conventions or PR workflow, so I hope I am doing this correctly. Please let me know if I need to change anything.


zimbatm:
thanks!

PR #883, opened 2022-01-12
Fix bundler not installing into local .direnv

maxnordlund:
The most important part is to add `GEM_HOME` to `GEM_PATH`, but to be extra safe also make sure that the directories exists.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -957,8 +957,13 @@ layout_ruby() {
   export BUNDLE_BIN
   export GEM_HOME
 
+  mkdir -p "$GEM_HOME/bin"
+  mkdir -p "$BUNDLE_BIN"
+
   PATH_add "$GEM_HOME/bin"
   PATH_add "$BUNDLE_BIN"
+  
+  path_add GEM_PATH "$GEM_HOME"
 }
 
 # Usage: layout julia


zimbatm:
Thanks, is there any test that could be run to verify this behaviour?

maxnordlund:
Hm, I don't know. I verified it locally, but that's a weak argument. Maybe setup a small project/directory with a Gemfile and .tool-versions and run `bundle install` and see if there's stuff in `.direnv/`?

<details>
<summary>Example Gemfile</summary>

```ruby
# frozen_string_literal: true

source "https://rubygems.org"

git_source(:github) { |repo_name| "https://github.com/#{repo_name}" }

# gem "rails"

gem "solargraph", group: :development

```
</details>

stdlib.sh
```diff
@@ -957,8 +957,13 @@ layout_ruby() {
   export BUNDLE_BIN
   export GEM_HOME
 
+  mkdir -p "$GEM_HOME/bin"
+  mkdir -p "$BUNDLE_BIN"
+
   PATH_add "$GEM_HOME/bin"
   PATH_add "$BUNDLE_BIN"
+  
+  path_add GEM_PATH "$GEM_HOME"
```
zimbatm, line 966: How come bundler doesn't install into GEM_HOME? I haven't touched bundler in a while so maybe that behaviour has changed.

PR #884, opened 2022-01-14
Ensure status log messages are printed with normal color

tmatilai:
Reset color for normal status messages.

This might helps in a case when an evaluated command has not cleaned up after itself. But also makes it possible to add support for coloring all stderr output (PR coming).

While at it, also set variables in `log_error()` only if needed


zimbatm:
Thanks, I just rebased the PR to use escape codes instead so it doesn't depend on curses to function.

PR #889, opened 2022-01-22
Allow ignoring specified environment variables

### LINKED ISSUE #401: Feature: ability to specify a list of ignored environment variables
Hello there, first of all thanks for building direnv!

Recently I was bitten by the fact that direnv controls the environment variables that TMUX uses for bookkeeping. Luckily I was able to get the source and see that we are already ignoring some special environment variables in https://github.com/direnv/direnv/blob/master/env_diff.go so I just had to add
```
	if strings.HasPrefix(key, "TMUX") {
		return true
	}
```
run make and have my setup working again.

Do you think it would be a good idea to be able to pass a list of ignored variables/patterns? Let me know and I can work on building a PR for this.

tmatilai:
Add `[env] / ignore_keys` configuration to ignore the listed environment variables, in addition to the default ones.

Fixes #401


internal/cmd/rc.go
```diff
@@ -139,6 +139,7 @@ func (rc *RC) Load(previousEnv Env) (newEnv Env, err error) {
 	direnv := config.SelfPath
 	newEnv = previousEnv.Copy()
 	newEnv[DIRENV_WATCHES] = rc.times.Marshal()
+	newEnv[DIRENV_IGNORE_KEYS] = strings.Join(config.EnvIgnoreKeys, ",")
```
tmatilai, line 142: Passing through environment variable avoids the need to read configuration file in at least `direnv apply_dump`. 

As a (wanted) side effect, it also adds appending to the list in `.envrc` for example in some `direnv_load ... direnv dump` scenarios. Although I think we can leave that undocumented if this still changes.

tmatilai (commented):
I'm not much of a Go programmer, so please guide me if horrible things were made. 🙂

internal/cmd/config.go
```diff
@@ -118,6 +126,12 @@ func LoadConfig(env Env) (config *Config, err error) {
 			return
 		}
 
+		if env[DIRENV_IGNORE_KEYS] != "" {
+			config.EnvIgnoreKeys = strings.Split(env[DIRENV_IGNORE_KEYS], ",")
+		} else {
```
zimbatm, line 131: The config option is enough. Historically direnv was configured with env vars but that's more of an artefact now.

internal/cmd/env_diff.go
```diff
@@ -43,13 +49,15 @@ func NewEnvDiff() *EnvDiff {
 func BuildEnvDiff(e1, e2 Env) *EnvDiff {
 	diff := NewEnvDiff()
 
+	ignoredKeys := NewIgnoredKeys(e2)
```
zimbatm, line 52: Let's pass the IngoredKeys instance as a function argument so it can be passed from the config object instead.

internal/cmd/env_diff.go
```diff
@@ -131,16 +139,29 @@ func (diff *EnvDiff) Serialize() string {
 	return gzenv.Marshal(diff)
 }
 
-//// Utils
+//// IgnoredKeys
+
+// NewIgnoredKeys creates a NewIgnoredKeys including all keys to ignore
+func NewIgnoredKeys(env Env) *IgnoredKeys {
```
zimbatm, line 145: This should be derived from the config object instead.

zimbatm (commented):
Sounds good overall, but I think we should only use the config unless there is a compelling argument.

tmatilai:
Thanks for the review!

> I think we should only use the config unless there is a compelling argument.

My use case was actually to dynamically ignore some environment variables (with known prefix) during the `.envrc` evaluation, so the `DIRENV_IGNORE_KEYS` was handy.

But after all, my use case for the whole feature faded away, so I'm totally fine with the config-only approach. But at the same time, let's see when I have the time and inspiration to make the requested changes.

If someone wants to hijack the PR in the meanwhile, please feel free to. 🙂

zimbatm:
Sounds good. I'm happy to keep the set of features small. So let's wait until somebody else is requesting for it :)

NeckBeardPrince:
Cool. Never got merged.

PR #892, opened 2022-01-29
doc: fix the broken link to arch linux

gsilvan:
The link to the arch-aur-package is dead, as this package is now available in the official repo.


zimbatm:
thanks!

PR #894, opened 2022-02-02
Switch default back to not load dotenv files

aneeshusa:
Follow-up on https://github.com/direnv/direnv/pull/878#issuecomment-1005145150.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -31,6 +31,8 @@ type Config struct {
 	WhitelistExact  map[string]bool
 }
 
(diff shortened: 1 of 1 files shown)


aneeshusa:
Happy to adjust the code as desired (e.g. new option name or anything else), just wanted to put this up for consideration as a permanent change. If the plan is to keep/make `config.SkipDotenv = false` as the default in e.g. the next major release, then I don't think this PR is useful to temporarily add back-compat.

zimbatm:
Thanks, I fixed it in #911

PR #901, opened 2022-02-25
Use setenv in vim to allow non alphanumeric vars

jared-w:
ref: `:h setenv()` https://vimhelp.org/builtin.txt.html#setenv%28%29

Of note, the `:h expr-env` documentation says

> The functions getenv() and setenv() can also be used and work for
> environment variables with non-alphanumeric names.
> The function environ() can be used to get a Dict with all environment
> variables.

### DIFF
--- a/internal/cmd/shell_vim.go
+++ b/internal/cmd/shell_vim.go
@@ -33,16 +33,16 @@ func (sh vim) Dump(env Env) (out string) {
 }
 
 func (sh vim) export(key, value string) string {
-	return "let $" + sh.escapeKey(key) + " = " + sh.escapeValue(value) + "\n"
+	return "call setenv(" + sh.escapeKey(key) + "," + sh.escapeValue(value) + ")\n"
 }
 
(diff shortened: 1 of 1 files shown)


zimbatm:
Nice. Is this backwards-compatible with older versions of vim?

jared-w:
It took a bit of digging to find the exact commit, but it appears this is dated back to patch `8.1.1305` (https://github.com/vim/vim/commit/691ddeefb545d8488e5a495af61caba2e57b3de9).

Those appear to [correspond](https://www.arp242.net/vimlog/) to the vims shipped by default in Ubuntu 20.04 LTS (and later), and Debian 11 (having just missed the Debian 10 cutoff). If that's ancient enough, this'll work, if it's not we might need to revisit how we're doing this. Perhaps by exposing a different command that uses `setenv` and calling the appropriate one in `direnv.vim`?

zimbatm:
yeah, that's good enough, thanks

PR #909, opened 2022-03-18
feat: allow conda environment names to be detected from environment.yml

shane-axiom:
Currently an error is encountered if an environment name is not
specified to `layout anaconda`. This change attempts to detect
the environment name from the environment.yml file by grepping
for the `name` property and processing the result.

The result is that .envrc files can now simply contain
`layout anaconda` as long as the local environment.yml
file specifies a `name`.

No new dependencies are introduced as `grep` is already used
and additional processing is performed using Bash parameter
expansion.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -856,7 +856,19 @@ layout_anaconda() {
     env_loc="$env_name_or_prefix"
   else
     # "foo" name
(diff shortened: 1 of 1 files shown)


shane-axiom:
plane

PR #911, opened 2022-03-26
Don't load .env files by default

zimbatm:
Loading the `.env` files is a change in behaviour and broke
backward-compatibility. It sounds really nice at first but it turns out
that many users have .env files littered in their repository and cause a
direnv reload. It's possible to turn that feature off but now they have
one extra step to instruct their users.

This commit inverts the behaviour and turns that feature off by default
to restore back-compat.

See also #878.


henrik242:
Great!

vschoener:
I think the document misses the part where is the toml config, how to create it, and what to configure to use dotenv files back again. (Or I didn't find the info :p)

Issue #913, opened 2022-03-27
Provide a strict_env compatible version of source_up


jfly:
(This request is very similar to https://github.com/direnv/direnv/issues/728, but with `source_up`)

Right now, `source_up` searches upwards (using `find_up`) until it finds a file. If it doesn't, it exits 1, and if you've got `strict_env` turned on, loading the `.envrc` file immediately stops.

Is this intentional? The [docstring on `source_up`](https://github.com/direnv/direnv/blob/abdd443bd5477557509e969e17f74244fec2c55b/stdlib.sh#L419-L423) sort of makes it sound like maybe it was intended to happily ignore missing files?

> Loads another ".envrc" if found with the find_up command.

If it *is* intentional that `source_up` exits nonzero when there's no ancestor file, how would you feel about a PR adding a new `source_up_if_exists` method?

zimbatm:
The exit status is intentional to be able to distinguish if a file was loaded or not.

I don't mind having `source_up_if_exists` being added. A simple workaround is to use `source_up || true`.

Zebradil:
I spend some time trying to figure out why my `.envrc` doesn't export variables. For me it could've been helpful to see an error message about failed `source_up`.

jfly:
@Zebradil check out https://github.com/direnv/direnv/pull/921, I think it'll make that situation a little better.

Zebradil:
@jfly yes, that looks great, thank you!

PR #921, opened 2022-04-08
source_up_if_exists: A strict_env compatible version of source_up

jfly:
This fixes https://github.com/direnv/direnv/issues/913

While I was in here, I opted to make the existing `source_up` script log
explicitly when it can't find a file. I think that's useful both with
strict_env enabled and disabled.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -416,6 +416,26 @@ watch_dir() {
   eval "$("$direnv" watch-dir bash "$1")"
 }
 
+# Usage: _source_up [<filename>] [true|false]
+#
+# Helper function for source_up and source_up_if_exists. The second parameter
+# determines if it's an error for the file we're searching for to not exist.
+_source_up() {
+  local envrc file=${1:-.envrc}
+  local ok_if_not_exist=${2}
+  envrc=$(cd .. && (find_up "$file" || true))
+  if [[ -n $envrc ]]; then
+    source_env "$envrc"
+  else
+    if $ok_if_not_exist; then
+      return 0
+    else
+      log_error "No ancestor $file found"
+      return 1
+    fi
+  fi
+}
+
 # Usage: source_up [<filename>]
 #
 # Loads another ".envrc" if found with the find_up command. Returns 1 if no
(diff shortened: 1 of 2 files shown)


stdlib.sh
```diff
@@ -418,14 +418,32 @@ watch_dir() {
...
-  dir=$(cd .. && find_up "$file")
-  if [[ -n $dir ]]; then
-    source_env "$dir"
+  local envrc file=${1:-.envrc}
+  envrc=$(cd .. && (find_up "$file" || true))
+  if [[ -n $envrc ]]; then
+    source_env "$envrc"
+  else
+    log_error "No ancestor $file found"
+    exit 1
```
zimbatm, line 432:
```suggestion
    return 1
```
This lets the user then decide above what to do

stdlib.sh
```diff
@@ -418,14 +418,32 @@ watch_dir() {
...
+# Loads another ".envrc" if found with the find_up command. If one is not
+# found, nothing happens.
+#
+# NOTE: the other ".envrc" is not checked by the security framework.
+source_up_if_exists() {
+  local envrc file=${1:-.envrc}
+  envrc=$(cd .. && (find_up "$file" || true))
+  if [[ -n $envrc ]]; then
+    source_env "$envrc"
   fi
```
zimbatm, line 447:
```suggestion
  source_up "${1:-}" || true
```
With the above change it becomes possible to deduplicate the logic

zimbatm (changes requested):
Thanks for the PR!

There are two minor suggestions. If you could also extend the stdlib docs in the man folder that would be perfect.

stdlib.sh
jfly: Oops, fixed!

stdlib.sh
jfly: That works, but with my change to `source_up`, it means you get a red "No ancestor .envrc found" message even though this function still succeeds.

I've tweaked things around a little bit to try to DRY this up. I'm honestly not sure if it's better or worse. What do you think, @zimbatm?

zimbatm (approved):
thanks, I pushed a small fixup and it's good to go

jfly:
Thanks, @zimbatm!

ssbarnea:
Thanks for implementing this feature, I was looking for a way to avoid a confusing warning.

jfly:
Oh, that's on me, not @zimbatm, sorry! I agree that this feels like an unnecessary footgun: could we get the generated files out of version control?

PR #931, opened 2022-04-15
Expand `~/` in whitelist paths

### LINKED ISSUE #926: Allow flexible whitelist by allowing `~/` or `$HOME` expansion
Currently I would need to have this in my `$XDG_CONFIG_HOME/direnv/direnv.toml`:

```toml
[whitelist]
prefix = [ "/Users/erikw/src/something/" ]
```

however as dotfiles are share across different system which have different directory hierchy e.g. Linux having `/home` instead of `/Users` like macOS, I would need to adapt the `direnv.toml` per system type.

Solutions could be
```toml
[whitelist]
prefix = [ "~/src/something/" ]
```

or

```toml
[whitelist]
prefix = [ "$HOME/src/something/" ]
```

**Additional context**
Separate issue -- we should really move away from the dated whitelist terminology and adopt **allowlist** instead!

erikw:
Expand a path string prefixed with ~/ to the current user's home directory.

Example: if current user is `user1` with home directory in `/home/user1`,
then `~/project` -> `/home/user1/project`

It's useful to allow paths with `~/`,  so that `direnv.toml` can be reused via
dotfiles repos across systems with different standard home paths
(compare Linux `/home` and macOS `/Users`).

Feedback is welcome e.g. if/how this should be tested. For now manual
testing of prefix & exact has been done confirming that full paths as
well as `~/`-paths works.

I looked here for how to expand tilde:
 https://gist.github.com/miguelmota/9ab72c5e342f833123c0b5cfd5aca468

I also found a third party module that expands tilde
https://github.com/prep/tilde
that we can swap in instead of the `expandTildePath` method in this PR
if desired. I assumed the project would like to minimize dependencies
and made this simple function instead.

Feel free to amend or push more commits to the PR if wanted!

Fixes #926

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -61,6 +61,22 @@ type tomlWhitelist struct {
 	Exact  []string
 }
 
+// Expand a path string prefixed with ~/ to the current user's home directory.
+// Example: if current user is user1 with home directory in /home/user1, then
+// ~/project -> /home/user1/project
+// It's useful to allow paths with ~/, so that direnv.toml can be reused via
+// dotfiles repos across systems with different standard home paths
+// (compare Linux /home and macOS /Users).
+func expandTildePath(path string) (pathExpanded string) {
+	pathExpanded = path
+	if strings.HasPrefix(path, "~/") {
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks!

PR #935, opened 2022-04-22
Add convenient aliases for allow/deny commands

### LINKED ISSUE #832: Add "direnv block" as an alias of "direnv deny"
### Is your feature request related to a problem?

I seem to always accidentally type `direnv block` when I want `direnv deny`.

I believe other users may do the same.

### Describe the solution you'd like

Modify command-line parsing so that the `block` subcommand is an alias of the `deny` subcommand.
Also make `permit` an alias of `allow` while we're at it.
The aliases may even be hidden from the user if you wish.

### Describe alternatives you've considered

Currently I have a function in my `.zshrc` to translate my mistakes into the proper `direnv` command line.

```bash
# Translate "direnv block" to "direnv deny".
# https://unix.stackexchange.com/a/607909/12311
direnv () {
  setopt local_options extended_glob unset
  local -i i=$argv[(i)^-*]
  # $argv[i] is the first non-option argument
  # (or empty if there is none).
  # See if we want to change it.
  case $argv[i] in
    permit) argv[i]="allow";;
    block) argv[i]="deny";;
  esac
  command direnv "$@"
}
````

### Additional context

My confusion probably originates from the use of the word "block" in messages from `direnv`. For example:
```
direnv: error /home/rodney/ops/config/.envrc is blocked. Run `direnv allow` to approve its content
```

Also, language-wise I'm not an expert here, but I _think_ "block" is a closer antonym to "allow" than "deny" is. For example, "blocklist" and "allowlist" (the modern terms for "blacklist" and "whitelist"). "Block" comes from Dutch via Old English; "allow" comes vaguely from Latin, but it's mostly from French via Old English. "Deny" and "permit" are almost straight from Latin.

### Other information

Thanks for `direnv`!

rvl:
It's easy to forget the correct command name -- so support them all!

 - allow  / block
 - permit / deny
 - grant  / revoke

Closes #832

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,7 @@
+UNRELEASED / 2022-04-??
+==================
+  * Add "block" and "revoke" as aliases of the "deny" command (#935)
+  * Add "permit" and "grant" as aliases of the "allow" command (#935)
 
 2.31.0 / 2022-03-26
 ==================
--- a/internal/cmd/cmd_allow.go
+++ b/internal/cmd/cmd_allow.go
@@ -8,10 +8,11 @@ import (
 
 // CmdAllow is `direnv allow [PATH_TO_RC]`
 var CmdAllow = &Cmd{
-	Name:   "allow",
-	Desc:   "Grants direnv to load the given .envrc or .env",
-	Args:   []string{"[PATH_TO_RC]"},
-	Action: actionWithConfig(cmdAllowAction),
+	Name:    "allow",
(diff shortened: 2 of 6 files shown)


zimbatm (commented):
Look good. The CI is failing with a gofmt issue. Once that's fixed the PR should be good to go.

rvl:
Thanks @zimbatm. CI should be fixed now.

zimbatm:
thanks!

PR #946, opened 2022-05-21
go: bump golang.org/x/sys for linux/loong64 support

xen0n:
This is a new architecture being introduced in go1.19 (golang/go#46229); it's a desktop/server architecture, so there are going to be direnv users. (Me included; hence this patch.)

BTW thanks for the great project!


zimbatm (approved):
sounds good!

Once Go 1.19 is out, feel free to add it to the released binaries as well (see install.sh and the `dist` GNUMakefile target) .

PR #947, opened 2022-05-23
Enable ppc64le builds

imphil:
Currently, only big-endian PowerPC builds are provided.
This commit enables little-endian PowerPC builds as well.

Little-endian PowerPC is the default for all "modern" PowerPC
offerings by IBM and arguably the more common PowerPC variant
today.


zimbatm:
sounds good!

PR #949, opened 2022-05-25
fix: remove try block

avioli:
`try {} except e {}` is now deprecated (elvish v0.18.0) and will be
removed in the next version - 0.19.0.

Since direnv will most probably run in pre-0.18.0 elvish
environments it is maybe for the best to remove the try block.

As far as I can see - its use is to catch exceptions from running
`direnv export elvish`, which should only happen if direnv vanishes
from the user's path. Any other error should most probably fail
anyway, rather than echo the exception but proceed.

### DIFF
--- a/internal/cmd/shell_elvish.go
+++ b/internal/cmd/shell_elvish.go
@@ -13,20 +13,16 @@ var Elvish Shell = elvish{}
 func (elvish) Hook() (string, error) {
 	return `## hook for direnv
 set @edit:before-readline = $@edit:before-readline {
-	try {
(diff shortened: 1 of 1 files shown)


avioli:
If I `cd` into a directory with `.envrc` I used to get:

```sh
direnv: error /tmp/.envrc is blocked. Run `direnv allow` to approve its content
[&reason=[&cmd-name=/opt/homebrew/bin/direnv &exit-status=1 &pid=96438 &type=external-cmd/exited]]
```

now I get

```sh
```

zimbatm:
@icidasset @mmlb do you have any opinion on this PR?

PR #962, opened 2022-06-30
Update layout anaconda to accept a path to a yml file

Arrowbox:
Conda assumes the environment config is provided by
environment.yml when the --file argument is not
provided. This allows for specifying a path to
a different file. For example, a project that has
a two environment definitions based on host platform.

Some of the projects I work with don't use environment.yml
in the root directory. I've been using this in my direnvrc to
handle it.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -863,17 +863,26 @@ layout_python3() {
   layout_python python3 "$@"
 }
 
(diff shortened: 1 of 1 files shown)


zimbatm:
The amount of code for that function is getting ridiculous, but fine

PR #969, opened 2022-07-13
Add stdlib pyenv docs

gertvdijk:
It appears to me that `layout pyenv` is not mentioned in the stdlib manpage.

Also sync omissions to `man/direnv-stdlib.1.md` from manual changes to `man/direnv-stdlib.1` (https://github.com/direnv/direnv/pull/921).


zimbatm:
thanks!

PR #973, opened 2022-07-22
feat: add environ format

zimbatm:
Adds a `direnv export environ` command that outputs the diff in
`<key>(=<value)?\0` format. It's a simple format that is quite close to
/proc/<pid>/environ on Linux. Each line is NUL-terminated. If the `=` is
not present on a line, consider that the <key> should be unset.

The idea was to find a format that could eventually replace the evals in
the hooks in the various shells.


zimbatm:
Here is a new hook that replaces `eval "$(direnv hook bash)", for example:
```
_direnv_hook() {
  local previous_exit_status=$? line;
  trap -- '' SIGINT;
  while IFS= read -d '' -r line; do
    if [[ "$line" = *"="* ]]; then
      echo export "$line";
    else
      echo unset "$line";
    fi;
  done < <(direnv export environ);
  trap - SIGINT;
  return $previous_exit_status;
};
if ! [[ "${PROMPT_COMMAND:-}" =~ _direnv_hook ]]; then
  PROMPT_COMMAND="_direnv_hook${PROMPT_COMMAND:+;$PROMPT_COMMAND}"
fi
```

zimbatm:
Same for zsh:
```sh
_direnv_hook() {
  trap -- '' SIGINT;
  local line;
  while IFS= read -d '' -r line; do
    if [[ "$line" = *"="* ]]; then
      echo export "$line";
    else
      echo unset "$line";
    fi;
  done < <(direnv export environ);
  trap - SIGINT;
}
typeset -ag precmd_functions;
if [[ -z "${precmd_functions[(r)_direnv_hook]+1}" ]]; then
  precmd_functions=( _direnv_hook ${precmd_functions[@]} )
fi
typeset -ag chpwd_functions;
if [[ -z "${chpwd_functions[(r)_direnv_hook]+1}" ]]; then
  chpwd_functions=( _direnv_hook ${chpwd_functions[@]} )
fi

```

PR #977, opened 2022-07-28
Add Fig as an autocomplete method

brendanfalk:
[Fig](https://fig.io) provides autocomplete for 300+ CLI tools including direnv. It looks like this:

image

We think this will help users become more efficient with direnv CLI, so we would love to have Fig listed as an autocomplete method in your docs. Thanks so much and please let me know if you have any questions!


zimbatm:
Can you add a new wiki page instead? https://github.com/direnv/direnv/wiki#tools-integration The hooks page is not the right place to advertise your tool.

PR #985, opened 2022-09-09
Docs/xdg config home explainer

murb:
the docs mention '`$XDG_CONFIG_HOME`, but for me (as a mac user) it wasn't clear that this is typically ~/.config ($XDG_CONFIG_HOME is not set)


zimbatm (commented):
Sounds good, thanks

PR #990, opened 2022-09-29
Additionally source libraries from XDG_CONFIG_DIRS

infinisil:
Allows the system administrator to configure direnv libraries and rc files using by default /etc/xdg/direnv/lib/*.sh and /etc/xdg/direnv/direnvrc

See https://github.com/NixOS/nixpkgs/pull/192667#discussion_r979811832 for motivation

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -19,7 +19,16 @@ direnv="$(command -v direnv)"
 DIRENV_LOG_FORMAT="${DIRENV_LOG_FORMAT-direnv: %s}"
 
 # Where direnv configuration should be stored
-direnv_config_dir="${DIRENV_CONFIG:-${XDG_CONFIG_HOME:-$HOME/.config}/direnv}"
+IFS=: xdg_config_dirs=( ${XDG_CONFIG_DIRS:-/etc/xdg} )
+xdg_config_dirs+=( "${XDG_CONFIG_HOME:-$HOME/.config}" )
+
+declare -a direnv_config_dirs
+for dir in "${xdg_config_dirs[@]}"; do
+  direnv_config_dirs+=( "$dir/direnv" )
+done
+if [[ -n "$DIRENV_CONFIG" ]]; then
+  direnv_config_dirs+=( "$DIRENV_CONFIG" )
+fi
 
 # This variable can be used by programs to detect when they are running inside
 # of a .envrc evaluation context. It is ignored by the direnv diffing
@@ -1371,16 +1380,23 @@ __main__() {
   trap __dump_at_exit EXIT
 
   # load direnv libraries
-  for lib in "$direnv_config_dir/lib/"*.sh; do
-    # shellcheck disable=SC1090
-    source "$lib"
+  for direnv_config_dir in "${direnv_config_dirs[@]}"; do
+    for lib in "$direnv_config_dir/lib/"*.sh; do
+      # shellcheck disable=SC1090
+      source "$lib"
+    done
   done
 
-  # load the global ~/.direnvrc if present
-  if [[ -f $direnv_config_dir/direnvrc ]]; then
-    # shellcheck disable=SC1090,SC1091
-    source "$direnv_config_dir/direnvrc" >&2
-  elif [[ -f $HOME/.direnvrc ]]; then
+  # load global direnvrc's if present
+  for direnv_config_dir in "${direnv_config_dirs[@]}"; do
+    if [[ -f $direnv_config_dir/direnvrc ]]; then
+      # shellcheck disable=SC1090,SC1091
+      source "$direnv_config_dir/direnvrc" >&2
+    fi
+  done
+
+  # Legacy .direnvrc
+  if [[ -f $HOME/.direnvrc ]]; then
     # shellcheck disable=SC1090,SC1091
     source "$HOME/.direnvrc" >&2
   fi


stdlib.sh
```diff
@@ -19,7 +19,16 @@ direnv="$(command -v direnv)"
 DIRENV_LOG_FORMAT="${DIRENV_LOG_FORMAT-direnv: %s}"
 
 # Where direnv configuration should be stored
-direnv_config_dir="${DIRENV_CONFIG:-${XDG_CONFIG_HOME:-$HOME/.config}/direnv}"
+IFS=: xdg_config_dirs=( ${XDG_CONFIG_DIRS:-/etc/xdg} )
```
zimbatm, line 22:
```suggestion
direnv_config_dir="${DIRENV_CONFIG:-${XDG_CONFIG_HOME:-$HOME/.config}/direnv}"
IFS=: xdg_data_dirs=( ${XDG_DATA_DIRS:-/usr/local/share/:/usr/share} )
```
I find the XDG spec confusing. I think in this scenario, the `~/.config/direnv/lib/*.sh` can be considered "user" configuration, but the shared libraries are probably part of XDG_DATA_DIRS? Or it should have been XDG_DATA_HOME already.

zimbatm (commented):
thanks, it's moving in the right direction

stdlib.sh
infinisil: The spec says this:

> `$XDG_DATA_HOME` defines the base directory relative to which user-specific data files should be stored

It doesn't explain what "user-specific data files" are, but from my experience I'd say it's "files that contain persistent user-specific data but which are created and managed by the program". I don't think such direnv libraries fit in there, since they're neither user-specific, nor created or managed by the program. And `$XDG_DATA_DIRS` is just the extension of `$XDG_DATA_HOME` to multiple additional directories.

I really think `$XDG_CONFIG_HOME` (and its extension to `$XDG_CONFIG_DIRS`) is the right place for these libraries, they really are configuration: They don't store user-specific data, are reusable between arbitrarily many installation, they're kind of like plugins.

stdlib.sh
```diff
@@ -1371,16 +1380,23 @@ __main__() {
...
+      # shellcheck disable=SC1090
+      source "$lib"
+    done
   done
 
-  # load the global ~/.direnvrc if present
-  if [[ -f $direnv_config_dir/direnvrc ]]; then
-    # shellcheck disable=SC1090,SC1091
-    source "$direnv_config_dir/direnvrc" >&2
-  elif [[ -f $HOME/.direnvrc ]]; then
```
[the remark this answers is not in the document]
infinisil: We don't want to load `~/.direnvrc` multiple times though, so a variable is needed, check out the new commit: https://github.com/direnv/direnv/pull/990/commits/3f6363d01390480d7245d3fec4927885a7c0109a

Is this alright?

stdlib.sh
```diff
@@ -19,7 +19,16 @@ direnv="$(command -v direnv)"
...
+IFS=: xdg_config_dirs=( ${XDG_CONFIG_DIRS:-/etc/xdg} )
+xdg_config_dirs+=( "${XDG_CONFIG_HOME:-$HOME/.config}" )
+
+declare -a direnv_config_dirs
+for dir in "${xdg_config_dirs[@]}"; do
+  direnv_config_dirs+=( "$dir/direnv" )
+done
+if [[ -n "$DIRENV_CONFIG" ]]; then
+  direnv_config_dirs+=( "$DIRENV_CONFIG" )
+fi
```
infinisil, line 31: @zimbatm Btw this also slightly changes behavior: Before when `DIRENV_CONFIG` was set, `$XDG_CONFIG_HOME`/`~/.config` was ignored, but now it's loaded regardless. I can easily change the behavior to work as before, what do you think?

stdlib.sh
zimbatm: I agree for `~/.config/direnv/lib/*.sh`, but as soon as you start scanning outside of the user's home, it becomes re-distributable code. For nix-direnv, you wouldn't expect the user to change the implementation themselves but rather to propose the changes upstream.

This also shows this: https://github.com/NixOS/nixpkgs/pull/192667#discussion_r980135486

stdlib.sh
zimbatm: Given how widely direnv is deployed, it's best to keep as much back-compat as possible. DIRENV_CONFIG is for when you want to override the default behaviour.

stdlib.sh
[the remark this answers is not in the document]
zimbatm: It's not quite there yet. It should only load *the first* direnvrc, starting with `$direnv_config_dir/direnvrc`, then (maybe) the rest of the XDG config dirs, and finally the legacy one. But I think this is best left untouched as it doesn't solve a specific use-case.

stdlib.sh
infinisil: I don't understand, are you saying that because libraries outside `~` are redistributable (though I'd argue libraries in `~` are equally as redestributable), `XDG_DATA_DIRS` is the correct variable to read from? That doesn't make sense to me, `XDG_DATA_DIRS` like `XDG_DATA_HOME` is for data, not configuration or libraries or redistributable things. For GDPR compliance for example, if somebody asks me to wipe all user data, I'd wipe all `XDG_DATA_DIRS` and `XDG_DATA_HOME`, the applications should all continue working with the same configuration though

stdlib.sh
infinisil: I fixed this in https://github.com/direnv/direnv/pull/990/commits/f106329094d9ad67523e64f6837b1b9b89e36d71 now, it has the old behavior again

stdlib.sh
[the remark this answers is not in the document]
infinisil: I'm not sure I understand, can you show me how you'd change the code?

brandung-sjorek:
@infinisil @zimbatm Maybe my PR comes in handy here? See #1038 … I discovered after crearting my PR, so maybe I'll first take a look at your conversation (especially those regarding `nix` specifics).

bbenne10:
Is there any movement on this PR? I am waiting on this PR to approve and merge nixos/nixpkgs#192667, which would be extremely helpful in integrating nix-direnv (and direnv itself) into nix installs natively.

infinisil:
I'm waiting on clarifications of @zimbatm's comments, to me this PR would be good to merge.

PR #996, opened 2022-10-14
fix(test): use lowercase -d flag for base64 decoding of DIRENV_DIFF

Delapouite:
Hello

Coreutils' `base64` program has no `-D` flag (uppercase) to decode data : https://man.archlinux.org/man/base64.1

Thanks


zimbatm:
thanks!

PR #1006, opened 2022-10-27
pass BASH_PATH to make, matches the nixpkgs derivation

zowoq:


zimbatm:
thanks, is it blocking nixpkgs? I'm trying to keep direnv as backwards-compatible as possible so it can be packaged in Debian and friends.

zowoq:
No, was just going to do the FIXME and ended up bumping everything out of habit. I've dropped the other changes from the PR.

zowoq:
@zimbatm ping

PR #1010, opened 2022-11-02
rc: add DIRENV_STATUS with status of direnv loading

vincentbernat:
It can be "denied" if the environment loading was denied. "failed" if the environment failed to load. "allowed" if everything was successful.


zimbatm (commented):
Can you also add it to the `direnv status` command output? Otherwise looks good.

vincentbernat:
I am not sure what to put exactly. There is already "RC allowed true/false" in the output.

branchvincent:
This would be great to have! That way other tools (like terminal prompts) can check the status without calling `direnv`

branchvincent:
Is there anything I can do to help push this forward?

vincentbernat:
From my side, I don't know what to do as there is already something in `direnv status` to indicate if the environment is loaded or not, unless I am misunderstanding the request.

vincentbernat:
I have rebased the patch to ensure it still works with b80f567dd8e2209acfefdfcbc7ac50ad4276ec3c.

PR #1012, opened 2022-11-03
install.sh: can specify direnv version

### LINKED ISSUE #1011: Able to specify direnv version when installing with `install.sh`
`installer.sh` always installs the latest version of direnv, which might be an unwanted behavior especially if we are building an immutable system.

Able to specify direnv version through environment variable (`version` for example), same as `bin_path`.
example:
```bash
export version="v2.32.1"
curl -sfL https://direnv.net/install.sh | bash
```

aabccd021:
Closes #1011

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -69,9 +69,15 @@ set -euo pipefail
   fi
   echo "bin_path=$bin_path"
 
+  if [[ -n "$version" ]]; then
+    release="tags/${version}"
+  else
+    release="latest"
+  fi
+
   log "looking for a download URL"
   download_url=$(
-    curl -fL https://api.github.com/repos/direnv/direnv/releases/latest \
+    curl -fL "https://api.github.com/repos/direnv/direnv/releases/$release" \
     | grep browser_download_url \
     | cut -d '"' -f 4 \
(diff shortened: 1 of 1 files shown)


zimbatm (approved):
thanks

install.sh
```diff
@@ -69,9 +69,15 @@ set -euo pipefail
   fi
   echo "bin_path=$bin_path"
 
+  if [[ -n "$version" ]]; then
```
zimbatm, line 72:
```suggestion
  if [[ -n "${version:-}" ]]; then
```
otherwise `set -u` will abort the script because the variable is unset

install.sh
```diff
@@ -69,9 +69,15 @@ set -euo pipefail
...
+    release="tags/${version}"
+  else
+    release="latest"
+  fi
```
zimbatm, line 76:
```suggestion
  fi
  echo "release=$release"
```
might as well also print that back out for feedback

install.sh
aabccd021: Thank you, I forgot to add `: "${version:=}"` before the if statement

PR #1019, opened 2022-11-26
Add layout for PDM (python tool)

david-vicente:
An adaptation of #995 to support PDM.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1021,6 +1021,31 @@ layout_pyenv() {
   [[ -n "$PYENV_VERSION" ]] && export PYENV_VERSION
 }
 
+# Usage: layout pdm
+#
+# Similar to layout_python, but uses PDM to build a
+# virtualenv from the pyproject.toml located in the same directory.
+#
+layout_pdm() {
+    PYPROJECT_TOML="${PYPROJECT_TOML:-pyproject.toml}"
+    if [ ! -f "$PYPROJECT_TOML" ]; then
+        log_status "No pyproject.toml found. Executing \`pmd init\` to create a \`$PYPROJECT_TOML\` first."
+        pdm init --non-interactive --python "$(python3 --version 2>/dev/null | cut -d' ' -f2 | cut -d. -f1-2)"
+    fi
+
+    VIRTUAL_ENV=$(pdm venv list | grep "^\*"  | awk -F" " '{print $3}')
+
+    if [ -z "$VIRTUAL_ENV" ] || [ ! -d "$VIRTUAL_ENV" ]; then
+        log_status "No virtual environment exists. Executing \`pdm info\` to create one."
+	      pdm info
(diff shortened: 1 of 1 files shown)


doolio:
Isn't a selling point of `pdm` that it removes the need for a virtual environment?

david-vicente:
PDM can be used in both ways. You can either use virtualenvs, just like Poetry, or you can use [PEP 582](https://peps.python.org/pep-0582/) which doesn't use virtualenvs. However PEP 582 is just a proposal and hasn't been adopted officially by Python yet, so PDM's default is still virtualenv based workflow.

doolio:
Oh that's great to hear. I'm still trying to learn all these tools.

ChauhanT:
Hi, are there plans to include this ? A layout for pdm would be great !

obrienmd:
PEP 582 was rejected and PDM now [recommends virtual environments](https://pdm.fming.dev/latest/usage/pep582/), so I think this commit implements PDM in its desired form as-is. Given PDM seems to be used fairly commonly these days (grumble, grumble, something something python packaging ecosystem), would love to see this shipped by default with direnv.

domdinicola:
any updates on this?

tshu-w:
vote for this

tshu-w:
vote for this

dhensen:
Hi @zimbatm, what do we need to do to get this merged?

stdlib.sh
```diff
@@ -1021,6 +1021,31 @@ layout_pyenv() {
...
 
+# Usage: layout pdm
+#
+# Similar to layout_python, but uses PDM to build a
+# virtualenv from the pyproject.toml located in the same directory.
+#
+layout_pdm() {
+    PYPROJECT_TOML="${PYPROJECT_TOML:-pyproject.toml}"
+    if [ ! -f "$PYPROJECT_TOML" ]; then
+        log_status "No pyproject.toml found. Executing \`pmd init\` to create a \`$PYPROJECT_TOML\` first."
```
thalin, line 1032:
```suggestion
        log_status "No pyproject.toml found. Executing \`pdm init\` to create a \`$PYPROJECT_TOML\` first."
```

stdlib.sh
```diff
@@ -1021,6 +1021,31 @@ layout_pyenv() {
...
+        pdm init --non-interactive --python "$(python3 --version 2>/dev/null | cut -d' ' -f2 | cut -d. -f1-2)"
+    fi
+
+    VIRTUAL_ENV=$(pdm venv list | grep "^\*"  | awk -F" " '{print $3}')
+
+    if [ -z "$VIRTUAL_ENV" ] || [ ! -d "$VIRTUAL_ENV" ]; then
+        log_status "No virtual environment exists. Executing \`pdm info\` to create one."
+	pdm info
```
dairiki:
```suggestion
        log_status "No virtual environment exists. Executing \`pdm install\` to create one."
	pdm install
```

PR #1033, opened 2022-12-21
stdlib/layout_python: exclude patchlevel from $python_version

dne:
There's not much point in keeping separate virtualenvs for different patchlevels from the same Python release branch.


zimbatm:
sounds good, thanks!

PR #1043, opened 2023-01-02
Avoid use of regex in bash hook output

scop:
Not that this would be performance critical, but in principle.

Also makes the match test more exact, rather than hitting anything with a `_direnv_hook` substring.

### DIFF
--- a/internal/cmd/shell_bash.go
+++ b/internal/cmd/shell_bash.go
@@ -15,7 +15,7 @@ _direnv_hook() {
   trap - SIGINT;
   return $previous_exit_status;
 };
-if ! [[ "${PROMPT_COMMAND[*]:-}" =~ _direnv_hook ]]; then
+if [[ ";${PROMPT_COMMAND[*]:-};" != *";_direnv_hook;"* ]]; then
   if [[ "$(declare -p PROMPT_COMMAND 2>&1)" == "declare -a"* ]]; then
     PROMPT_COMMAND=(_direnv_hook "${PROMPT_COMMAND[@]}")
   else


internal/cmd/shell_bash.go
```diff
@@ -15,7 +15,7 @@ _direnv_hook() {
   trap - SIGINT;
   return $previous_exit_status;
 };
-if ! [[ "${PROMPT_COMMAND:-}" =~ _direnv_hook ]]; then
+if [[ ";${PROMPT_COMMAND-};" != *";_direnv_hook;"* ]]; then
```
zimbatm, line 18:
```suggestion
if [[ ";${PROMPT_COMMAND-};" != *"_direnv_hook"* ]]; then
```
The item might be alone in the list or at the start of the list.

scop: The added semicolons we have around the strings both on the left and right sides of the `!=` cover those cases, too.

If it's alone there, the test becomes `[[ ";_direnv_hook;" != *";_direnv_hook;"* ]]` which is false as wanted; if it's the first followed by something else it becomes `[[ ";_direnv_hook;something_else;" != *";_direnv_hook;"* ]]` which is also false as wanted.

Removing the semicolons per the suggestion would cause undesired results e.g. if `$PROMPT_COMMAND` would be `this_is_not_the_direnv_hook_we_are_looking_for`. Theoretical, but we could just do the right thing for that, too :)

R-VdP:
I just wanted to make a PR for this because the current implementation also fails when you have something like this in your `.bashrc`: `alias '!=true &&'`.

This is a contrived example of course, but if we'd move the exclamation mark inside the `[[ ... ]]`, then such a confusion can no longer happen since aliases are not resolved in that case.

If this PR gets merged, the issue also goes away, since there is no negation anymore then.

scop:
Sure, thanks.

zimbatm:
only took a year to get merge :-p

PR #1045, opened 2023-01-03
stdlib: use_guix: Switch to guix shell.

fishinthecalculator:
This patch switches the use_guix behavior to the new version of guix environment, which became the new supported method to spawn on-the-fly environments since the 1.4.0 Guix release.


zimbatm:
thanks!

PR #1048, opened 2023-01-11
Allow disabling direnv from further modifying the environment

### LINKED ISSUE #550: How to disable direnv automatic loading of environments in the current shell?
I cannot find any environment variable for doing this. I know that I can specifically deny a specific env from loading but that implies mutations of the configuration which I do not want to change my direnv configuration.

I would have expected something like:

```bash
export DIRENV_DISABLE=1
cd /my/project/whose/env/is/allowed
# .. Nothing.
```

I tried something like this in my `~/.bashrc`:

```sh
if command -v direnv >/dev/null 2>&1 \
      && ! [ "1" = "${DIRENV_DISABLE:-null}" ]; then
    eval "$(direnv hook bash)"
fi
```

however, I cannot disable it temporarily as follow:

```bash
export DIRENV_DISABLE=1
cd /my/project/whose/env/is/allowed
# .. Nothing.
unset DIRENV_DISABLE
cd /my/project/whose/env/is/allowed
# Loading env
```

I feel that there should be 2 possibilities: `DIRENV_DISABLE` to disable to whole hook (no printing that the env is denied and that one can allow it) and another var such as `DIRENV_DENY_ALL` which would have the same effect as if all env were denied.

reegnz:
When DIRENV_DISABLE environment variable is set, direnv stops manipulating environment variables. When DIRENV_DISABLE environment variable is unset again, direnv picks up from where it left off and start managing the environment variables again.

Keep in mind that setting this variable doesn't unset the variables that direnv has set in the environment, it just toggles direnv, it doesn't roll back all of the effects it has done before the environment variable is set.

Fixes #550

### DIFF
--- a/internal/cmd/cmd_export.go
+++ b/internal/cmd/cmd_export.go
@@ -21,6 +21,12 @@ func exportCommand(currentEnv Env, args []string, config *Config) (err error) {
 	log.SetPrefix(log.Prefix() + "export:")
 	logDebug("start")
 
+	_, set := currentEnv["DIRENV_DISABLE"]
+	if set {
+		logDebug("direnv is disabled by environment variable")
+		return nil
+	}
+
 	var target string
 
 	if len(args) > 1 {


reegnz:
I'm open for a cleanup-solution as well, but I'd rather have both a 'dirty' and a 'clean' disable.

Also this solution is a way simpler implementation than a full cleanup, and people using it should already be familiar with that caveat.

jraygauthier (approved):
LGTM

todd-a-jacobs:
> I'm open for a cleanup-solution as well, but I'd rather have both a 'dirty' and a 'clean' disable.
> 
> Also this solution is a way simpler implementation than a full cleanup, and people using it should already be familiar with that caveat.

The solution there might require having to ascend the tree of .direnv files to collect changes, and then running a new shell with `--login` or `--interactive` preceded by `env -u FOO [...]` with a `-u`  argument for each variable modified by direnv. I say this because I don't think that direnv currently tracks what variables it adds or modifies, especially if multiple .direnv files are invoked as you descend some file tree.

Looking at the previous paragraph, that's probably the missing feature: a pushd/popd type function for direnv that tracks a stack of modifications. I'm not a GO person, so I'm not sure whether or not that's trivial, but it could be as simple as caching the location of each .direnv (and any includes in each .direnv) in a cache file descending from the topmost shell process.

Thoughts or constructive criticisms?

reegnz:
> I say this because I don't think that direnv currently tracks what variables it adds or modifies, especially if multiple .direnv files are invoked as you descend some file tree.

direnv definitely does track what changes it's making, that's what is being stored in the `DIRENV_DIFF` environment variable it sets.

terlar:
This would be really nice, I have some nested shells that triggers direnv while it shouldn't so would be nice to set this to opt-out from the direnv hook if I already entered such a shell.

reegnz:
Same for me, I'm also spawning some nested shells where I don't want direnv to continue loading. Right now I've modified the hook to deal with this:
```
if [ -z "${DIRENV_DISABLE:-}" ]; then
  eval "$(direnv hook zsh)"
fi
```
This works for the nested shell use-case, but if I disable it, I cannot enable it in the nested shell later on (although I specifically don't want to do that for my use-case).

It would be great if this could be a native feature of direnv though.

dephiros:
Stumbled on this when looking for a solution to disable direnv. Would love for this to get merge since it seems ready @reegnz 🙏

Cnly:
Considering it's not currently possible to temporarily disable / enable direnv only within a specific shell, I think this PR would be great. Is there any plan to merge it?

jgb:
Another vote on this PR: it would be quite useful.

eins:
Any update on this ?
@jraygauthier approved this in 2023

reegnz:
Since this PR has still not been merged yet, here's a workaround for zsh for anyone still facing this issue:

For `zsh` users, instead of `eval "$(direnv shell zsh)"` use this zsh script instead in your `.zshrc`:

```zsh
# check if direnv is installed
if (( ! ${+commands[direnv]} )); then
  return
fi

_direnv_hook() {
  trap -- '' SIGINT
  eval "$("/opt/homebrew/bin/direnv" export zsh)"
  trap - SIGINT
}

direnv_hook_disable() {
  typeset -ag precmd_functions
  if (( ${precmd_functions[(I)_direnv_hook]} )); then
    precmd_functions=(${precmd_functions:#_direnv_hook})
  fi
  typeset -ag chpwd_functions
  if (( ${chpwd_functions[(I)_direnv_hook]} )); then
    chpwd_functions=(${chpwd_functions#_direnv_hook})
  fi
}

direnv_hook_enable() {
  typeset -ag precmd_functions
  if (( ! ${precmd_functions[(I)_direnv_hook]} )); then
    precmd_functions=(_direnv_hook $precmd_functions)
  fi
  typeset -ag chpwd_functions
  if (( ! ${chpwd_functions[(I)_direnv_hook]} )); then
    chpwd_functions=(_direnv_hook $chpwd_functions)
  fi
}

direnv_hook_toggle() {
  typeset -ag precmd_functions
  if (( ${precmd_functions[(I)_direnv_hook]} )); then
    direnv_hook_disable
  else
    direnv_hook_enable
  fi
}

direnv_hook_enable
```

For `bash` users, instead of `eval "$(direnv shell bash)"` use this script instead in your `.bashrc`:

```bash
_direnv_hook() {
  local previous_exit_status=$?
  trap -- '' SIGINT
  eval "$("/opt/homebrew/bin/direnv" export bash)"
  trap - SIGINT
  return $previous_exit_status
}

direnv_hook_disable() {
  if [[ ";${PROMPT_COMMAND[*]:-};" == *";_direnv_hook;"* ]]; then
    if [[ "$(declare -p PROMPT_COMMAND 2>&1)" == "declare -a"* ]]; then
      PROMPT_COMMAND=("${PROMPT_COMMAND[@]/_direnv_hook/}")
    else
      PROMPT_COMMAND=";${PROMPT_COMMAND};"
      PROMPT_COMMAND="${PROMPT_COMMAND//;_direnv_hook;/;}"
      PROMPT_COMMAND="${PROMPT_COMMAND#;}"
      PROMPT_COMMAND="${PROMPT_COMMAND%;}"
    fi
  fi
}
direnv_hook_enable() {
  if [[ ";${PROMPT_COMMAND[*]:-};" != *";_direnv_hook;"* ]]; then
    if [[ "$(declare -p PROMPT_COMMAND 2>&1)" == "declare -a"* ]]; then
      PROMPT_COMMAND=(_direnv_hook "${PROMPT_COMMAND[@]}")
    else
      PROMPT_COMMAND="_direnv_hook${PROMPT_COMMAND:+;$PROMPT_COMMAND}"
    fi
  fi
}
direnv_hook_toggle() {
  if [[ ";${PROMPT_COMMAND[*]:-};" != *";_direnv_hook;"* ]]; then
    direnv_hook_enable
  else
    direnv_hook_disable
  fi
}

direnv_hook_enable
```

Then you can toggle the direnv hook with the `direnv_hook_toggle` command.

reegnz:
@eins he's not a maintainer, so his approval is useless in this repo.

eins:
> @eins he's not a maintainer, so his approval is useless in this repo.
I see, thanks for the reply @reegnz .

Kabouik:
I would love to see this PR merged too. I use `direnv` to deploy Guix reproducible environments which are useful when code needs to be evaluated, but there are times when I know I just want to edit some text files or visit the directory without actually evaluating code. In those cases, entering the local shell is superfluous and can even be an issue (like if it was never created, or was garbage collected, and its creation requires compiling a lot of stuff). This simple PR would help greatly for that use case.

PR #1052, opened 2023-01-17
Fix using PWD in .env files

### LINKED ISSUE #1051: Cannot reference PWD in .env file for variable expansion
**Describe the bug**
Direnv allows for variable expansion in `.env` files. The `PWD` environment variable is however [removed in the direnv process](https://github.com/direnv/direnv/blob/master/internal/cmd/mod.go#L17) from the environment variables to not interfere with finding .envrc.

**To Reproduce**
In any folder put the following two files:
```.envrc
dotenv
```

```.env
TEST1=$PWD
TEST2=$HOME
```

TEST1 will be empty, while TEST2 will have expanded properly.

**Expected behavior**
PWD in .env files should expand to the folder the .env file resides in.

**Environment**
 - OS: [e.g. macOS Ventura]
 - Shell: [zsh]
 - Direnv version [e.g. 2.32.2]

**Additional context**
Other .env impementations don't have this issue. I've checked the following, which all seem to work as expected:
https://github.com/bkeepers/dotenv
https://github.com/motdotla/dotenv-expand

Both work as expected and expand PWD properly.

reegnz:
This change introduces PWD support to the `dotenv` functionality, so that it can resolve PWD in `.env` files.

When .env is loaded, PWD will be set to the same folder that .env file can be found in variable can be used.
A user would also expect that the PWD variable is matching the folder that the .env file resides in, and symlinks should not interfere with that either.

Fixes #1051

### DIFF
--- a/internal/cmd/cmd_dotenv.go
+++ b/internal/cmd/cmd_dotenv.go
@@ -1,9 +1,9 @@
 package cmd
 
 import (
+	"path/filepath"
 	"fmt"
 	"os"
-
 	"github.com/direnv/direnv/v2/pkg/dotenv"
 )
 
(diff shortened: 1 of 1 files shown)


seanorama (approved):

seanorama:
This would help a lot! Can we get it merged?

zimbatm:
Just pushed a small fix, otherwise it looks good to go. Let me know what you think!

reegnz:
Hey @zimbatm I'm not entirely sure why the resolving of the softlinks is a problem. Could you elaborate?
I've found it to be more reliable if there's no ambiguity in case folders are symlinked to other places.

reegnz:
OK, I've contemplated on it a bit, let's get this merged, I think I'm satisfied with this option.

zimbatm:
It depends on the usage. Maybe some users want to use the `.env` file as a template and symlink it into many places. Let's say the file contains `PATH=$PWD/bin:$PATH`. Then, it would be weird if the PATH contained a reference to the source folder instead of the destination.

zimbatm:
finally got around to do release v2.34.0

PR #1061, opened 2023-02-15
+1 use case

ask0n:
We are using direnv as a replacement for [terragrunt](https://terragrunt.gruntwork.io/).
More details can be found [here](https://medium.com/@senior-devops/how-to-extend-terraform-with-direnv-a4a3fef092c5).


zimbatm:
thanks, I want to keep that list short. Can you add this to the wiki instead? For example in https://github.com/direnv/direnv/wiki/Terraform

Issue #1065, opened 2023-02-28
Skip "taking a while to execute" Warning


F1ko:
# Is your feature request related to a problem? Please describe.

Upon reaching a (configurable) timeout a warning (`[...] taking a while to execute. Use CTRL-C to give up.`) is being printed.

However, in certain situations it may be a good idea to skip the warning altogether instead of raising the timeout limit.
An example could be a debugging project that starts a certain setup and interactive shell.
The sole purpose of such a project would be to enter the directory, set a couple of environment variables followed by starting a debug container and an interactive shell which will most likely never terminate (and therefore the warning will be printed sooner or later regardless of the configurable timout).

# Describe the solution you'd like

Setting a global or local variable to not show the warning.

# Describe alternatives you've considered

Two possible alternatives:
- Not using direnv and wasting time by copy-pasting instructions from a README to start interactive commands.
- Using direnv (perhaps even setting a high timeout value) and at some time being suprised by a warning popping up

# Additional context

n/a

perlun:
> Upon reaching a (configurable) timeout a warning (`[...] taking a while to execute. Use CTRL-C to give up.`) is being printed.

Slightly OT, but where can this be configured? :thinking: I couldn't find it in the docs.

kumy:
https://github.com/direnv/direnv/blob/9afbc87c58e23f32d72bb98d4e71d507a143c6b3/man/direnv.toml.1.md#warn_timeout

> This feature is disabled if the duration is lower or equal to zero. Will be overwritten if the environment variable DIRENV_WARN_TIMEOUT is set to any of the above values.

so `export DIRENV_WARN_TIMEOUT=0` or setting the `warn_timeout = 0` in `direnv.toml` should already do it?

perlun:
> so `export DIRENV_WARN_TIMEOUT=0` or setting the `warn_timeout = 0` in `direnv.toml` should already do it?

Thanks. :bow: Tried this, but it didn't work. `DIRENV_WARN_TIMEOUT=0` seemed to make the warning appear immediately.

But, setting it to a higher value like `DIRENV_WARN_TIMEOUT=100s` does seem to work. :tada: This will do the trick for me for now, thanks for the suggestion.

synthead:
Honestly, setting it to 0 and having the message immediately appear feels like a bug.  It'd be great if this disables the timeout instead.

PR #1066, opened 2023-02-28
Skip timeout

f1ko:
Setting `export DIRENV_SKIP_TIMEOUT=TRUE`  on a global level will skip all further timeout warnings accordingly.

Fixes https://github.com/direnv/direnv/issues/1065

### DIFF
--- a/internal/cmd/commands.go
+++ b/internal/cmd/commands.go
@@ -72,15 +72,16 @@ func init() {
 func cmdWithWarnTimeout(fn action) action {
 	return actionWithConfig(func(env Env, args []string, config *Config) (err error) {
 		done := make(chan bool, 1)
-		go func() {
-			select {
-			case <-done:
-				return
-			case <-time.After(config.WarnTimeout):
-				logError("(%v) is taking a while to execute. Use CTRL-C to give up.", args)
-			}
-		}()
-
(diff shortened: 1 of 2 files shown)


f1ko:
Hi @zimbatm , could you please take a look at this small PR?

There is no breaking changes whatsoever and it would fix two issues at the same time (https://github.com/direnv/direnv/issues/1065 and https://github.com/direnv/direnv/issues/1084).

darklow:
We're experiencing same issue and this PR would really help it.

internal/cmd/commands.go
```diff
@@ -72,15 +72,16 @@ func init() {
...
-		go func() {
-			select {
-			case <-done:
-				return
-			case <-time.After(config.WarnTimeout):
-				logError("(%v) is taking a while to execute. Use CTRL-C to give up.", args)
-			}
-		}()
-
+		if env["DIRENV_SKIP_TIMEOUT"] != "TRUE" {
```
zimbatm, line 75: I'm trying to move away from env vars to configure direnv as it has some weird recursive implications.

How about extending the config.WarnTimeout semantic so that <= 0 means no wait?

zimbatm:
I implemented this slightly differently in #1209. Thanks for nudging me!

PR #1071, opened 2023-03-08
stdlib: use_node: strip leading v from version

forivall:
In line with [this helpfully compiled information on usage of .node-version files' suggested compatible format of .node-version files](https://github.com/shadowspawn/node-version-usage#suggested-compatible-format), it would be helpful for direnv to strip off leading `v` prefixes of a version specifier for broader compatability.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1185,6 +1185,8 @@ use_node() {
     via=".node-version"
   fi
 
+  version=${version#v}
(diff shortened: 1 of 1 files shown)


srl295 (approved):

shadowspawn:
See also #833, where I found one other place with similar code.

Issue #1079, opened 2023-03-23
referenced .envrc does not exist


laccd-minassn:
Windows image on direnv for version 2.32.2 (works in prev versions). When using Git Bash on Windows, entering a directory that has .envrc, produced a error referencing the file where the path is garbled:

![image](https://user-images.githubusercontent.com/117323120/227383393-f7b028d8-b8ac-4aab-9e18-3b1f8c684339.png)

Previously similar issue: #707 
![Screenshot 2023-03-23 160311](https://user-images.githubusercontent.com/117323120/227383521-1d2a6a32-9f22-43d3-ae32-3fba3afb91dd.png)

daniel-bartley:
I get the same bug on windows:

```
scoop install direnv
echo eval "$(direnv hook bash)" > ~/.bashrc
source ~/.bashrc

# for debugging 
echo $PWD
echo direnv --version
echo bash --version

# test
mkdir -p ~/random 
~/random/.envrc >> "echo Hello direnv"
cd random
direnv allow 
direnv reload

# test again
cd ..
cd random/

```

Debug/Error messages:

```
$ echo $PWD
/c/Users/1234567/
$ direnv --version
2.32.2
$ bash --version
GNU bash, version 5.2.12(1)-release (x86_64-pc-msys)

$ source ~/.bashrc
andom/.envrc does not exist^4567
direnv: error exit status 1

$ direnv reload
andom/.envrc does not exist^4567
direnv: error exit status 1

```

Issue #1020, opened 2022-11-28
Regression in 2.32.2 on Windows: Certain path characters are interpreted as escape sequences


goostleek:
**Describe the bug**

If one of the directory names in a path to the `.envrc` starts with certain characters (eg. `b`) then `direnv` mangles the path while searching for the `.envrc`

**To Reproduce**

1. Create a `dir` directory and `dir/.envrc` file then navigate to `dir`

```bash
mkdir dir && cd dir && touch .envrc
```
In this case, everything works as expected, `.envrc` is loaded 👌 
```
direnv: loading /d/dir/.envrc
```

2. Still being in `dir` directory create a `bug` directory and `bug/.envrc` file then navigate to `bug`

```bash
 mkdir bug && cd bug && touch .envrc
```

In this case, `direnv` mangles the path and fails 💥 

```
direnv: referenced /d/diug/.envrc does not exist
direnv: error exit status 1
```

**Expected behavior**

Mind the broken path `/d/diug/.envrc`. It should be `/d/dir/bug/.envrc`. It seems the `.../b...` character  was interpreted as a `backspace` escape sequence.

**Environment**
 - OS: `MSYS_NT-10.0-19044`
 - Shell: `bash`
 - Direnv version: `2.32.2`

**Additional context**

It seems the bug is a regression introduced in the `2.32.2` release. It does not occur in a previous `2.32.1` version.

goostleek:
After a quick look, I think it has something to do with https://github.com/direnv/direnv/pull/975 (I can be wrong).

rminderhoud:
Confirming there is a regression here specifically introduced in version 2.32.1

```
# 2.32.2
$ direnv allow
ose-online/.envrc does not exist
direnv: error exit status 1

#2.32.1
$ direnv allow
direnv: referenced C:\dev\rose\rose-online\.envrc does not exist

# 2.32.0
$ direnv allow
direnv: loading C:\dev\rose\rose-online\.envrc
direnv: export +POSTGRES_PORT
```

PR #1085, opened 2023-04-08
Fix incorrect escape sequences during Loads under git-bash

SingingTree:
https://github.com/direnv/direnv/pull/975 had the unintended consequence of causing the `RC::Load` call to sometimes fail when running under git-bash (and some others Windows specific configs).

The `RC::Load` -> `cmd.Output()` calls end up interpreting characters like \b or \t as escape sequences, which is problematic for paths like 'myProject\buckets\test'.

This patch addresses this by forcing the use of `/` as a separator, even on Windows. This works fine in my testing in git-bash. This shouldn't impact systems where `/` is already in use.

This also adds a test case that covers the behaviour being adjusted. In order to run the bash tests on Windows, I had to disable a number that appear to not play nice -- but after hacking those out I was able to verify the new test case.

Fixes:
- https://github.com/direnv/direnv/issues/1020
- https://github.com/direnv/direnv/issues/1079

### DIFF
--- a/internal/cmd/rc.go
+++ b/internal/cmd/rc.go
@@ -183,12 +183,16 @@ func (rc *RC) Load(previousEnv Env) (newEnv Env, err error) {
 		prelude = "set -euo pipefail && "
 	}
 
+	// Non-Windows platforms will already use slashes. However, on Windows
+	// backslashes are used by default which can result in unexpected escapes
+	// like \b or \r in paths. Force slash usage to avoid issues on Windows.
+	slashSeparatedPath := filepath.ToSlash(rc.Path())
 	arg := fmt.Sprintf(
 		`%seval "$("%s" stdlib)" && __main__ %s %s`,
 		prelude,
 		direnv,
 		fn,
-		BashEscape(rc.Path()),
+		BashEscape(slashSeparatedPath),
 	)
 
 	// G204: Subprocess launched with function call as argument or cmd arguments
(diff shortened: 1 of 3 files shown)


mloskot:
Any chance to get this reviewed and merged?
I've packaged direnv for winget and I noticed similar issue while testing it with Git Bash, see https://github.com/direnv/direnv/pull/1096 It would be awesome if this could be fixed, hopefully with this PR.

zimbatm (commented):
Thanks!

Do you know if `direnv edit` is affected by the same issue? There is another instance of path escaping in internal/cmd/cmd_edit.go:60

mloskot:
> tinkering shows the existing \ behaviour only seems to cause issues in rare cases.

Yes, I think that is correct conclusion.
e.g. I recall paths like `d:\tmp` or `d:\azure` as troublesome

zimbatm:
thanks

mloskot:
I've just packaged the latest release for winget, https://github.com/microsoft/winget-pkgs/pull/107614, and installed it locally.
I followed the tutorial in Git Bash in `/d/tmp` akd `D:\tmp` and all works fine:

image

Thanks for the fix!

zimbatm:
thanks, can you also try `direnv edit`?

mloskot:
@zimbatm AFAICT, it works like a charm!

https://github.com/direnv/direnv/assets/80741/0d661736-7412-4990-b3af-e1655faf37b6

PR #1086, opened 2023-04-08
Fix golangci-lint issues

SingingTree:
Fixes issues preventing golangci-lint from passing. Following this `golangci-lint run` and `make test` (which calls into golangci-lint) should not fail at the lint step.

- Change unused argument names to `_`.
- Add toml and json tags to a couple of structs that are serialized as toml and json respectively.


zimbatm:
thanks!

PR #1099, opened 2023-05-20
Update direnv.toml.1.md

Ativerc:
Made the description for global > load_dotenv more clearer.

Added some backticks around a "true".


Ativerc:
The documentation needs to mention the minimum version of direnv that supports `load_dotenv`. I have version 2.21.2 and I thought it would work but it doesn't.

PR #1140, opened 2023-08-01
fix(test): test-in-envrc was a noop

bukzor-sentryio:
Previously, the `test -n "$DIRENV_IN_ENVRC"` was not being run by direnv, since it did not appear in `.envrc`. This wasn't noticed because there was no assertion of the `echo "in envrc"`.

Now, an environment value is exported and checked, which prevents this kind of mistake in future.

### DIFF
--- a/test/direnv-test-common.sh
+++ b/test/direnv-test-common.sh
@@ -160,12 +160,14 @@ test_start "empty-var-unset"
 test_stop
 
 test_start "in-envrc"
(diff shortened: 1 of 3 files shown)


zimbatm:
thanks, looks like the fish test is still broken:
```
fish: Unknown command: ./test-in-envrc
./test/direnv-test.fish (line 176): 
    ./test-in-envrc
    ^
```

Issue #520, opened 2019-09-17
[question] Avoiding `direnv allow` for usage in scripting?


nottrobin:
I'd like to use direnv from within a Python tool I'm writing. So I want to be able to run something like:

``` python3
subprocess.check_call(["direnv", "exec", ".", "my-command"])
```

So that I can allow the user to have made use of a `.envrc` file. The biggest blocker to this that I'm having is the need for `direnv allow .`.

Is there any way that direnv can be configured by the user to avoid the need for `direnv allow .` on each directory?

zimbatm:
If the command was to fail, for example there is a typo on the `.envrc`. How would you display that to the user?

The best thing to do is to let the user run the command after inspecting the `.envrc`. Anything else would circumvent the purpose of that protection and put the user at risk.

nottrobin:
Yeah thanks I can see why that's necessary.

New question - is there any way for me to neatly check from a script for whether direnv considers a `.envrc` to be allowed or not? I found `direnv status`, but ideally I'd like a command that returns `0` if the `.envrc` is allows, and `1` otherwise. Does such a command exist?

zimbatm:
Looking at what the VSCode plugin does, it looks like they are parsing the output when it fails. Probably the most robust way though. https://github.com/rubymaniac/vscode-direnv/blob/014f12bb5295b6e6a37c7dda97d8bfc1ee8b40ab/src/main.ts#L36

I wouldn't be against adding a `direnv status --json` that contains that type of information.

zimbatm:
related: https://github.com/direnv/direnv/issues/244#issuecomment-520487528

Issue #244, opened 2017-01-10
Is there a way I can tell if direnv hooks have been set?


agibralter:
I'm trying to figure out a way to see if (e.g. for bash) `eval "$(direnv hook bash)"` has been run in the current session. Is there a consistent way to do that across shells?

yermulnik:
`direnv status`?

zimbatm:
For bash, one way is `echo $PROMPT_COMMAND | grep direnv_hook`. Each shell has a different way of providing the hook, have a look at the output of `direnv hook zsh` for example. If you want to make that detection generic I think a new command is need so that `eval "$(direnv hook-detect bash)"` would return 0 if direnv is detected. These settings are set on the shell and not exported to the `direnv status` sub-process unfortunately.

zimbatm:
Can you explain a little bit where the requirement is coming from?

agibralter:
@zimbatm ahh interesting, thank you!

Basically I wanted to build a helper for local development at my company that leverages `direnv`. I want one of my shell helpers to warn users if they have not "installed" `direnv` correctly.

mgarstecki:
I need this too, and for the same use case: we use `direnv` in our scripting, and I'm writing a script to check the workstation environment is sane.

We leave our developers free on their work environment (Linux/MacOS, bash/zsh, ...) so grepping dotfile or checking the prompt would be brittle.

A specific line in `direnv status` to check if it is active in the current shell would be very useful to this use case :)

zimbatm:
I wouldn't mind getting a PR for `direnv status --json`. Right now the output is for humans and not guaranteed to remain the same. Have a look in `cmd_status.go` and you'll see it should be pretty straight-forward.

The only way to detect reliably would be if `loadedRc` is nil and `foundRC` is not. In that case it means the hook didn't spring into action.

Proposed output:
```json
{
  "config": {
    "ConfigDir": "~/.config/direnv",
    "BashPath": "/usr/bin/bash",
    "SelfPath": "/usr/local/bin/direnv"
  },
  "state": {
    "foundRC": null,
    "loadedRC": null
  }
}
```

mishoo78:
hello good people and thank you @zimbatm for the answer. i think it was the right one.

PR #1142, opened 2023-08-11
`direnv status --json`

shivaraj-bh:
Was mentioned twice:
- https://github.com/direnv/direnv/issues/244#issuecomment-520487528
- https://github.com/direnv/direnv/issues/520#issuecomment-532892797

It will be used here:
https://github.com/srid/nix-health/issues/4

I have very little exposure working with GO, please correct me if I am wrong anywhere.

### DIFF
--- a/internal/cmd/cmd_status.go
+++ b/internal/cmd/cmd_status.go
@@ -3,40 +3,78 @@ package cmd
 import (
 	"fmt"
 	"path/filepath"
+	"encoding/json"
 )
 
 // CmdStatus is `direnv status`
 var CmdStatus = &Cmd{
 	Name: "status",
 	Desc: "prints some debug status information",
+	Args: []string{"[--json]"},
 	Action: actionWithConfig(func(env Env, args []string, config *Config) error {
-		fmt.Println("direnv exec path", config.SelfPath)
-		fmt.Println("DIRENV_CONFIG", config.ConfDir)
-
-		fmt.Println("bash_path", config.BashPath)
-		fmt.Println("disable_stdin", config.DisableStdin)
-		fmt.Println("warn_timeout", config.WarnTimeout)
-		fmt.Println("whitelist.prefix", config.WhitelistPrefix)
-		fmt.Println("whitelist.exact", config.WhitelistExact)
-
-		loadedRC := config.LoadedRC()
-		foundRC, err := config.FindRC()
-		if err != nil {
-			return err
-		}
-
-		if loadedRC != nil {
-			formatRC("Loaded", loadedRC)
+		if (len(args) > 1 && args[1] == "--json") {
+			loadedRC := config.LoadedRC()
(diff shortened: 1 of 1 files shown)


zimbatm:
hey, thanks for the contrib!

The change looks reasonable overall. The only thing that needs to be looked into is why the loadedRC is not marked as allowed, even though it was just loaded:

```
{
  "config": {
    "ConfigDir": "/home/zimbatm/.config/direnv",
    "SelfPath": "/home/zimbatm/go/src/github.com/direnv/direnv/direnv"
  },
  "state": {
    "foundRC": {
      "allowed": true,
      "path": "/home/zimbatm/go/src/github.com/direnv/direnv/.envrc"
    },
    "loadedRC": {
      "allowed": false,
      "path": "/home/zimbatm/go/src/github.com/direnv/direnv/.envrc"
    }
  }
}

```

shivaraj-bh:
> The change looks reasonable overall. The only thing that needs to be looked into is why the loadedRC is not marked as allowed, even though it was just loaded

I am guessing that's because `LoadedRC()` calls `RCFromEnv` that returns a `RC` struct with empty `allowPath` and the only other way `Allowed()` will be true is if `direnv.toml` whitelists a path or prefix. 
https://github.com/direnv/direnv/blob/cece664987549dade37a4314feaf7084190a6685/internal/cmd/rc.go#L89-L117 

Is this a bug?

shivaraj-bh:
@zimbatm I have raised a [PR](https://github.com/direnv/direnv/pull/1157) to fix this, let me know if my understanding here is right.

srid:
I would love to see this get merged and included in next release.

thenbe:
I was about to open an issue about the output of `direnv status --json` being unintuitive. But when I find this thread, I started suspecting that I am misunderstanding the output.




Does the output seem confusing to you guys? Or am I misunderstanding it?

For context, all I'm trying to detect is whether direnv has already been set or if `direnv allow` needs to be called. From this output, I can conclude that `direnv allow` has already been set if `allowed = 0`, which is what I meant by "unintuitive" earlier.

shivaraj-bh:
@thenbe You are right, it does seem unintuitive. 

The reason why you see integers now instead of bool (which was the case when the PR was merged), is due to this [PR](https://github.com/direnv/direnv/pull/1158). To be specific, these specific lines from `internal/cmd/rc.go`: 
```go
type AllowStatus int
const (
	Allowed AllowStatus = iota
	NotAllowed
	Denied
)
```
Due to the use of `iota` the states get assigned 0, 1, 2 in the order that they are written above. That justifies the output you see.

FWIW, I think it could very well be an issue to improve the readability of the output, feel free to create one.

thenbe:
Oh, this makes sense now that I realize the number represents a status code and not a counter of some sort.

As for improving readability, I've opened a separate ticket here #1211 with one suggestion for how we can do so.

PR #1144, opened 2023-08-11
Print correct path in source_env log message

### LINKED ISSUE #1143: source_env prints incorrect path in its log message
**Describe the bug**

When using `source_env` to source `.envrc` from a subdirectory, the subdirectory is printed twice in the printed path.

**To Reproduce**

Run this in bash:
```bash
mkdir -p subdir
EDITOR=tee direnv edit subdir <<<"echo From subdir"
EDITOR=tee direnv edit . <<<"source_env subdir"
```

You will get something like:
```
direnv: loading ~/tmp/.envrc                                                                                                                               
direnv: loading ~/tmp/subdir/subdir/.envrc
From subdir
```

**Expected behavior**
The printed message should be:
```
direnv: loading ~/tmp/.envrc                                                                                                                               
direnv: loading ~/tmp/subdir/.envrc
From subdir
```
The difference is in the middle line.

**Environment**
 - OS: NixOS
 - Shell: should not matter (zsh)
 - Direnv version 2.32.3

**Additional context**

I'll post a PR with a fix.

wentasah:
Without this change, when source_env loads a file from subdirectory the subdirectory is duplicated in the printed path. For example, if `~/tmp/.envrc` contains:

    source_env subdir

direnv prints:

    direnv: loading ~/tmp/subdir/subdir/.envrc

Fixes #1143

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -349,7 +349,7 @@ source_env() {
   pushd "$(pwd 2>/dev/null)" >/dev/null || return 1
   pushd "$rcpath_dir" >/dev/null || return 1
   if [[ -f ./$rcpath_base ]]; then
-    log_status "loading $(user_rel_path "$(expand_path "$rcpath")")"
+    log_status "loading $(user_rel_path "$(expand_path "$rcpath_base")")"
     # shellcheck disable=SC1090
     . "./$rcpath_base"
   else


zimbatm:
awesome thanks!

@mergify merge

PR #1146, opened 2023-08-13
Update nixpkgs and shellcheck

Mic92:


stdlib.sh
```diff
@@ -746,9 +746,9 @@ semver_search() {
 # A semantic dispatch used to describe common project layouts.
 #
 layout() {
-  local name=$1
+  local funcname="layout_$1"
   shift
-  eval "layout_$name" "$@"
+  "$funcname" "$@"
```
Mic92, line 751: Doublecheck this one. I'd say it shouldn't be required to have eval here.

zimbatm: Yes that's better

PR #1157, opened 2023-08-16
`allowPath` for `LoadedRC`

### LINKED ISSUE #1156: Empty allowPath for `LoadedRC` in `direnv status`
**Describe the bug**
`RCFromEnv` function creates a `RC` struct with empty `allowPath`, as a result `Loaded RC allowed ` is always `false` unless there is `direnv.toml` whitelisting the path or .envrc file. 

**To Reproduce**
Run `direnv status` on any directory which has a `.envrc` file and `direnv allow` has been run. You will see that `Loaded RC allowPath` is empty.

**Expected behavior**
`allowPath` shouldn't be empty to 

**Environment**
 - OS: MacOS, Linux
 - Shell: zsh
 - Direnv version: 2.32.3

shivaraj-bh:
Resolves #1156

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -204,11 +204,11 @@ func (config *Config) AllowDir() string {
 
 // LoadedRC returns a RC file if any has been loaded
 func (config *Config) LoadedRC() *RC {
-	if config.RCFile == "" {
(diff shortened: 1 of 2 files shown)


zimbatm:
thanks!

PR #1158, opened 2023-08-23
Don't prompt to allow if user explicitly denied

Gabriella439:
Currently if an `Allow` check fails direnv will prompt the user to `direnv allow` the file.  However, `direnv` will do that even if the user explicitly ran a `direnv deny` command, which is not necessarily the desired behavior.

In particular, a user might have `direnv` installed and enabled but might not wish to enable the `.envrc` file for every repository that they interact with.  However, currently `direnv` will always prompt the user to `direnv allow` every time they `cd` into a repository with a `.envrc` file.

This fixes that by recording when the user explicitly denied a `.envrc` file so that `direnv` knows not to prompt again.

There is one key difference between how `direnv allow` and `direnv deny` record the hash.  `direnv allow` records a grant based on the file's (name × content) hash whereas `direnv deny` records a revocation based on the file's name alone.  The reason behind this difference is so that `direnv` doesn't keep prompting the user to `direnv allow` a file if the file changes over time (e.g. the user pulls in new changes to the file from the repository origin).

This change also includes some small fixes to not fail or emit a warning if a user runs `direnv allow` or `direnv deny` twice in a row. Before those fixes `direnv` would fail with an error message like:

```
direnv: error stat …/.local/share/direnv/allow/…: no such file or directory
```

… and after this change `direnv` will silently exit (successfully) if there is nothing to be done.

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -202,6 +202,11 @@ func (config *Config) AllowDir() string {
 	return filepath.Join(config.DataDir, "allow")
 }
 
+// DenyDir is the folder where all the "deny" files are stored.
+func (config *Config) DenyDir() string {
+	return filepath.Join(config.DataDir, "deny")
+}
+
 // LoadedRC returns a RC file if any has been loaded
 func (config *Config) LoadedRC() *RC {
 	if config.Env[DIRENV_FILE] == "" {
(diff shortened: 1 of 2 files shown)


Gabriella439:
cc: @zimbatm - do i need to ping someone for review?

zimbatm:
No it's just me that is being busy

zimbatm:
thanks!

PR #1162, opened 2023-09-06
add support for armv7l platform in install.sh

ardje:
Hello,
the machine=arm is known under several uname -m's.
I have added 2 fixes:
1) armv7l -> arm
2) in the part where the download url is found, both arm64 and arm is found for machine=arm so I added an end of line anchor to the grep.

2) can have unforeseen consequences if other machines expect a loose match instead of a rigid match.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -40,6 +40,9 @@ set -euo pipefail
     i686 | i386)
       machine=386
       ;;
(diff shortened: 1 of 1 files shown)


zimbatm:
thanks!

PR #1163, opened 2023-09-08
stdlib: add use_flakepath function

Lillecarl:
Function is very similiar to use_flake, except it parses the PATH variable of the nix environment and appends it.

Users of other shells that don't want to use "shellHook" to exec back into their prefered shells, and only need PATH modified might appreciate this.

In contradiction to use_flake we don't create a profile, if you use nix-collect-garbage while inside a shell like this, you'll lose your PATH entries.

@zimbatm I'd like your input, since you're the author of use_flake 😄

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1290,13 +1290,7 @@ use_flakepath() {
   watch_file flake.nix
   watch_file flake.lock
 
-  OLDIFS=$IFS
-  IFS=":"
-
-  for path in $(nix print-dev-env "$@" --json | jq ".variables.PATH.value" -r); do
-    PATH_add $path
-  done
-  export IFS=$OLDIFS
+  PATH_add "$(nix develop -c sh -c 'echo $PATH')"`
 }
 
 # Usage: use_guix [...]
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -377,7 +377,7 @@ source_env_if_exists() {
 
 # Usage: env_vars_required <varname> [<varname> ...]
 #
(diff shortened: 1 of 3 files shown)


stdlib.sh
```diff
@@ -1278,6 +1278,27 @@ use_flake() {
...
+  watch_file flake.nix
+  watch_file flake.lock
+
+  OLDIFS=$IFS
+  IFS=":"
+
+  for path in $(nix print-dev-env "$@" --json | jq ".variables.PATH.value" -r); do
+    PATH_add $path
+  done
+  export IFS=$OLDIFS
```
zimbatm, line 1299: This would remove the dependency on jq:
```suggestion
  PATH_add "$(nix develop -c sh -c 'echo $PATH')"`
```

zimbatm (commented):
nice

stdlib.sh
Lillecarl: Copilot suggested me this:

```export PATH=$(nix develop . --command printenv PATH)``` However that'd require people to have printenv (which is part of coreutils). Considering it's this simple when direnv takes care of the PATH diffing I'm not sure if it fits better as a footnote in the docs somewhere rather than a "first class function".

Thoughts? 😄

zimbatm: Better to use `sh -c 'echo $PATH'` because `/bin/sh` is on every Unix machine

zimbatm:
yeah, it seems a bit too simple for the stdlib. Maybe add it to the wiki instead?

Lillecarl:
I'll look into it 😄

PR #1169, opened 2023-09-15
Set .envrc file permission to 0600

### LINKED ISSUE #1168: Created .envrc file is world-readable by default
**Describe the bug**

`direnv edit .` creates `.envrc` files with file permission 644, i.e. readable by any user.

**To Reproduce**

```
direnv edit .
stat .envrc
```

**Expected behavior**

File is has perms 640 or 600, like SSH private keys created by ssh-keygen.

**Environment**
 - OS: Linux 6.5.3
 - Shell: zsh
 - Direnv version: 2.32.3

**Additional context**

avorima:
The file permission is initially set by the EDITOR. For normal text files 0644 makes sense, but envrc files can contain sensitive information that should not be world-readable.

Set the file permission of .envrc to 0600 before writing the allow file.

Fixes #1168


sambacha:
This would be dirtying the worktree, as git only maintains \`**100644\` octet by default** 

**git is single user. It does not track file ownership and permissions sufficient to secure a multiple user system or sensitive information.**

avorima:
I did not consider that someone would actually commit their envrc file to git. I guess this behavior should be configurable then

zimbatm:
Agreed with Sam; it's likely that this change is going to break existing use cases.

I think the answer is to introduce a mechanism for secret handling, like a new "get_secret <name>" stdlib function with pluggable backends.

```sh
secret_backend sops ./secrets.yaml
# or
secret_backend 1password MyWallet

export GITHUB_TOKEN=$(get_secret GITHUB_TOKEN)
# or
export_secret GITHUB_TOKEN
```

avorima:
You can already do that with (go)pass for example, so I don't know if it's worth adding something like this to the stdlib.

Anyways, thanks for looking into it.

PR #1170, opened 2023-09-18
man/direnv-stdlib: fix obsolete opam-env example

mzacho:
It seems that `opam-env` has been changed to `opam env`, and `opam-env exec` has been changed to `opam exec`.  Thus attempting to use the current example yields the error:

```
environment:516: opam-env: command not found
```

This PR fixes the example to

```
direnv_load opam exec direnv dump
```

Which has the intended effect of dumping the environment generated by running `eval $(opam env)` using `direnv dump`, and then wrapping the result with `direnv_load`.

Tested using opam version 2.1.4.


zimbatm:
thanks!

PR #1171, opened 2023-09-22
Add PowerShell Support

### LINKED ISSUE #124: PowerShell support
Alternatively to Cygwin, it should be possible to support the Microsoft PowerShell. The only unclear requirement that needs to be researched is if PowerShell would support evaluating code before each prompt.
- Check how eval works and if there is a hook for executing direnv before each prompt
- Install bash trough http://mingw.org/wiki/msys
- Add a new shell_powershell.go that supports generating `set` instructions from the env diff.

bamsammich:
Fixes #124 

Add support for PowerShell. Used [this](https://github.com/PowerShell/PowerShell/issues/14484#issuecomment-1731647083) as an example to create the hook.

I chose to be vague in documentation as to which profile the hook call should be added to. Also needed to copy `BashEscape` to use backticks instead of backslashes for escaping special characters and to remove the `$` before the output.


bamsammich:
Saw the shell-specific test suite and am translating tests in `direnv-test-common.sh` to a PowerShell script. Commit for tests is coming soon.

bamsammich:
> Saw the shell-specific test suite and am translating tests in `direnv-test-common.sh` to a PowerShell script. Commit for tests is coming soon.

All tests that exist in `test/direnv-test-common.sh` have been translated for PowerShell in `./test/direnv-test.ps1`

zimbatm (approved):
very cool, thank you

PR #1172, opened 2023-09-23
update NixOS installation instructions

Gerg-L:
Changed NixOS installation URL to the options search instead of the direnv package


Mic92:
@mergify queue

Issue #1177, opened 2023-10-04
DeprecationWarning during Python version discovery with Python 3.12.0


SinclaM:
**Describe the bug**

After updating to Python 3.12, the following DeprecationWarning will appear when direnv loads a Python environment:

```bash
<string>:1: DeprecationWarning: 'pkgutil.find_loader' is deprecated and slated for removal in Python 3.14; use importlib.util.find_spec() instead
```

**To Reproduce**

* Ensure Python has been updated to 3.12.0
* Create an `.envrc` with the content `layout python3` in a new empty folder and authorize it with `direnv allow`
* Upon entering the newly created python environment, the DeprecationWarning will appear

**Expected behavior**
The Python environment should load without warnings.

**Environment**
 - OS: macOS 12.6.7
 - Shell: zsh
 - Direnv version: 2.32.3

**Additional context**

The cause of the warning is the now-deprecated `pkgutil.find_loader` function, used in this line:
https://github.com/direnv/direnv/blob/93cd37820c29a1f96a0e64ff1557a7a912866d4f/stdlib.sh#L824

SinclaM:
Fixed as of af06cb0.

pro-sumer:
When can we expect a release that contains that PR?

(I'm suddenly getting this warning and it's annoying)

kakulukia:
pkgutil.find_loader is deprecated << im also seeing this with direnv 2.35.0 and python 3.13 is it not fully fixed?

qadzek:
To suppress this warning:

```
export PYTHONWARNINGS="ignore::DeprecationWarning"
layout python
unset PYTHONWARNINGS
```

Ray-Eldath:
> pkgutil.find_loader is deprecated << im also seeing this with direnv 2.35.0 and python 3.13 is it not fully fixed?

me too. seeing this on direnv version 2.37.1:
```
direnv: loading /home/cadfd3db-f9fd-470b-b319-74fd83f27e89/vllm/sshare/.envrc
<string>:1: DeprecationWarning: 'pkgutil.find_loader' is deprecated and slated for removal in Python 3.14; use importlib.util.find_spec() instead
direnv: export +VIRTUAL_ENV -BASH_COMMAND ~LINENO ~PATH
```

PR #1176, opened 2023-10-04
stdlib: update layout_python to resolve deprecation warning

Scribbd:
Since python 3.12 was released, `layout python` produces the following warning: `<string>:1: DeprecationWarning: 'pkgutil.find_loader' is deprecated and slated for removal in Python 3.14; use importlib.util.find_spec() instead`. This commit resolves the warning by using the suggested alternative.

Issues: This will break `layout python2` and `layout python3` for versions before 3.4.

PR #1178, opened 2023-10-04
Use importlib.util.find_spec during python version discovery to fix DeprecationWarning

SinclaM:
This PR addresses #1177 by using `importlib.util.find_spec` during Python version discovery instead of `pkgutil.find_loader`, which has been deprecated in Python 3.12.

This removes a DeprecationWarning that appears when entering a Python environment while using 3.12.


SinclaM:
Ah, never mind. #1176 beat me to it.

PR #1185, opened 2023-10-17
Update direnv.1.md

jwilkins:
add XDG_DATA_HOME default location


man/direnv.1.md
```diff
@@ -157,6 +157,9 @@ ENVIRONMENT
 `XDG_CONFIG_HOME`
 : Defaults to `$HOME/.config`.
 
+`XDG_CONFIG_DATA`
```
Mic92, line 160: I don't see this being used by direnv.

zimbatm: fixed, it was XDG_DATA_HOME

PR #1206, opened 2023-12-04
Update installation.md for Gentoo

hansfn:
Use the Guru project which is official and well maintained.

- [go overlay](https://github.com/Dr-Terrible/go-overlay/tree/master/dev-util/direnv) - version 2.21.2
- [Guru](https://gitweb.gentoo.org/repo/proj/guru.git/tree/app-shells/direnv) - version 2.32.3


zimbatm:
actually his last contrib was 3 years ago so this seems safe to merge

PR #1209, opened 2023-12-05
feat: allow to disable warn timeouts

### LINKED ISSUE #1084: Avoid "is taking a while to execute. Use CTRL-C to give up" warning for commands that are not supposed to finish
I need to run separate shell session (I'm using `nix develop`) command for this. However, direnv complains for these with

`direnv: ([/nix/store/bav6wajkryhkvzkpncisr6xfbd4qvw1j-direnv-2.32.2/bin/direnv export zsh]) is taking a while to execute. Use CTRL-C to give up.
`

Is there a way to disable this waiting in direnv? Because really, after cd to directory, I expect to enter this bash session and stay inside indefinitely.

zimbatm:
Set zero or less to disable warn timeouts.

Fixes #1084

### DIFF
--- a/man/direnv.toml.1.md
+++ b/man/direnv.toml.1.md
@@ -60,6 +60,9 @@ A duration string is a possibly signed sequence of decimal numbers, each with
 optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
 Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 
+This feature is disabled if the duration is lower or equal to zero.
+Will be overwritten if the environment variable `DIRENV_WARN_TIMEOUT` is set to any of the above values.
+
 ## [whitelist]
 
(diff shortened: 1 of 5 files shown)


man/direnv.toml.1.md
```diff
@@ -60,6 +60,8 @@ A duration string is a possibly signed sequence of decimal numbers, each with
 optional fraction and a unit suffix, such as "300ms", "-1.5h" or "2h45m".
 Valid time units are "ns", "us" (or "µs"), "ms", "s", "m", "h".
 
+This feature is disabled if the duration is lower or equal to zero.
+
```
f1ko, line 64: I'd suggest mentioning that the global option can be overwritten:
```suggestion
Will be overwritten if the environment variable `DIRENV_WARN_TIMEOUT` is set to any of the above values.

```

f1ko (commented):
Thanks for picking it up!

Minor suggestion for the documentation.

PR #1221, opened 2023-12-24
Update stdlib.sh to avoid deprecation warning

nmutalik:
Before this, layout python or layout pyenv would always output the following:
`DeprecationWarning: 'pkgutil.find_loader' is deprecated and slated for removal in Python 3.14; use importlib.util.find_spec() instead`

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -821,7 +821,7 @@ layout_python() {
   else
     local python_version ve
     # shellcheck disable=SC2046
-    read -r python_version ve <<<$($python -c "import pkgutil as u, platform as p;ve='venv' if u.find_loader('venv') else ('virtualenv' if u.find_loader('virtualenv') else '');print('.'.join(p.python_version_tuple()[:2])+' '+ve)")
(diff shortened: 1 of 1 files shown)


dipensompurasr:
This is breaking layout_python2 as python2 do not have importlib.util. Resulting following error.

```[1] $ echo "layout python2" > .envrc
direnv: loading ~/temp/py2/.envrc
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ImportError: No module named util
direnv: Could not find python's version
direnv: export ~XPC_SERVICE_NAME```

slinkp:
`layout python2` breakage also reported as this bug:
https://github.com/direnv/direnv/issues/1258

zimbatm:
Please send a PR. I'm not using python enough to know what is the right thing to do here.

dipensompurasr:
> Please send a PR. I'm not using python enough to know what is the right thing to do here.

https://github.com/direnv/direnv/pull/1309

PR #1229, opened 2024-01-15
Added `export` function in `direnv help`

Maxim-Durand:
This PR comes from https://github.com/direnv/direnv/pull/1226#issuecomment-1891682072

Moving away from `direnv export` as a private function means we should document it in `direnv help`.


zimbatm:
thanks!

Issue #68, opened 2013-07-27
How to turn down verbosity?


xpe:
I'd rather not see log messages when I change directories and the environment is adjusted. How can I turn that off?

avinassh:
Can anyone tell me how do I hide messages/logs? The env variable `$DIRENV_LOG_FORMAT` is already empty in my case.

jasoncodes:
I ended up [wrapping the hook function](https://github.com/jasoncodes/dotfiles/blob/9eaee03e84051929cf0a811e7c358b98107be106/shell/config/direnv.sh#L9) to silence the output I didn’t care about.

avinassh:
Where I should put this `direnv.sh` file? 

also, I use zshell, so I think I can copy the contents of [hook function](https://github.com/jasoncodes/dotfiles/blob/master/shell/config/direnv.sh#L4,L14) in `~/.zshrc`. Let me try that.

jasoncodes:
Aside: Take a look at http://github.com/freshshell/fresh if you would like to avoid having to copy shell config snippets around.

avinassh:
That looks great! I will try that. Hope it works now. 

For my earlier query, I appended following to my `~/.zshrc`. 

```shell
# for direnv (check direnv.net)
if which direnv &> /dev/null; then
  eval "$(direnv hook zsh)"

  _direnv_hook() {
    eval "$(
      (
        (
          direnv export "$SHELL" 3>&1 1>&2 2>&3 3>&- |
            egrep -v -e '^direnv: (loading|export|unloading)'
        ) 3>&1 1>&2 2>&3 3>&-
      )
    )"
  };
fi
```

But now when I cd into the directory, it throws everything on the screen:

```shell
$cd work
export DIRENV_MTIME=1434946587;export TESTME=$'avinassh@avinassh.com';export DIRENV_DIFF=$'eJyqVipQsqqu1VHKA1JKLp5Brn5h8UBKyUpJVz-0OLWoWD-xLFPfPLlYSQcm7Rvi6esKVGBoYmxiaWJmamEOlHP3DIl3DA3x8A-Kd_V19PQBygM15iUWZzgANZeUpmTm6yXn50JVOvv7-nqGhLgSUFxbCwgAAP__jmEwfQ==';export DIRENV_DIR=$'-/Users/avi/work';
```

jasoncodes:
[...]

PR #1231, opened 2024-01-19
Add new env var DIRENV_LOG_VERBOSE to configure logging of export diff

hlubek:
See #68 .

This change adds a new env var that can be set to `DIRENV_LOG_VERBOSE=0` to make the logging of exported variables much quieter (single line).

If logging should be completely disabled, the existing var can be set to `DIRENV_LOG_FORMAT=`, but this will disable all logging.

### DIFF
--- a/internal/cmd/cmd_export.go
+++ b/internal/cmd/cmd_export.go
@@ -100,8 +100,12 @@ func exportCommand(currentEnv Env, args []string, config *Config) (err error) {
 		}
 	}
 
-	if out := diffStatus(previousEnv.Diff(newEnv)); out != "" {
-		logStatus(currentEnv, "export %s", out)
+	if out, count := diffStatus(previousEnv.Diff(newEnv)); out != "" {
+		if format, ok := currentEnv["DIRENV_LOG_VERBOSE"]; ok && format == "0" {
+			logStatus(currentEnv, "export %d environment variables", count)
+		} else {
+			logStatus(currentEnv, "export %s", out)
+		}
 	}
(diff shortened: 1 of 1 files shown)


haru02w (approved):

jankatins:
https://github.com/direnv/direnv/pull/1234 has been merged which can be used to hide the diff completely (but no counts like this PR)

zimbatm (changes requested):
Can you move the config to the config file? I'm trying to move away from using env vars to configure direnv.

Otherwise looks good on principle.

sudoforge:
> Can you move the config to the config file? I'm trying to move away from using env vars to configure direnv.
> 
> Otherwise looks good on principle.

actually, i'd prefer to have this be configurable in a specific `.envrc` (using `use_flake` which outputs quite a few lines of env changes). #1234 doesn't provide for that use case, but this does.

kamushadenes:
Hello, can this be merged?

hlubek:
Hi there, I just noticed this open PR and merged the changes from master. I would suggest to show the count if the diff is hidden - this only adds one line to the output. I initially thought about adding a new config field `show_diff_count`, but that adds unnecessary complexity to understand two different boolean flags (where not every combination makes sense).

sudoforge:
i strongly disagree with showing output if the hidden option is enabled  the point of hiding it is to prevent showing any output.

hlubek:
@sudoforge Would you rather introduce a new config field to show the count of exported variables?

PR #1234, opened 2024-02-01
feat: hide env diff

zimbatm:
Some users don't want to see what environment variables have changed in
the shell.

Fixes part of #68


jankatins:
https://github.com/direnv/direnv/pull/1231 has a similar approach but prints a count.

PR #1235, opened 2024-02-02
Add dotenv shell expansion

### LINKED ISSUE #1051: Cannot reference PWD in .env file for variable expansion

joeriddles:
fixes #1051

- Add shell expansion as a last resort when using `dotenv` command in `.envrc` file

### DIFF
--- a/pkg/dotenv/parse.go
+++ b/pkg/dotenv/parse.go
@@ -7,6 +7,7 @@ package dotenv
 import (
 	"fmt"
 	"os"
+	"os/exec"
 	"regexp"
 	"strings"
 )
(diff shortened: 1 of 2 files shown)


reegnz:
I don't think running arbitrary shell commands in a .env file is acceptable. It would defeat direnv's protections, as it doesn't hash .env files, only .envrc files.
If we'd allow arbitrary shell expansion, you'd add a language feature that's neither shell script (so not easily source-able), nor would it be compatible with existing dotenv libraries in various languages.

joeriddles:
Great points @reegnz, thank for the input! I think you've raised some great points (and you already have a PR up for this issue). I'm going to go ahead and close this PR and read some more on direnv's built-in security measures.

PR #1242, opened 2024-02-27
Add Murex support

tiymat:
Added support for the [Murex](https://github.com/lmorg/murex) shell, and implemented `direnv-test-common.sh` as a Murex script.

### DIFF
--- a/GNUmakefile
+++ b/GNUmakefile
@@ -104,7 +104,8 @@ tests = \
 				test-fish \
 				test-tcsh \
 				test-zsh \
-				test-pwsh
+				test-pwsh \
+				test-mx
 
 # Skip few checks for IBM Z mainframe's z/OS aka OS/390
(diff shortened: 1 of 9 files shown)


tiymat:
Hmm, looking at the logs, the nix test seems to have failed because the version of Murex used by the test doesn't support the `$ARGV` variable, which I believe was introduced in version 5.0 this past summer.

I could rewrite the test to avoid its use but it wouldn't matter, as there's a Murex issue with environment variables that was fixed more recently than that release which is needed to support direnv. I mentioned this to a friend more knowledgeable than me about Nix who made a pull request to update the flake.lock, but apparently that PR is running into issues too.

zimbatm (approved):
looks good, thanks! I just fixed the other PR and then rebased this one and things seem to be working.

PR #1243, opened 2024-02-27
Update nixpkgs

joshahussey:
Update flake lock to latest nixpkgs to access new murex version


joshahussey:
Not sure what to do about the check failure. Looks like a go file has too many permissions.

Issue #1254, opened 2024-03-12
direnv with powershell fails to export Nix flake variables


BlackHoleFox:
**Describe the bug**
When using `direnv`, Powershell, and Nix flakes on macOS (maybe the first attempt to do so?), a bad syntax Powershell error stops everything from working. 

I'm new to this all I don't know if the fault lies with `direnv` or `nix-direnv` for how environment variable names are being produced (and quoted) but here I am since it looks trivially fixable inside `direnv`.

In a project of mine I have a Nix flake that exports a few packages via `pkgs.mkShell` and nothing else. When using just `nix develop` it all works fine. I then attempted to add `direnv` and `nix-direnv` to it and hit a problem.

`.envrc`:
```
use flake
```

After configuring my shell hook, I found a large error appearing when `cd`ing into the project's folder:
```
direnv: loading ~/x/project/.envrc
direnv: using flake
warning: Git tree '/Users/fox/x/project' is dirty
[...]

PR #1255, opened 2024-03-12
Powershell: Delimit envrionment variable names when setting

BlackHoleFox:
This is the simplest fix for the problem outlined in #1254: It just uses a brace pattern to delimit what the environment variable name key should be regardless of how `key` needs escaped. This lets it determine that the escape quotes are _only_ for key instantiation. 

With this patch `cd`ing in and out of my project directory works exactly like I'd expect and has no errors.

There might be another fix, like stripping unnecessary quoting, but this seems more robust and like it will work in more cases. I don't know how to write a test for this specific issue though since I'm unsure of how to minimize a Nix flake's output into `direnv`'s test case definitions. If regression tests are wanted, help is appreciated please :)

cc @bamsammich as the original PowerShell support implementer 

Closes https://github.com/direnv/direnv/issues/1254

### DIFF
--- a/internal/cmd/shell_pwsh.go
+++ b/internal/cmd/shell_pwsh.go
@@ -58,7 +58,7 @@ func (sh pwsh) export(key, value string) string {
 	if !regexp.MustCompile(`'.*'`).MatchString(value) {
 		value = fmt.Sprintf("'%s'", value)
 	}
-	return fmt.Sprintf("$env:%s=%s;", sh.escape(key), value)
+	return fmt.Sprintf("${env:%s}=%s;", sh.escape(key), value)
 }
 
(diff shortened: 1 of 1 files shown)


bamsammich (approved):
LGTM

bamsammich:
@BlackHoleFox I thought there was a PR check to run tests, but that doesn't appear to be true. Did you run the PowerShell test suite to confirm they still pass?

bamsammich (changes requested):
After testing this a bit, we didn't get the expected behavior with this change. The problem exists in escaping the key, meaning that this change will set `${env:'key'}=value` literally in the environment, including quotes.

To fix this, you must comb through the escape cases to find the proper method for escaping environment variable names/keys.

Issue #129, opened 2014-06-20
Unload hook


nyarly:
When leaving an .envrc directory, it'd be nice to be able to run a script to undo changes that happened when entering the dir.

For instance, I'm using chruby for use_ruby - so long as I switch from ruby project to ruby project, it's fine, but _leaving_ direnv dirs, I'm left with "whichever ruby I had last" - being able to `chruby 2.0` whenever direnv unloads would be handy.

sknop:
Resetting the shell title is exactly the use I am looking for.

zimbatm:
Hey @sknop, I think you can achieve what you want by changing the PS1 no ?

zimbatm:
not tested but something like that:
```sh
in_direnv() {
  if [ -n "$DIRENV_DIR" ]; then
    echo -n -e "\033];In DirEnv\007"
  else
    echo -n -e "\033];Out of DirEnv\007"
  fi
}
PS1='\$(in_direnv) >'
```

sknop:
Ah, did not think of using the DIRENV_DIR variable, just discovered that. Not a bad idea, I'll try that.

Thanks

zimbatm:
Just fixed the script, it's working now

mattford63:
I have many dirs each has a code base for a different customer. I can launch the appropriate VPN/sshuttle session on entering the dir.  But I can't unload it again :-(

zimbatm:
Hi @mattford63, what would happen if you had two terminals that `cd` into the same customer folder?

My main issue with implementing the unload hook is that it supposes some sort of system-wide state. The second issue is, what would happen if the terminal is closed? I think this can be fixed but it would be fairly easy to leave the system in an inconsistent state.

ChallahuAkbar:
I've got a case where this would be useful as well. In order to work on a Rails project that I'm part of, I need to run `bundle exec rake sunspot:solr:start:` when I enter the folder. However, once I leave that folder I'd like to `bundle exec rake sunspot:solr:stop`, because I don't need to leave the solr server running in the background taking up resources. Of course I could simply try to remember to stop it when I'm done with it, but that would kind of defeat the whole point of using direnv for this in the first place.

zimbatm:
Yeah that would be useful. I haven't figured out how to deal with multiple
shells entering or leaving the folder so if you have an idea let me know.

nyarly:
[...]

PR #1262, opened 2024-04-01
WIP: add unload hook

skogsbrus:
This is a WIP proposal for adding an unload behavior to direnv, as discussed in #129. Mostly intended as a conversation starter on how this should be implemented instead.

This implementation is very hacky:

- Allows circumvention of 'direnv allow', since the unload hook can source another script which can be modified without 'direnv allow' being required to run
- Does not introduce any new abstraction for user-provided hooks; instead relies on the hook command being exported as an env variable and then evaluated
- Defining long unload hooks is cumbersome without sourcing another script
- Seems to also be executed on 'direnv deny', which I guess makes sense

Sample output:

```
johanan@airm2 (add-unload-hook) ~/code/direnv cd foo                                      ~/code/direnv
direnv: loading ~/code/direnv/foo/.envrc
johanan@airm2 (add-unload-hook) ~/code/direnv/foo cat .envrc                          ~/code/direnv/foo
export DIRENV_UNLOAD="echo 'Execute unload hook'"
johanan@airm2 (add-unload-hook) ~/code/direnv/foo cd ..                               ~/code/direnv/foo
direnv: unloading
Execute unload hook
```
.envrc samples:

```
export DIRENV_UNLOAD="source .unloadrc"
```

```
export DIRENV_UNLOAD="echo 'Executing unload hook'"
```


tgrushka:
After 10+ years (#129) and a working PR, the refusal to implement unload hooks feels arbitrary. Load-time commands (e.g., `docker compose up`, `kubectl config set-context --current --namespace="$NAMESPACE"`, etc.) are allowed—why block cleanup? This is a design contradiction, not a technical limitation. A clear 'no' with rationale would be better than silent dismissal.

zimbatm:
I appreciate the sentiment, and wish I had more time to properly review every PR. The issue is more one of available focus than anything else. Something that looks like a simple PR to you, typically has a ramification of consequences for me. If you want to help out maintain direnv with me, I'd be happy for you to take ownership of this feature.

PR #1272, opened 2024-04-15
fix: close tmp file

testwill:


internal/cmd/cmd_fetchurl.go
```diff
@@ -73,7 +73,8 @@ func cmdFetchURL(_ Env, args []string, config *Config) (err error) {
 		return err
 	}
 	defer resp.Body.Close()
-
+	defer tmpfile.Close()
```
zimbatm, line 76: Can you move it next to line 66 so it's close to the init? LGTM otherwise.

zimbatm (approved):
thanks!

PR #1286, opened 2024-06-06
stdlib: export GOBIN for layout_go

nachtjasmin:
This ensures that new binaries are installed into project-specific directories instead of the global one (if set). Since it only affects **the installation of new binaries** (using `go install`), it does not conflict with existing setups.

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -788,7 +788,7 @@ layout() {
 # Usage: layout go
 #
 # Adds "$(direnv_layout_dir)/go" to the GOPATH environment variable.
-# Furthermore "$(direnv_layout_dir/go/bin" is set as the value for the GOBIN environment variable and added to the PATH environment variable.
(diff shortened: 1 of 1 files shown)


stdlib.sh
```diff
@@ -788,11 +788,13 @@ layout() {
 # Usage: layout go
 #
 # Adds "$(direnv_layout_dir)/go" to the GOPATH environment variable.
-# And also adds "$PWD/bin" to the PATH environment variable.
-#
+# Furthermore "$(direnv_layout_dir/go/bin" is set as the value for the GOBIN environment variable and added to the PATH environment variable.
```
zimbatm, line 791:
```suggestion
# Furthermore "$(direnv_layout_dir)/go/bin" is set as the value for the GOBIN environment variable and added to the PATH environment variable.
```

zimbatm:
thanks!

Issue #1274, opened 2024-04-23
direnv mangles environment variables on windows


sprig:
**Describe the bug**
On windows, direnv (even with an empty .envrc) attempts to upcase environment variables that it shouldn't be touching since they are not modified in the config;
```
T:\projects\test> direnv allow                                                                               
T:\projects\test> direnv export pwsh                                                                   
direnv: loading /t/projects/test/.envrc
direnv: export +COMMONPROGRAMFILES +COMSPEC +PATH +PROGRAMFILES +SYSTEMDRIVE +SYSTEMROOT +WINDIR -ComSpec -CommonProgramFiles -Path -ProgramFiles -SystemDrive -SystemRoot -windir
```

Furthetmore, it appears that the expression that direnv produces for powershell is malformed. Note the double single quotes:
```
Remove-Item -Path 'env:/'SystemRoot'';Remove-Item -Path 'env:/'SystemDrive'';$env:PROGRAMFILES='C:\Program Files';$env:COMMONPROGRAMFILES='C:\Program Files\Common Files';$env:COMSPEC='C:\Windows\system32\cmd.exe';$env:SYSTEMDRIVE='C:';Remove-Item -Path 'env:/'ComSpec'';Remove-Item -Path 'env:/'windir'';Remove-Item -Path 'env:/'CommonProgramFiles'';$env:WINDIR='C:\Windows';
...
```
And it produces the following errors, as in #1214, either when hooking into the shell or trying to source the output of `direnv export pwsh`:
```
Remove-Item: A positional parameter cannot be found that accepts argument 'SystemDrive'.
Remove-Item: A positional parameter cannot be found that accepts argument 'Path'.
Remove-Item: A positional parameter cannot be found that accepts argument 'ProgramFiles'.
Remove-Item: A positional parameter cannot be found that accepts argument 'CommonProgramFiles'.
Remove-Item: A positional parameter cannot be found that accepts argument 'ComSpec'.
Remove-Item: A positional parameter cannot be found that accepts argument 'SystemRoot'.
Remove-Item: A positional parameter cannot be found that accepts argument 'windir'
```

Isolating just the produced expression for the comspec variable, we get:
```
❯ cat .\comspec
Remove-Item -Path 'env:/'ComSpec''

T:\projects\test
❯ cat .\comspec|Invoke-Expression
Remove-Item: A positional parameter cannot be found that accepts argument 'ComSpec'.
```

**To Reproduce**
[...]

Issue #1026, opened 2022-12-10
Quick question is it possible to support win natively?


RESDXChgfore9hing:
**Is your feature request related to a problem? Please describe.**
if can it would be great.

**Describe the solution you'd like**
natively support the execution in windows.It would be awesome to use this app even on public library or generally cross platform.
since windows is still a dominant os,it would be great to have the flexibility.

**Describe alternatives you've considered**
i remember seeing a online site that says cmd/bat is just a subset of uniix and it has a mapping.
maybe its possible that it the unix specific command can be platform indpendent.

**Additional context**

zimbatm:
It's possible but requires embedding a Bash emulator to avoid issues related to PATH translations. If you search the issue tracker you will find the discussions around it. I have a long-running "gosh" branch that explores doing that.

RESDXChgfore9hing:
i see,but won't it be easier to has a windows flag and just do a custom logic to translate the $PATH to %PATH%
Do we need a hardcoded absolute PATH value? I thought its all runtime resolved?

Like say for my java compile script,.sh vs .bat>
in unix
export PATH=$EXE4J_JAVA_HOME:$PATH
in windows
set PATH=%EXE4J_JAVA_HOME%;%PATH%

______
it dont see that different to me.am I missing something?

RESDXChgfore9hing:
but I think the bigger problem now would be the hooking.I think the direnv is working even in the master branch.
>direnv
direnv v2.32.2
Usage: direnv COMMAND [...ARGS]

__________
it seems to work so I think the problem is with the hooking. maybe this can help(keep here for everyone to refer)
https://github.com/ajeetdsouza/zoxide/issues/218
______
Ohhhh,i think I see what you mean.
You mean the hardcoded path in the .envrc right?(if the user did put any absolute path)
Then you would have 2 case,unix path* and a windows path* required to make the same script cross platform

zimbatm:
The issue is that Mingw and Cygwin translate paths in various places, such as environment variables, and command-line arguments, .... Because direnv is a Go program that invokes Bash, it's hard to get predictable behaviour.

For now, this is not something I would like to pursue in my free time. I would be open to investigating this under contract if a company comes along and needs this feature for their business ( => https://numtide.com/contact ).

RESDXChgfore9hing:
the problem is you are trying to emulate unix* in the very 1st place no?
____________
If you dont emulate.there is not much PATH translation going on.
[...]

Issue #796, opened 2021-04-24
Incorrect path format is exported on Windows 10 with mintty / git bash, breaking the PATH and command resolution


ChristopherSchubert:
Paths exported from direnv are being converted to a Windows format when running in a terminal emulator on Windows.  This breaks the terminal if the PATH is modified - any subsequent command in the terminal can not be found.

This was touched on here: #253 but I am not having the same issues - this is just a formatting issue with the paths.  

**Environment**
 - OS: Windows 10
 - Shell: git bash / mintty
 - Direnv version 2.28

``` bash
$ cd "${target_dir}"           # direnv not enabled yet.
direnv: error C:\Users\Chris\direnv-test\.envrc is blocked. Run `direnv allow` to approve its content
```
``` bash
$ which direnv                 # No spaces in path to direnv executable
/c/Users/Chris/com.appalachia/appa/direnv
```
``` bash
$ cat "${HOME}/.bashrc"        # .bashrc correctly hooking direnv into bash.
eval "$(direnv hook bash)"
```
``` bash
$ echo "${PATH}"               # $PATH looks fine at this point.
[...]

Issue #638, opened 2020-05-15
direnv rewrite?


zimbatm:
Except that then you'd never get it to work properly on Windows, unless you bundled both an msys and cygwin version.  And you'd still have to solve the cross-system environment munging problem: PATH on cygwin, and *all* env vars containing anything resembling paths on msys/git bash.  (Assuming you could link msys or cygwin to Go code at all, which I have yet to find a single example on the internet of anybody successfully doing.)

The status of my previous experiment in getting direnv to work on Windows was that the "pass an extra argument" trick will get it to basically work on cygwin with some adjustment to the tests (and an inability to handle empty strings), but that on msys/git bash that trick breaks horribly.  AFAICT, the *only* way to get uncorrupted vars on msys (and empty vars on Windows, *period*) is to *pipe* the cygwin/msys environment into direnv.  That is, direnv can never trust the contents of the actual process environment outside of its own (non-empty, and obfuscated so msys doesn't munge them) state variables.

Basically you need the shell-side hook to be something like

* call some direnv command that's the first half of export, to see if the dir or watchlist indicate a change
* run, say, `bash -c 'declare -x' | direnv export-part-2 myshelltype` to run the second half of export, parsing the bash declarations to get a valid environment

I mean, technically you could do that by having direnv export simply output that second command when it detects a change, but you still have to have the *parent* shell output the enviroment or it's no good.

That realization is what prompted the, "well hell, might as well just write it in bash to begin with, then it's one executable, portable everywhere, and a lot fewer process swaps, no Go API worries, etc.".  Plus, direnv is already 25% bash by volume.  :wink:  (That's line count; it's >30% by character count.)

But if you prefer to use the more convoluted way to get windows to work, or to give up on windows altogether, those are certainly options that exist.  I'm just looking at it from the POV of, "what value is Go adding here?"  And AFAICT the answer is, "not much".  And for Windows support, it's an active detriment.

There is apparently an msys go compiler based on gccgo, but it [doesn't do cygwin](https://stackoverflow.com/a/61522751).  I also don't know if said compiler actually produces msys-linked programs, or if it's just meant to build mingw programs (which aren't the same thing).  That is, if it builds Go programs that think their target runtime is Windows, then it's useless.  And even if it didn't, then cygwin would still not be doable.

_Originally posted by @pjeby in https://github.com/direnv/direnv/pull/636#issuecomment-627482713_

zimbatm:
[...]

PR #1291, opened 2024-06-11
feat: cygwin and msys2 support (bash)

W1M0R:
This PR adds support for two common Linux-on-Windows environments: Cygwin (Bash) and MSYS2 (Git Bash).

These environments have [lots in common](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/docs/what-is-msys2.md#msys2-vs-cygwin) with a [few notable differences](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/wiki/How-does-MSYS2-differ-from-Cygwin.md#runtime). It is good enough for this PR to see MSYS2 as a [variant of Cygwin](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/wiki/Devtopics.md#merge-with-cygwin), but with different defaults.

The key difference (relevant to this PR) is in the way that these environments [deal with](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/docs/filesystem-paths.md#process-arguments) paths in [process arguments](https://cygwin.com/cygwin-ug-net/using.html#pathnames-win32) and [environment variables](https://cygwin.com/cygwin-ug-net/setup-env.html). And the key similarity is that they both offer the [cygpath](https://cygwin.com/cygwin-ug-net/cygpath.html) utility to handle path conversions manually.

This PR uses the `cygpath` utility to:

1. fix environment variables and process arguments for use with direnv's `stdlib.sh` (Unix-style paths)
1. fix environment variables and process arguments for use with direnv's Go runtime on Windows (Windows-style paths)
1. export environment variables according to the conventions of MSYS2
1. export environment variables according to the conventions of Cygwin
1. export the `PATH` environment variable in Unix-style
1. enhance relevant `stdlib.sh` functions (such as PATH_add) for automatic conversions to Unix-style paths (MSYS2)
1. prevent the Bash export hook from converting environment variables to Windows-style paths (MSYS2)

This PR was tested on Cygwin, MSYS2 "Minimal" (Git Bash), MSYS2 "Standard" (MINGW64) and WSL, with all unit tests passing and all direnv commands working as expected. More specifically, I am satisfied that the following direnv commands work well: hook, allow, deny, reload, export, edit, exec and status. I also tested this on an existing direnv project and am happy with the results. 

The functionality in this PR is only activated in [conventional MSYS2 environments](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/wiki/MSYS2-introduction.md#shells) and conventional Cygwin environments. Other environments revert back to the regular direnv behaviour. Care was taken to reduce the performance impact for non-cygpath environments by doing cheap calculations to skip the cygpath stack as early as possible. Relevant checks are in `stdlib.sh` and in `cygpath.go`. 

@zimbatm Windows has been my primary development environment for a long time and I don't expect that to change anytime soon, so I'd be happy to maintain this contribution. I also spend a lot of time in the Windows Subsystem for Linux (WSL) and use the Nix Package Manager together with direnv whenever I have the opportunity (and have done so for quite a few projects), so I have a relatively good understanding of the typical direnv build and runtime environments, and it is easy enough for me to test (and maintain) this contribution in different contexts.

## Related Issues

This PR will resolve the following issues (when using direnv inside an MSYS2 Git Bash or a Cygwin Bash environment):

1. #253
3. #796 
4. #1274
5. #1026 
11. #343
12. #142
13. #123  

@pjeby also has a PR aimed towards Cygwin support: #630 (also see #638).

## Future Work

1. A future PR could aim to replace all cygpath calls with a native Go implementation based on the [cygpath source code](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/utils/cygpath.cc), the [cygwin_conv_path](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/path.cc#L4120) function and/or the [conv_path_list](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/path.cc#L1598) function. The MSYS2 runtime also has its [own path conversion function](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/msys2_path_conv.h#L144) that can be consulted. However, this could be tricky, and potentially complicate the direnv codebase a lot, and we may still end up needing cygpath as a fallback. It is also worth noting that one of the reasons to drop cygpath is to avoid the performance penalty of calling an external process on Windows, however, in the `stdlib.sh` we would just be replacing calls to cygpath with calls to direnv (unless we want to maintain two native cygpath implementations - one for Bash in the `stdlib.sh` and one for Go).
1. Both [MSYS2](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/docs/ci.md) and [Cygwin](https://github.com/cygwin/cygwin-install-action) provide GitHub actions. It may be a good idea to set something like this up. 
1. An MSYS2-specific build of direnv could be added to the [MSYS2 package registry](https://packages.msys2.org/search?t=binpkg&q=direnv), to allow direnv to be installed on MSYS2 using `pacman -S direnv` (like [fish](https://packages.msys2.org/base/fish) and [make](https://packages.msys2.org/base/make) for example).
1. A Cygwin-specific build of direnv could be uploaded to the [Cygwin package repository](https://cygwin.com/packages.html), to allow direnv to be installed on Cygwin using the [Cygwin setup program](https://sourceware.org/cygwin-apps/setup.html).
1. If other shells (Fish, Zsh, etc) are a thing in MSYS2/Cygwin, then support could be added for them.
1. Minimal changes were made to the `stdlib.sh`. If usage on MSYS2/Cygwin gains traction, it might be helpful to write a stdlib extension with Cygwin/MSYS2 helper functions (e.g. `is_msys2`, `is_cygwin`, `has_cygpath`, etc). If an extension is not desired, then it could be included in the stdlib.
1. This PR adds useful debug messages to identify calls to cygpath from the stdlib and from the Go runtime. If we need to squeeze more performance out of direnv by reducing cygpath calls even further, we can grep on debug messages that contain the word `cygpath` and then run a few scenarios to verify opportunities for optimization.

## Implementation Notes

1. The regular cross-platform build and deploy process of direnv is sufficient for this PR to work. The functionality in this PR is not dependent on direnv being built in an MSYS2/Cygwin environment.
1. Environment variable normalisation is performed during direnv startup, and not during direnv export, to enable the direnv diffing algorithm to work its magic.
1. Cygpath calls are reduced to a minimum by first using basic heuristics to determine whether the path manipulation requires cygpath.
1. Since MSYS2 does automatic path conversions by default for process arguments and environment variables, we only need to fix the way the `PATH` environment is exported (Go runtime) and manipulated (`stdlib.sh`).
1. Since Cygwin does not do automatic path conversions, we need to convert relevant arguments and environment variables from Unix to Windows style (Go runtime). We also fix `PATH` export.
1. To export the automatic environment modifications made by MSYS2, just continue using `eval "$(direnv export bash)"`. To export an environment that looks more like your `.envrc` file, use `eval "$(MSYS2_ENV_CONV_EXCL="*" direnv export bash)"`. The `eval "$(direnv hook bash)"` command performs the export using the latter method, since it provides a familiar default experience to direnv users and is also faster since it skips a chunk of MSYS2's environment processing.
1. The `make test` target in the `GNUmakefile` tests Bash support, but specifically excludes the tests for the other shells supported by direnv, e.g. Elvish, Fish, etc. This is because the conventional MSYS2/Cygwin environments use Bash, and I didn't want to make any guarantees about support for non-conventional environments.
1. The `GNUmakefile` provides a target to setup a build environment using `scoop`, that can be used for building on MSYS2 (Git Bash) and Cygwin. However, using `scoop` is not the conventional approach on those environments. The minimal MSYS2 environment (Git Bash) has no package manager, so using `scoop` is sensible. The regular MSYS2 environment includes the `pacman` package manager for installing packages. The Cygwin environment does not have a package manager, but includes a setup executable that allows you to choose the packages you want in your environment. The difference between these approaches, is that `scoop` will install the native Windows packages as typically made available by the tool authors, whereas MSYS2 `pacman` and Cygwin `setup.exe` will install versions of the packages that are compiled in those environments with specific patches applied to make Windows tooling more `POSIX-y` (Cygwin) or Unix tooling more `Windows-y` (MSYS2).

## Developer Notes

1. [bash.exe is an MSYS2 process](https://github.com/msys2/MSYS2-packages/blob/48e7d579b809371d8ed067fd12065fffe19b5ac1/msys2-runtime/0032-Handle-ORIGINAL_PATH-just-like-PATH.patch#L13) (as opposed to direnv which is a native Windows executable) and expects a [Unix-style](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/cygwin/environ.cc#L324) `PATH` environment variable.
1. The [cygpath.exe](https://github.com/msys2/msys2-runtime/blob/28d69fba269dd4a9f4281f8af7c2775292241e8b/winsup/utils/cygpath.cc) utility is available by default in an [MSYS2](https://github.com/msys2/msys2.github.io/blob/0fd74c553f69b91a2b21bb1c8c262db7f83dc73a/web/docs/filesystem-paths.md#manual-unix--windows-path-conversion) environment and a [Cygwin](https://cygwin.com/cygwin-ug-net/cygpath.html) environment. 
1. Git Bash is a minimal MSYS2 MINGW64 environment that does not have the `pacman` package manager (which is included by default with the standard MSYS2 installation).

## Comments

@zimbatm @tiymat @joshahussey While testing this PR in WSL, I came across the following code (introduced in #1242):

https://github.com/direnv/direnv/blob/ff09cc4cb7b7e400b8baa16a310ef1b175c651ca/test/direnv-test.mx#L290

When running the unit tests, it looks like a silent failure, in that the tests report success, but looking at the output it seems like it actually fails. I suspected that this was a typo and renamed it to `VIRTUAL_ENV`, however, this caused the unit tests to fail hard, so I decided to leave it as is.

@garyo If you are [still interested](https://github.com/direnv/direnv/issues/638#issuecomment-923127120) in using direnv in an MSYS2/Cygwin environment, then this PR can help. It should be as simple as cloning the PR, installing go, and running `make` followed by `make install`.


W1M0R:
The current solution has the following negatives:

1. slow exports (calling out to cygpath.exe)
2. bash hook slow and dirty after every command (i.e .envrc is loaded at every command)

I've changed this PR to a draft, and will try a different approach.

.gitattributes
W1M0R, line 1: Some of the tests require that these files be in LF. On Windows, the default on clone (on my system) is CRLF. This ensures that it uses EOL as expected by the tests.

.gitignore
W1M0R, line 1: On Windows, the result of make build produces a direnv.exe in the root. This ignores that file.

GNUmakefile
W1M0R, line 1: Some make targets were added to facilitate testing.

fmt-go-lint: Some unit tests failed until I forced golangci-lint to --fix the issues it found.

I use MSYSTEM to ensure that a limited set of tests are run. Since I'm only testing on Git Bash, I don't want to make guarantees for other variants.

init-msys2: This helps to setup a semi-reproducible environment for Windows builds. This could potentially be used in CI.

test-msys2-env: Just double checks the dependencies.

test-msys2-bash: Run only the bash tests.

test-msys2-go: Run only the go tests.

test-msys2: Run both the bash and the go tests.

internal/cmd/env.go
W1M0R, line 1: Allow us to process variables before handing them over to the rest of direnv. This ensures more consistent DIRENV_DIFF results.

test/direnv-test-common.sh
W1M0R, line 1: On MSYS2, symlinks are full file copies. And on Windows, the modtime of state-A and state-B and symlink are all the same. I touch the files to ensure that their modtime is changed.

W1M0R:
I'm happy with this PR now. It is now much faster and the hook now works as expected, without re-exporting on every prompt refresh. Pre-processing the env at direnv startup made the diff more consistent than the previous solution (that only processed the env on export).

This version of direnv works well for my projects on Git Bash.

pjeby:
> @pjeby Would you be able to give this implementation a try on your system?

It looks like this is msys specific, and I use Cygwin.  So as long as the PR's changes don't affect Cygwin I should be good.

W1M0R:
Thanks @pjeby. This PR should not activate on Cygwin environments.

From what I gather from these sources (and a few others), it looks like I am correctly distinguishing between Cygwin and MSYS2 environments using the MSYSTEM variable:

1. https://cygwin.com/cygwin-ug-net/setup-env.html
2. https://www.msys2.org/wiki/How-does-MSYS2-differ-from-Cygwin/

W1M0R:
This PR doesn't yet handle direnv exec properly. I'm looking into it.

pjeby:
Now that you've added all this cygpath usage, it might be useful to note that making `direnv export` do a `cygpath -p` on the `PATH` variable before output when running under Cygwin would make this also work for Cygwin.

Doesn't need to be in this PR or anything, but I thought it might be useful to know.  I'm currently using a patched `_direnv_hook` function that does:

```bash
[[ "$PATH" != *'\'* ]] || export PATH=$(/usr/bin/cygpath -p "$PATH")
```

right after running the `eval "$(direnv export bash)"` bit.

So if direnv did the cygpath -p internally when running on Cygwin, it'd be compatible with Cygwin shells, at least to the extent my current setup is.  (AFAIK Cygwin only mangles PATH, not command-line arguments or other environment variables.)

(Of course, this could also be done by changing the output of `direnv hook`, but then it'd need changing for every supported shell.)

W1M0R:
Thanks for the tip! If this PR makes the cut, then I'll see if I can run with the momentum and get some basic Cygwin support into another PR.

My first implementation of this PR involved a similar approach where I resolved the PATH issue directly in the export, but this ended up working against the way direnv performs its environment diffing. This is not that big of an issue when you are only exporting the environment, but it does pose problems when you start using the shell hook, which is probably how most people use direnv anyway.

I'd probably investigate performing the cygpath call earlier in the stack for Cygwin, in a similar way I did for this PR.

internal/cmd/cmd_exec.go
W1M0R, line 1: The `direnv exec` command did not want to work in windows, with the final nail in the coffin that syscall.Exec is not supported by Windows. The execEnv function was added to provide an opportunity to call into the msys2 variant of that functionality.

internal/cmd/look_path.go
W1M0R, line 1: The `lookPath` function is used by `direnv exec`, and is originally modelled after the Unix version of the built-in go `exec.LookPath` function. This function now calls into the msys2 variant, that uses a lookPath that is modelled after the Windows version of `exec.LookPath`.

.gitattributes
zimbatm: Wouldn't it be enough to use `* text=auto eol=lf`?

internal/cmd/msys2.go
```diff
@@ -0,0 +1,348 @@
...
+	if len(os.Getenv("MSYSTEM")) == 0 {
+		// This is Windows, but not an MSYS2 environment.
+		return m
+	}
+
+	if _, err := exec.LookPath("cygpath.exe"); err != nil {
+		// This is an MSYS2 environment...without cygpath. We require cygpath
+		// for some initial setup.
+		return m
+	}
```
zimbatm, line 58: These requirements need to be added to the docs. Otherwise, it's hard for users to know whether direnv is working or not.

internal/cmd/msys2.go
```diff
@@ -0,0 +1,348 @@
...
+	if err != nil {
+		return "", err
+	}
+	return strings.TrimSpace(string(output)), nil
+}
+
+// toWinMixedPath uses cygpath.exe to convert the Unix input path to a Windows
+// Mixed path.
+func (m msys2) toWinMixedPath(input string) (string, error) {
+	return m.execCygpath("-m", input)
```
zimbatm, line 159: How much logic is it to port over if we wanted all the logic embedded in direnv?

I'm asking because Windows is a bit slow in spawning new programs.

test/direnv-test-common.sh
zimbatm: This might have some implications on the direnv reload logic. Can you check if `direnv reload` and `direnv allow` behave as expected?

zimbatm (commented):
Hey @W1M0R, thanks for the contribution.

Past the minor changes, the only blocker is that I need a Windows machine to test this. I will need your help to maintain some of that code. If you are ready to do this, we can proceed.

internal/cmd/msys2.go
pjeby: To implement cygpath, you would at minimum need to be able to parse /etc/fstab, which controls the drive mapping prefix for both Cygwin and msys (unless msys has an API for this that you can call?).  See also [the source code](https://cygwin.com/git/gitweb.cgi?p=newlib-cygwin.git;a=blob;f=winsup/utils/cygpath.cc) for the bit that doesn't involve /etc/fstab.

(Oh, and in order to *find*/etc/fstab you'll probably need to run cygpath first, as there are many possible install locations for msys and cygwin; my msys is in a subdirectory of `~/scoop/apps/`, for example.)

.gitattributes
W1M0R: It would, I'll update it accordingly.

test/direnv-test-common.sh
W1M0R: I'll check and get back to you.

W1M0R:
> Hey @W1M0R, thanks for the contribution.
> 
> Past the minor changes, the only blocker is that I need a Windows machine to test this. I will need your help to maintain some of that code. If you are ready to do this, we can proceed.

Thanks @zimbatm. I'll let you know once I have implemented the changes.

Windows is my primary development environment (has been for a long time and I don't expect that to change anytime soon), so I'd be happy to maintain the relevant portions.

internal/cmd/msys2.go
W1M0R: @zimbatm I agree, the spawning part can be slow. This PR already does some effort to reduce the amount of calls to cygpath. This specific function, for example, is called only during construction of the `msys2` object (which happens once during a single execution of direnv).

But you are right, it can be improved by implementing the relevant parts of cygpath itself, directly in Go. If we don't need a fully equivalent cygpath implementation, then we might get away with a heuristic based solution. Here is an example: since the current implementation only cares to know where `/` and `/tmp` are mapped to, we could potentially resolve those paths without using cygpath, by looking at the `TMP`, `TMPDIR` or `TEMP` environment variable for the `/tmp` value and maybe the `MSYSTEM_PREFIX` value (or something similar) to derive a value for the `/` path.

I don't want to optimize too early though. I'm hoping that user feedback might identify other situations where we might need to use cygpath (or not).

It may even be possible to pull in the actual c source code of cygpath and use it together with CGO for a fully equivalent solution, but this will complicate builds. Maybe one day someone will take a stab at implementing cygpath entirely in Go.

To answer your question more directly: For how this PR uses cygpath, I think we could get away with a very basic version of it implemented in Go using the heuristic method proposed above.

It is also worth mentioning that the `stdlib.sh` already has a reference to cygpath (in the `source_env` function) and uses it when it is available. With that in mind, if it will already be called for every `source_env` invocation on msys2, then the two calls made in this PR are not that much of an overhead.

internal/cmd/msys2.go
W1M0R: No problem. Do you want me to add entries to the `installation.md` file, the `README.md` file and the `direnv.1.md file`?

Just to be clear, this is the standard MSYS2 environment, so user's don't need to do anything to set this up. An msys2 environment would be broken if it didn't have the `MSYSTEM` variable set and `cygpath.exe` available in the path.

test/direnv-test-common.sh
W1M0R: They behave as expected. When I execute those commands, I expect the environment to reload, and in both cases it does.

internal/cmd/msys2.go
W1M0R: @zimbatm I've made a slight optimization: cygpath is now only called once at direnv startup, specifically to resolve the PATH variable. Maybe we can leave the "implement cygpath logic in Go" for another PR?

internal/cmd/msys2.go
W1M0R: I updated the code slightly to log an error as a hint to the user that something seems odd about their MSYS2 environment.

W1M0R:
@pjeby This PR now has support for Cygwin. It turns out there were a few additional environment variables that needed some special handling for direnv to function properly. Will you be able to review the PR to see if there is anything off about the Cygwin support?

@zimbatm I realized there was enough commonality between Cygwin and MSYS2, so I expanded this PR to cover both environments.

pjeby:
So I'm looking over the code, and maybe I'm reading this wrong due to limited context in-diff, but it looks to me at first glance like there's going to be multiple cygpath calls on *every* direnv execution, which is likely to visibly slow down prompts, loading new .bashrc, etc.  From the code, it looks like it's there to make DIRENV_ stuff show up as unix paths in stdlib.sh, but I'm not sure why?  My setup works fine without that.  The only variable I do any conversion on is `PATH` -- and only on detecting a windows-format PATH coming *back* to Cygwin.

Note that Cygwin commands mostly *accept* Windows paths, they just mostly don't manipulate them well.  And I think that expecting Cygwin and MSYS users to take the time to write their .envrc or scripts to use cygpath is a fine tradeoff, especially since it then only happens when an .envrc is being run.

W1M0R:
> So I'm looking over the code, and maybe I'm reading this wrong due to limited context in-diff, but it looks to me at first glance like there's going to be multiple cygpath calls on _every_ direnv execution, which is likely to visibly slow down prompts, loading new .bashrc, etc. From the code, it looks like it's there to make DIRENV_ stuff show up as unix paths in stdlib.sh, but I'm not sure why? My setup works fine without that. The only variable I do any conversion on is `PATH` -- and only on detecting a windows-format PATH coming _back_ to Cygwin.

@pjeby Thanks for going through the code. Yes, every invocation of direnv on Cygwin will invoke a few cygpath calls, for converting key environment variables that come in as Unix paths, to Windows paths. Since direnv is a native Windows executable, the Go runtime expects Windows-style paths. 

An easy way to see the affect of this, is during the LoadConfig phase of direnv. It looks like direnv is succeeding because LoadConfig is designed to "fail" silently.  The unit tests helped in identifying this issue, but I got the clearest view of the problem by enabling debug logging, using a `DIRENV_DEBUG=1 direnv export bash`. The debug messages show that, without my fix, direnv fails to load the config correctly. With my fixes, all of those silent failures go away. Granted, one might still be able to fine tune the process and do some more optimizations. 

Another place where this has an effect is in the diffing algorithm (when watched files are being checked for modtime changes). This becomes clear with you use the direnv hook, not so much if you are manually using direnv export. If Go can't find the files being watched (i.e. if the watched Unix paths are not converted to Windows paths before checking their mod stats), then the diffing algorithm doesn't work.
 
> Note that Cygwin commands mostly _accept_ Windows paths, they just mostly don't manipulate them well. And I think that expecting Cygwin and MSYS users to take the time to write their .envrc or scripts to use cygpath when managing those vars is a fine tradeoff, especially since it then only happens when an .envrc is being run. (Or alternately, only running cygpath when calling an .envrc, since that happens much less often.)

If by this you mean that users should manage the DIRENV_ stuff (using cygpath) themselves, then I'd disagree on that point: user's shouldn't need to fix (or know) about those direnv internals. For path variables in general, this PR does not call cygpath on your behalf in Cygwin (the stdlib.sh unit tests show they are no-ops), it does however call cygpath for MSYS2 systems when manipulating the PATH using the direnv stdlib. Without this feature, I had to modify my existing .envrc files to be MSYS2 aware, but with that feature, I could just use my existing .envrc files - this is a nice user experience.

The cygpath calls inside the stdlib will only be called when the direnv diffing algorithm determines that a .envrc change was made, so I'm not too concerned about the performance impact of that (since that file and its watches don't change that regularly). Although there is probably more room for optimization there (i.e. I see that the stdlib invokes direnv quite a few times for something like `direnv exec`). In my testing, I didn't notice a painful slowdown of direnv.

If you have more time available, I'd appreciate it if you can clone this PR and test it on some of your direnv projects.

pjeby:
I'm a bit confused by some of your statements, see e.g.:

```
$ DIRENV_DEBUG=1 direnv export bash
direnv: export:04:53:06 cmd_export.go:22: start
direnv: export:04:53:06 cmd_export.go:35: loading RCs
direnv: export:04:53:06 cmd_export.go:43: updating RC
direnv: export:update:04:53:06 cmd_export.go:46: Determining action:
direnv: export:update:04:53:06 cmd_export.go:47: toLoad: "C:\\cygwin\\home\\pje\\obp\\.envrc"
direnv: export:update:04:53:06 cmd_export.go:48: loadedRC: &cmd.RC{path:"C:\\cygwin\\home\\pje\\obp\\.envrc", allowPath:"", times:cmd.FileTimes{list:(*[]cmd.FileTime)(0xc0001a0000)}, config:(*cmd.Config)(0xc0000fe000)}
direnv: export:update:04:53:06 file_times.go:192: getLatestStat: C:\cygwin\home\pje\obp\.envrc: Lstat: 1713897990, Stat: 1713897990 -> preferring Stat
direnv: export:update:04:53:06 file_times.go:140: Check: C:\cygwin\home\pje\obp\.envrc: up to date
direnv: export:update:04:53:06 file_times.go:192: getLatestStat: C:\cygwin\home\pje\.local\share\direnv\allow\eac2b675c00aba9923aa30237edfa3b70202f1d78df5398feba517f9b2b5fa62: Lstat: 1713897994, Stat: 1713897994 -> preferring Stat
direnv: export:update:04:53:06 file_times.go:140: Check: C:\cygwin\home\pje\.local\share\direnv\allow\eac2b675c00aba9923aa30237edfa3b70202f1d78df5398feba517f9b2b5fa62: up to date
direnv: export:update:04:53:06 file_times.go:192: getLatestStat: C:\home\pje\obp\.envrc: Lstat: 1713897990, Stat: 1713897990 -> preferring Stat
direnv: export:update:04:53:06 file_times.go:140: Check: C:\home\pje\obp\.envrc: up to date
direnv: export:update:04:53:06 cmd_export.go:60: no update needed

$ set | grep DIRENV
DIRENV_DIFF=eJzsV...,==
DIRENV_DIR='-C:\cygwin\home\pje\obp'
DIRENV_FILE='C:\cygwin\home\pje\obp\.envrc'
DIRENV_WATCHES=eJy...k6
```

This looks like it's working just fine on Cygwin with Windows paths, without needing any cygpath adjustments?  Also, it appears that cygwin's dirname and basename commands work fine on Windows paths, at least for my quick checks.  The above are all using direnv 2.32.3, installed via scoop on Windows, under Cygwin bash.  (i.e., not using this PR.)

To be fair, I'm not making use of the stdlib on my PC, only basic exports, so if there's some issue there I'm not aware.  I'm just not clear on how you're getting unix paths for DIRENV_ stuff in the first place.

Are you sure you're not running in an MSYS environment rather than a Cygwin one?  MSYS is *extremely* aggressive about converting filename arguments and environment variables, where Cygwin is more laid back and only converts PATH by default.

W1M0R:
Thank you for checking! Your feedback definitely indicates that I'm missing something. Let me give it another go.

W1M0R:
I updated the code with more checks to prevent unnecessary calls to cygpath.

> The unit tests helped in identifying this issue, but I got the clearest view of the problem by enabling debug logging, using a DIRENV_DEBUG=1 direnv export bash.

I was wrong about that statement, it was `DIRENV_DEBUG=1 make test` that helped me identify where direnv was tripping up because of unexpected Unix paths provided to the Go runtime. The unit test scripts set up certain variables (that would not be present in your example), like DIRENV_CONFIG, DIRENV_BASH, XDG_DATA_HOME and XDG_CONFIG_HOME, all in Unix format. That breaks direnv, so I have to temporarily convert those variables (if they are present) using cygpath.

With the added checks, cygpath usage is significantly reduced, practically down to the PATH environment variable. Thanks for highlighting the issue.

@pjeby Do you see any other red flags?

pjeby:
I'm mostly worried that you're changing a lot of existing direnv-on-cygwin semantics in ways that aren't sufficiently conservative.  For example, you've got new logic around symlinks that seem they might be based on the assumption that Cygwin doesn't use real symlinks.  But my setup uses `CYGWIN=winsymlinks:nativestrict`, meaning that it does, in fact use real Windows symlinks.

You've also got a lot of code that appears to check for cygpath as an indication of whether you're running on Cygwin/MSYS, but that's really not the way to go about it in stdlib.sh.  `OSTYPE=cygwin` or `OSTYPE=msys` are how you tell if a bash script is being run in the corresponding OS.

So at this point I'm worried that while your tests work in your environment, that even if they were made to work on mine, they don't necessarily match *other* users' environments, and that I may not even be spotting everything that's an issue.

TBH, I ran into similar issues when I was working on this, which is why we ended up splitting the objective into more bite-sized chunks and got some of those in.

What I would strongly suggest at this point is first doing a PR that only addresses *inbound* translation of those variables that direnv *directly uses for its own operations* (locating configuration/etc),  **without affecting .envrc handling** to start with.

Then, once some experience is gained with the limited use of cygpath, and there is feedback available from actual users, the next PR could address dealing with back-translation of specific outputs to be emitted from `direnv export` -- and only from *output* emission, not input translation.

I'm very concerned that the current approach is conflating these two (very different!) requirements in a way that's likely to lead to lots of bugs and corner cases.  direnv should *not* be translating input vars that aren't for its own immediate use, as that runs the risk of making the environment mangling situation *worse*, not better.  Both msys and cygwin have their own translation layers in each direction with Windows, and injecting *another* such layer (that AFAICT might be running twice?) doesn't seem like a great plan.  For *output*, this should be fine, since you are dealing with the variables being sent back and reverting the *final* "unix to windows" translation (if applicable).

It's like there's currently this `A -> B -> A -> B -> OUT` process.  I'm worried we're going to get `A -> B -> B' -> A'' -> B'' -> OUT` (or worse), instead of `A -> B -> IN(direnv only) + -> A -> B ->OUT -> OUT'`.   Does that make sense?

Sorry to be so negative, but I'm trying to do a conservative reading because the diffs don't show a lot of context and it's been years since I last dove deeply enough into direnv's internals to have the full picture in my head.  But from my recollection of my previous attempt the only *clean* way I could see to make this work properly on MSYS at least was to bypass its translation altogether (e.g. by having the .envrc runner output its environment vars to a pipe, and/or having the direnv hook function pipe exports into direnv export directly.)  Cygwin OTOH is in a "it works OKish now, please don't break it" state, provided PATH is fixed in the output of direnv export. (Hence my inclination to only do output translation at the final phase of direnv export.)

I would actually almost suggest that it's cleaner to add the cygpath calls to the output of `direnv hook` because there you can directly check `OSTYPE`, except that's a bash-only variable.  But provided the .envrc wrote its vars as output to a pipe rather than passing them via the environment, that would be the other good place to do it.  Better still: running the .envrc handler *from the hook* would bypass environment translation altogether, and piping the result into direnv for formatting would remove another translation layer, so there'd never be any translation at all.

But in any case, having a PR that only addresses input conversion of direnv-use-only paths that doesn't affect the actual process environment passed down anywhere else would be a *ton* easier to read and verify safety of.  And then doing the other parts as progressive enhancements would be more reviewable as well.

W1M0R:
> I'm mostly worried that you're changing a lot of existing direnv-on-cygwin semantics in ways that aren't sufficiently conservative. For example, you've got new logic around symlinks that seem they might be based on the assumption that Cygwin doesn't use real symlinks. But my setup uses `CYGWIN=winsymlinks:nativestrict`, meaning that it does, in fact use real Windows symlinks.

The current direnv-on-cygwin setup is not workable and should not be the standard to compare with. The unit tests fail and direnv exec is not supported. With this PR, unit tests succeeds and direnv exec works on Cygwin. In terms of the symlinks, the logic is very minimal (one or two lines) and only used in the unit test to deal with stock Cygwin environments (i.e. a CI-friendly environment).

> You've also got a lot of code that appears to check for cygpath as an indication of whether you're running on Cygwin/MSYS, but that's really not the way to go about it in stdlib.sh. `OSTYPE=cygwin` or `OSTYPE=msys` are how you tell if a bash script is being run in the corresponding OS.

It's not that much code, maybe you were looking at an older commit. I also don't check for Cygwin/MSYS2 anymore, since that doesn't matter. What does matter is whether cygpath is available and whether the path you want to convert looks like a Windows path or not.

> So at this point I'm worried that while your tests work in your environment, that even if they were made to work on mine, they don't necessarily match _other_ users' environments, and that I may not even be spotting everything that's an issue.

I'd love to know if the unit tests pass in your environment.
 
> TBH, I ran into similar issues when I was working on this, which is why we ended up splitting the objective into more bite-sized chunks and got some of those in.

I've also wondered how I can simplify this PR and break it into chunks. My goal is to have the unit tests pass on MSYS2/Cygwin, so any bite-sized contribution should satisfy that. If the chunk doesn't enable the tests to pass, then I don't see the value in the contribution. So currently, I think this is the smallest chunk that I could implement to satisfy that goal. I've already removed some stdlib stuff that I added originally, but found is not needed.

> What I would strongly suggest at this point is first doing a PR that only addresses _inbound_ translation of those variables that direnv _directly uses for its own operations_ (locating configuration/etc), **without affecting .envrc handling** to start with.

This alone would not make the unit tests pass.

> Then, once some experience is gained with the limited use of cygpath, and there is feedback available from actual users, the next PR could address dealing with back-translation of specific outputs to be emitted from `direnv export` -- and only from _output_ emission, not input translation.

This PR only exports PATH to the environment.

> I'm very concerned that the current approach is conflating these two (very different!) requirements in a way that's likely to lead to lots of bugs and corner cases. direnv should _not_ be translating input vars that aren't for its own immediate use, as that runs the risk of making the environment mangling situation _worse_, not better. Both msys and cygwin have their own translation layers in each direction with Windows, and injecting _another_ such layer (that AFAICT might be running twice?) doesn't seem like a great plan. For _output_, this should be fine, since you are dealing with the variables being sent back and reverting the _final_ "unix to windows" translation (if applicable).
> 
> It's like there's currently this `A -> B -> A -> B -> OUT` process. I'm worried we're going to get `A -> B -> B' -> A'' -> B'' -> OUT` (or worse), instead of `A -> B -> IN(direnv only) + -> A -> B ->OUT -> OUT'`. Does that make sense?

In this PR, `direnv` only translates input vars when they are for its own immediate use, and only for the duration and purpose of the specific direnv command. Except for PATH, which is always translated and exported.

> Sorry to be so negative, but I'm trying to do a conservative reading because the diffs don't show a lot of context and it's been years since I last dove deeply enough into direnv's internals to have the full picture in my head. But from my recollection of my previous attempt the only _clean_ way I could see to make this work properly on MSYS at least was to bypass its translation altogether (e.g. by having the .envrc runner output its environment vars to a pipe, and/or having the direnv hook function pipe exports into direnv export directly.) Cygwin OTOH is in a "it works OKish now, please don't break it" state, provided PATH is fixed in the output of direnv export. (Hence my inclination to only do output translation at the final phase of direnv export.)

This PR would transform the Cygwin experience from OKish to "as advertised".

> I would actually almost suggest that it's cleaner to add the cygpath calls to the output of `direnv hook` because there you can directly check `OSTYPE`, except that's a bash-only variable. But provided the .envrc wrote its vars as output to a pipe rather than passing them via the environment, that would be the other good place to do it. Better still: running the .envrc handler _from the hook_ would bypass environment translation altogether, and piping the result into direnv for formatting would remove another translation layer, so there'd never be any translation at all.

I considered OSTYPE, but found discussions online that suggested that it is not as a reliable method as one thinks. As indicated earlier, it is no longer necessary to know whether we are using cygpath in an MSYS2 environment or a Cygwin environment, it is good enough to know if cygpath is available and if the path in question is a Windows path or a Unix path.

Adding cygpath calls to the output of `direnv hook` is not enough, since it does not cater for how direnv deals with watched files and environment diffs. In my mind, this PR does the right thing.

> But in any case, having a PR that only addresses input conversion of direnv-use-only paths that doesn't affect the actual process environment passed down anywhere else would be a _ton_ easier to read and verify safety of. And then doing the other parts as progressive enhancements would be more reviewable as well.

This would not allow the unit tests to pass.

pjeby:
I'm not saying that it's not a good idea to do more, ever: I'm saying that this PR is currently hard for me to review and validate without re-reading most of the source of direnv, so breaking the PR into smaller pieces/phases would be immensely helpful in me being able to properly review it.

As it stands, the only level of review I can give right now is to look at stuff and go, "this looks concerning to me", because I don't have enough *contiguous* time available to me to load the entire program structure in my head - I'm able to devote *minutes* to it right now, not *hours*.  If you want me to do an actual review and/or build this patch, it might be several weeks before you get a response.  So I've been doing this by skimming and cross-referencing the PR in the interest of getting you faster/more feedback than, "I'll let you know in a few weeks."

If you'd rather wait for more detailed feedback or go without it from me, that's fine; I just assumed you'd rather have a low-fi review now than a more thorough one at some unspecified future date.  Conversely, if you want a higher fidelity review from me *now*, that's only going to happen if the PR has a lot smaller scope, hence my suggestion to break it up.  (As @zimbatm did with my last attempt at this, for I believe similar reasons.)

W1M0R:
> I'm not saying that it's not a good idea to do more, ever: I'm saying that this PR is currently hard for me to review and validate without re-reading most of the source of direnv, so breaking the PR into smaller pieces/phases would be immensely helpful in me being able to properly review it.

Understood. I was too excited. I think I can submit individual PRs that would introduce things gradually. 
 
> As it stands, the only level of review I can give right now is to look at stuff and go, "this looks concerning to me", because I don't have enough _contiguous_ time available to me to load the entire program structure in my head - I'm able to devote _minutes_ to it right now, not _hours_. If you want me to do an actual review and/or build this patch, it might be several weeks before you get a response. So I've been doing this by skimming and cross-referencing the PR in the interest of getting you faster/more feedback than, "I'll let you know in a few weeks."

And I appreciate your early feedback, it has guided me in several ways already.

> If you'd rather wait for more detailed feedback or go without it from me, that's fine; I just assumed you'd rather have a low-fi review now than a more thorough one at some unspecified future date. Conversely, if you want a higher fidelity review from me _now_, that's only going to happen if the PR has a lot smaller scope, hence my suggestion to break it up. (As @zimbatm did with my last attempt at this, for I believe similar reasons.)

I'm not in a rush. If time is the issue, then take your time, I'm going away for a bit of a holiday anyway and won't be back for a week or so. But if the PR is too big, even if you have enough time, then let me know, and I'll do incremental PRs for things like the gitattributes file and other small stuff.

W1M0R:
Closed because the PR is too big. Another attempt could be made by using some of the ideas in this PR as inspiration, and incrementally introducing capability.

pjeby:
Sorry I haven't gotten back around to this in the intervening time -- I ended up needing to do some job hunting, followed by some settling-in at the new position.

garyo-gpsw:
I really need this -- just need to be able to set $PATH in my .envrc on msys. Is there a fork somewhere, or any workaround?

PR #1298, opened 2024-07-13
feat: add opam support

smorimoto:
This PR adds support for [opam](https://github.com/ocaml/opam).


smorimoto:
direnv is mentioned in the official OCaml documentation, so having a formal way for direnv to layout does not seem like a bad idea. I often have projects that use OCaml with Node.js, and I just wanted to use `layout node` and `layout opam` together, but in the long run it's pretty convenient to be able to maintain changes on the opam side in one place.

stdlib.sh
```diff
@@ -804,6 +804,13 @@ layout_node() {
   PATH_add node_modules/.bin
 }
 
+# Usage: layout opam
+#
+# Sets environment variables from `opam env`.
+layout_opam() {
+  eval "$(opam env)"
```
zimbatm, line 811: What files does opam load to generate that env?

Adding a `watch_file` before the eval is useful, so direnv automatically reloads whenever the opam env changes.

zimbatm:
sounds good!

stdlib.sh
smorimoto: @zimbatm What `opam env` returns is not actually a file, but a set of environment variables!

smorimoto:
Gentle ping

stdlib.sh
zimbatm: Is there a per-project config file that Opam uses to figure out what version of Ocaml to use, for example?

man/direnv-stdlib.1.md
```diff
@@ -241,6 +241,10 @@ Sets the `JULIA_PROJECT` environment variable to the current directory.
 
 Adds "$PWD/node_modules/.bin" to the PATH environment variable.
 
+### `layout opam`
+
+Sets environment variables from `opam env`.
```
zimbatm, line 246: How does it behave when the environment is not initialized? Should this use case be taken care of?

stdlib.sh
zimbatm, line 811:
```suggestion
  eval "$(opam env "$@")"
```
Some users might want to pass extra flags.

stdlib.sh
```diff
@@ -804,6 +804,13 @@ layout_node() {
   PATH_add node_modules/.bin
 }
 
+# Usage: layout opam
+#
+# Sets environment variables from `opam env`.
+layout_opam() {
```
zimbatm, line 810:
```suggestion
layout_opam() {
  export OPAMROOT=$(direnv_layout_dir)/opam
```
Would it make sense to set a per-project opam root maybe?

stdlib.sh
smorimoto: It could be `*.opam` file(s), but there is no equivalent of the `.node-version` file in Node.js ecosystem in the OCaml ecosystem.

man/direnv-stdlib.1.md
smorimoto: If the environment is not initialized, the `opam env` command returns nothing. So, Basically, there should be nothing to take care of.

stdlib.sh
smorimoto: Indeed!

stdlib.sh
smorimoto: I think that should be `OPAMSWITCH`. If you set it, opam env returns an error if the opam switch is not initialized. In a sense, this is expected behavior.

smorimoto:
If opam switch is not initialized, this error will happen, but the OPAMSWITCH variable is exposed, so opam will definitely not point to any different switch and will not cause unexpected behavior.

```
❯ direnv allow

direnv: loading ~/src/github.com/kuis-isle3sw/IoPLMaterials/.envrc
[ERROR] The selected switch /Users/smorimoto/src/github.com/kuis-isle3sw/IoPLMaterials is not installed. Please fix the value of the OPAMSWITCH environment variable, or use the '--switch <name>' flag
direnv: export +OPAMSWITCH
```

smorimoto:
@zimbatm Ready to merge!

smorimoto:
@zimbatm Thank you for your review and the effort you put into creating and maintaining this software!

smorimoto:
@zimbatm Could you create a new release for this PR?

smorimoto:
Gentle reminder

PR #1305, opened 2024-07-18
Update elvish docs

cthulahoops:
The elvish install instruction use a deprecated (since 0.16) location for elvish modules.

I also needed a different command to install the command to generate the man page.


zimbatm (approved):
thanks!

PR #1309, opened 2024-08-06
Fixed broken layout python2 issue with stdlib.sh.

dipensompurasr:
https://github.com/direnv/direnv/pull/1221#issuecomment-2270771884


zimbatm:
thanks! @slinkp @nmutalik mind taking a look?

slinkp:
@zimbatm Confirmed works for me on both python 3 and python 2.  My shell is bash.
Sorry for the delay!

PR #1329, opened 2024-09-27
feat: add uv support

### LINKED ISSUE #1250: official support for uv
I'd like direnv to layout python using astral-sh/uv in an elegant way.

While this works, it breaks the prompt somehow. 
By default, without using direnv, uv uses the name of the base directory for the venv prompt.

``` --.config/direnv/direnvrc

realpath() {
    [[ $1 = /* ]] && echo "$1" || echo "$PWD/${1#./}"
}
layout_python-uv() {
    local python=${1:-python3}
    [[ $# -gt 0 ]] && shift
    unset PYTHONHOME
    else
        local python_version
        python_version=$("$python" -c "import platform; print(platform.python_version())")
        if [[ -z $python_version ]]; then
            log_error "Could not detect Python version"
            return 1
        fi
        VIRTUAL_ENV=$PWD/.direnv/python-venv-$python_version
    fi
    export VIRTUAL_ENV
        log_status "no venv found; creating $VIRTUAL_ENV"
        uv venv "$VIRTUAL_ENV" 
    fi

    PATH="${VIRTUAL_ENV}/bin:${PATH}"
    export PATH
    source "${VIRTUAL_ENV}/bin/activate"
}

```

**Additional context**

blackary:
Based on the two entries in the wiki https://github.com/direnv/direnv/wiki/Python#uv

The venv-only mode is a minimal version.
 
The project-mode (`uvp`) is a bit more heavy, creating a pyproject.toml, but convenient if you're going to use uv to then manage the project (with `uv add package`, etc.)

Fixes #1250


pjz:
As noted in #1250 , uv doesn't pay attention to the `VIRTUAL_ENV` environment var.  You should `export UV_PROJECT_ENVIRONMENT=$VIRTUAL_ENV` to get uv to use the correct venv dir.

blackary:
@pjz Added

MindTooth:
Not to bump, but care to explain the usecase for `uvp`?  I'm just curious. 😄

blackary:
@MindTooth Among other things, `uvp` creates a `pyproject.toml` file which keeps track of anything that you use `uv add` to install while in the directory. `uv` just creates a virtual environment, and nothing else. Personally I'm using `uvp` for most of my projects, so that I get a virtual environment created and activated, but also can easily recreate it later if I need to, since all my uv-installed packages are listed in pyproject.toml.

stdlib.sh
```diff
@@ -1068,6 +1068,71 @@ layout_pyenv() {
...
+#
+# Uses uv to create and a virutal environment, and then activates it.
+# See https://docs.astral.sh/uv/pip/environments/ for more details.
+# If no python version is specified, the default python version is used. uv will
+# also install the requested python version if it is not installed.
+layout_uv() {
+    # optional argument to specify python version
+    local python_version=$1
+    if [[ -d ".venv" ]]; then
+        VIRTUAL_ENV="$(pwd)/.venv"
```
zimbatm, line 1081:
```suggestion
        VIRTUAL_ENV="${PWD}/.venv"
```
Avoids the sub-shell

stdlib.sh
```diff
@@ -1068,6 +1068,71 @@ layout_pyenv() {
...
+    fi
+
+    if [[ -z $VIRTUAL_ENV || ! -d $VIRTUAL_ENV ]]; then
+        log_status "No virtual environment exists. Executing \`uv venv\` to create one."
+        # If there is a python version specified, use it to create the environment
+        if [[ -n $python_version ]]; then
+            uv venv --python "$python_version"
+        else
+            uv venv
+        fi
```
zimbatm, line 1091:
```suggestion
        uv venv "$@"
```
It might be simpler to forward all the arguments to `uv`. This would also allow removing the `layout_uvp` function.

stdlib.sh
```diff
@@ -1068,6 +1068,71 @@ layout_pyenv() {
...
 }
 
+# Usage: layout uv [<python_version>]
+#
+# Uses uv to create and a virutal environment, and then activates it.
+# See https://docs.astral.sh/uv/pip/environments/ for more details.
+# If no python version is specified, the default python version is used. uv will
+# also install the requested python version if it is not installed.
+layout_uv() {
+    # optional argument to specify python version
```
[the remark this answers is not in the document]
blackary: Fixed

stdlib.sh
blackary: That would be simpler, but I like that this method lines up more closely with my previous favorite layout, `layout pyenv` which also takes a python version. In this case, `--python` is the main argument which is actually useful with `uv venv`.

Although I would love to get away from having `layout_uv` vs `layout_uvp`, I'm not sure how to avoid that practically, since the base command for `uv` is `uv venv`, and for `uvp` it's `uv init`. I suppose we could do `uv "$@"`, but I really like how `layout uv` and `layout uvp` just work without any extra args needed. Otherwise, you would always have to do something like `layout uv venv`.

lsmurray:
What happens if the python version changes (e.g. project moves from 3.9 -> 3.10).

I believe the `layout_python` creates multiple venvs (one per python version). The code here won't do anything though - it will continue using the wrong version if a previous venv exists.

blackary:
> What happens if the python version changes (e.g. project moves from 3.9 -> 3.10).
> 
> I believe the `layout_python` creates multiple venvs (one per python version). The code here won't do anything though - it will continue using the wrong version if a previous venv exists.

Good question @lsmurray -- I was just playing with the `uv venv` functionality, and it actually preserves the same venv if the python version doesn't change, and it's so fast I could just change it to do `uv venv` every time. I don't think there's an equivalent easy fix for `uv init`, though.

SamDc73 (approved):

stdlib.sh
```diff
@@ -1068,6 +1068,63 @@ layout_pyenv() {
...
+
+  PATH_add "$VIRTUAL_ENV/bin"
+  export UV_ACTIVE=1
+  export VIRTUAL_ENV
+  export UV_PROJECT_ENVIRONMENT=$VIRTUAL_ENV
+}
+
+# Usage: layout uvp [<python_version>]
+#
+# Uses uv to create a new project and a virutal environment, and then activates it.
```
dpprdan, line 1095: Typo:
```suggestion
# Uses uv to create a new project and a virtual environment, and then activates it.
```

zimbatm (changes requested):
Looks good.

Please add some tests in tests/stdlib.bash so this stays maintainable, and document in man/direnv-stlib.1.md so users can find it. Then it should be ready.

blackary:
@zimbatm I tried to create test and man entries, but I'm not sure if I did them correctly. I especially wasn't sure about the correct way to write the tests. If you have any suggestions, that would be great.

stdlib.sh
```diff
@@ -1068,6 +1068,61 @@ layout_pyenv() {
...
+  local python_version=${1:-}
+  VIRTUAL_ENV="$(PWD)/.venv"
+
+  if [[ -z $VIRTUAL_ENV || ! -d $VIRTUAL_ENV ]]; then
+    log_status "No uv project exists. Executing \`uv init\` to create one."
+    # If there is a python version specified, use it to create the project
+    if [[ -n $python_version ]]; then
+      uv init --no-readme --python "$python_version"
+    else
+      uv init --no-readme
```
dpprdan, line 1109: Out of curiosity: Is there a particular reason to use the `--no-readme` flag? 

I think I'd prefer this to be as close as possible to the tool's (in this case `uv`'s) default behaviour as to a) avoid user's confusion and b) reduce the risk of getting out-of-sync with the tool.

If it's important, `no-readme` could be an optional argument to `layout_uvp`.

dpprdan:
`uv init` creates different file structures for [applications](https://docs.astral.sh/uv/concepts/projects/#applications) and [libraries](https://docs.astral.sh/uv/concepts/projects/#libraries).
So do we need an option to specify that a project is a library here (application is `uv init`'s default)?

blackary:
@dpprdan I added support for additional arguments to uvp after a `--`, what do you think of that pattern?

lsmurray:
@blackary - that pattern makes sense to me - have you tried this in an actual project yet? I tried and every time I open a folder it creates a new `.venv` and wipes out all the dependencies I've previously installed. I think that user experience is pretty bad. I modified the script a bit to only create a new `.venv` if the version changed. I don't think the same pattern works with arbitrary arguments. Maybe you could store the args somewhere and check if the arguments have changed?

Attaching my modified version here in case it is helpful. hopefully the comments are useful

```sh
# Usage: layout uv [<python_version>]
#
# Uses uv to create a virtual environment, and then activates it.
# See https://docs.astral.sh/uv/pip/environments/ for more details.
# If no python version is specified, the default python version is used. uv will
# also install the requested python version if it is not installed.
layout_uv() {
    # passed in python version
    local python_version=${1:-}
    local venv_dir=".venv"
    local venv_cfg="$venv_dir/pyvenv.cfg"
    local recreate_venv=false
    # current python version in venv
    local current_version

    # check if virtual environment directory exists
    if [[ -d $venv_dir ]]; then
        # check if python version is specified and venv config file exists
        if [[ -n $python_version && -f $venv_cfg ]]; then
            # get the current version from the config file
            current_version=$(sed -n 's/^version_info = //p' "$venv_cfg")
            # compare the current version with the passed in version
            if [[ $current_version != $python_version ]]; then
                recreate_venv=true
            fi
        fi
    else
        # if virtual environment directory does not exist, create it
        recreate_venv=true
    fi

    # if the virtual environment needs to be recreated, do so
    if [[ $recreate_venv == true ]]; then
        if [[ -n $python_version ]]; then
            uv venv --python "$python_version"
        else
            uv venv
        fi
    fi

    VIRTUAL_ENV="$(PWD)/$venv_dir"

    PATH_add "$VIRTUAL_ENV/bin"
    export UV_ACTIVE=1
    export VIRTUAL_ENV
    export UV_PROJECT_ENVIRONMENT=$VIRTUAL_ENV
}
```

blackary:
@lsmurray Good catch! I had it working properly before, but was trying to make sure it switched to the correct python version if you changed to a different one. Now I've updated it to use `.python-version` to specify what version was used, and check if needs to re-create it. I also added tests to check for this.

There doesn't seem to be a great way to automatically re-create with a new python version if you use `uv init`, so I've simply left that one with an error message for the user to notify them they need to change it in the appropriate files.

smorimoto:
Any updates?

Jibux:
Is it possible to export a VIRTUAL_ENV_PROMPT env var also?

jabidahscreationssystems (approved):

jankatins:
@blackary:
> Among other things, `uvp` creates a `pyproject.toml` file which keeps track of anything that you use `uv add` to install while in the directory. `uv` just creates a virtual environment, and nothing else. Personally I'm using `uvp` for most of my projects, so that I get a virtual environment created and activated, but also can easily recreate it later if I need to, since all my uv-installed packages are listed in pyproject.toml.

Stupid question: according to your above description, uvp is only useful once: when it creates the pyproject.toml and afterwards it should behave then like `layout uv` wouldn't it? Or is there a usecase where one would setup direnv but not the pyproject.toml?

Personally, both layouts feel specific to some specific usecases and not very generic. My usecase is that pyproject.toml is under version control, so whenever I add my personal envrc file, there is already a pyroject.toml. I would argue if you need to setup projects so often that one needs a uv init in direnv, I would argue it's easier to run an additional uv init than adding 60 lines to everyones direnv stdlib. 

I would actually have expected a layout_uv to only contain 
- the env var exports (to make working with that venv easier) -> but apparently even that is up for debate: https://github.com/direnv/direnv/issues/1250)
- a `uv venv`, if the .venv folder does not exist -> I don't want to try stuff out by installing package into the venv without adding them to pyproject.toml to discover that the next cd into the dir will clear it out 
- Maybe allow setting a default version if the venv does not exist. I don't want it recreating the venv if the version do not match (same reason as above: I don't want a manual `uv venv --clear --python 3.14` get wiped out in the next cd)

untested:

```shell
local python_version=${1:-}
if [ ! -d .venv ] ; then
   if [[ -n $python_version ]]; then
      uv venv --python "$python_version"
    else
      uv venv
    fi
fi
export VIRTUAL_ENV="${PWD}/.venv"
PATH_add "$VIRTUAL_ENV/bin"
export UV_ACTIVE=1
export UV_PROJECT_ENVIRONMENT=$VIRTUAL_ENV
```

(I know others who prefer to get their venv cleaned up, e.g. https://offby1.website/posts/uv-direnv-and-simple-envrc-files.html which uses uv sync to do it -> this might IMO be a second parameter: `layout uv 3.13 sync` but I would not use that and would not use any layout uv which makes it mandatory -> https://github.com/direnv/direnv/pull/1352/files looks very nice here!)

anton-ryzhov:
I'm currently switching some legacy (pre-uv) projects to it and experimented with this change for my setup. Here is my feedback.

I like that current `layout_uv` supports defining python versions — that is useful for legacy projects (that only use `requirements.txt` for example). This also supports [custom](https://github.com/direnv/direnv/issues/1250#issuecomment-2471315999) (os-managed, `pyenv`-installed, side-compiled, …) python instances:
```
layout uv /usr/bin/python3.13
#or
layout uv $HOME/pyenv/verions/python3.13/bin/python
#or
layout uv 3.13.7
#or
layout uv 3
```

Without explicit version definition it (via `uv`) respects other [ways of defining version](https://docs.astral.sh/uv/concepts/python-versions/#project-python-versions) — `pyproject.toml` and `.python-version`. That is good solution for modern non-`uv` projects, like `poetry` based.

But it looks that this flexibility was not designed or tested well. If I define a custom python [version], the current implementation stores that in `.python-version`. This causes issues:
 * this file should only have versions definition, but something like `$HOME/pyenv/verions/python3.13/bin/python` could be written there.
 * this file might be unexpected / unwanted for some projects. It could be then committed there accidentally because it was not expected so it was not gitignored.
 * for projects that already rely on this file and have it committed and shared among developers — it might be unwanted to update it with someone else's local settings.
 * for projects relying on `pyproject.toml` this file is redundant, they should track version in pyproject file instead.

I think `layout_uv` shouldn't interact with `.python-version` file. It should continue passing optional parameter to `uv venv --python $1` to enable all this flexibility and rely on [native uv's resolution](https://docs.astral.sh/uv/concepts/python-versions/#project-python-versions) when no custom version is set.

In this case it might not catch version change in some circumstances, but that's maybe fine — I'd give users more control in these rare cases and let them recreate `venv` manually. To not accidentally wipe any useful changes (like side-installed packages) in the current `venv` folder.

anton-ryzhov:
Hm, if the function would use an [env var](https://github.com/direnv/direnv/pull/1329#discussion_r2619236974) to define version, then we won't need to have it as parameter:
```
layout uv 3.13
```
turns into
```
export UV_PYTHON=3.13
layout uv
```
or
```
UV_PYTHON=3.13 layout uv
```


That frees up arguments to be used for anything else. We can just pass them to uv:
```
uv venv "$@"
```

That enables to pass various useful options that can't be set using env vars:
```
layout uv --prompt "venv name"
layout uv --relocatable
```
and even allows
```
layout uv --python 3.13
```

PR #1336, opened 2024-10-10
feat: logging filter

muhmud:
Allows you to filter the log messages that are outputted.

Should help with [this](https://github.com/direnv/direnv/issues/68)

### DIFF
--- a/internal/cmd/log.go
+++ b/internal/cmd/log.go
@@ -25,9 +25,9 @@ func setupLogging(env Env) {
 
 func logError(c *Config, msg string, a ...interface{}) {
 	if c.LogColor {
-		logMsg(c.LogFormat, msg, a...)
+		logMsg(defaultLogFormat, msg, a...)
 	} else {
-		logMsg(errorColor+c.LogFormat+clearColor, msg, a...)
+		logMsg(errorColor+defaultLogFormat+clearColor, msg, a...)
 	}
 }
 
--- a/internal/cmd/log.go
+++ b/internal/cmd/log.go
@@ -3,19 +3,15 @@ package cmd
 import (
 	"fmt"
 	"log"
-	"os"
-	"regexp"
 )
 
 const (
-	defaultLogFormat        = "direnv: %s"
-	errorLogFormat          = defaultLogFormat
-	errorLogFormatWithColor = "\033[31mdirenv: %s\033[0m"
-	clearColor              = "\033[0m"
+	defaultLogFormat = "direnv: %s"
+	errorColor       = "\033[31m"
+	clearColor       = "\033[0m"
 )
 
 var debugging bool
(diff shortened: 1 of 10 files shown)


internal/cmd/log.go
```diff
@@ -38,7 +39,16 @@ func logStatus(env Env, msg string, a ...interface{}) {
 	if !ok {
 		format = defaultLogFormat
 	}
-	if format != "" {
+	shouldLog := true
+	filter, ok := env["DIRENV_LOG_FILTER"]
+	if ok {
+		filterRegEx, err := regexp.Compile(filter)
```
zimbatm, line 45: If you want to use it on both sides, then this has to use the same regexp engine as bash. I believe it's PCRE.

Alternatively, it would be acceptable to introduce a `direnv log` command to invoke on the stdlib side.

stdlib.sh
```diff
@@ -133,13 +133,13 @@ direnv_layout_dir() {
...
-    # shellcheck disable=SC2059,SC1117
-    printf "${color_normal}${DIRENV_LOG_FORMAT}\n" "$msg" >&2
+  local msg=$* color_normal=''
+  if [[ -t 2 ]]; then
+    color_normal="\e[m"
+  fi
+  local -r direnv_msg=$(direnv log "$msg" 2>&1)
+  if [[ -n "$direnv_msg" ]]; then
+    printf "${color_normal}%s\n" "${direnv_msg}" >&2
   fi
```
zimbatm, line 143: Could you move the rest of the logic so that this function can be reduced to:
```
log_status() {
  direnv log -status -- "$*"
}
```
And port the log_error() too, and then finally, both sides will be fully in sync.

muhmud: Sure, sounds good

muhmud: I've made the changes; the only issue was the shade of red used, as it was different in `stdlib.sh` (`"\e[38;5;1m"`) as compared to the colour in go (`\033[31m`).

I used the colour in go.

zimbatm:
thanks, I went a bit further and moved the config to the config file, but also broke something in the process. I might push that out as a follow-up PR instead, unless you also want to help out with this.

muhmud:
Sure, I don't mind taking a look

muhmud:
I think I fixed the issue: it wasn't applying the default format if there was no TOML config file, yay for tests!

muhmud:
Are we good to merge?

zimbatm:
yup, thank you

safareli:
For future readers: 

if you have output like this
```
direnv: loading ~/dev/my/repo/.envrc
direnv: export +BLA +BLA +BLA +BLA +BLA
```

and want to only see loading  and not all exported env vars which takes lot of real estate  this is what workes:

```toml
[global]
log_filter="^loading"
```

PR #1347, opened 2024-11-09
fix: escape newlines in generated vimscript

### LINKED ISSUE #1346: `direnv export vim` can export invalid code
**Describe the bug**

The `direnv export vim` command can export invalid vimscript, when an environment variable contains newlines, since vimscript strings are not multi-line. Eg.

```vim
call setenv('buildPhase','{ echo "------------------------------------------------------------";
  echo " WARNING: the existence of this path is not guaranteed.";
  echo " It is an internal implementation detail for pkgs.mkShell.";
  echo "------------------------------------------------------------";
  echo;
  # Record all build inputs as runtime dependencies
  export;
} >> "$out"
')
```

**To Reproduce**

1. Add a multiline string as an attribute to the shell derivation
2. Use `direnv export vim`

**Expected behavior**

The newline characters aren't inserted literally, but with the `\n` escape sequence, which results in correct vimscript.

diniamo:
I haven't tested this, but something like this should work.

Fixes #1346

### DIFF
--- a/internal/cmd/shell_vim.go
+++ b/internal/cmd/shell_vim.go
@@ -47,5 +47,5 @@ func (sh vim) escapeKey(str string) string {
 
 // TODO: Make sure this escaping is valid
 func (sh vim) escapeValue(str string) string {
-	return "'" + strings.Replace(str, "'", "''", -1) + "'"
+	return "'" + strings.Replace(strings.Replace(str, "\n", "\\n", -1), "'", "''", -1) + "'"
 }


diniamo:
I have tested this, and it works.

@zimbatm I'd say this is a trivial change, any chance we could get it merged?

internal/cmd/shell_vim.go
```diff
@@ -47,5 +47,9 @@ func (sh vim) escapeKey(str string) string {
 
 // TODO: Make sure this escaping is valid
 func (sh vim) escapeValue(str string) string {
-	return "'" + strings.Replace(str, "'", "''", -1) + "'"
```
zimbatm, line 50: what happened to the -1? I am not familiar with vimscript

diniamo: And I'm not familiar with Go, but from looking at their docs, the Replacer struct replaces all matches by default.
image

zimbatm: thanks

PR #1352, opened 2024-11-13
Implement uv project support as a layout

offbyone:
This is an alternative implementation of #1250 that is approached differently from #1329 which is venv-focused. This PR specifically uses uv's project implementation, not a virtualenv.

I elected in the initial pass at it to assume:

* if you don't already have a project, this should error out
* you want the project's installed binaries on your PATH

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -1068,6 +1068,27 @@ layout_pyenv() {
   [[ -n "$PYENV_VERSION" ]] && export PYENV_VERSION
 }
 
+# Usage: layout uv
+#
+# Enables the uv project layout in the current directory, and syncs
+# the dependencies in the project.
+#
+# This relies on the `uv` command being available in the PATH, and performs a
+# sync on cd because uv is fast enough it's not impactful. It relies on uv's
+# configuration file and environment variables, rather than arguments.
+#
+layout_uv() {
+  uv sync || return 1
+
+  # activate the virtualenv after syncing; this puts the newly-installed
+  # binaries on PATH.
+  venv_path=$(pwd)/.venv
+  if [[ -f $venv_path ]]; then
+    # shellcheck source=/dev/null
+    source "$venv_path/bin/activate"
+  fi
+}
+
 # Usage: layout ruby
 #
 # Sets the GEM_HOME environment variable to "$(direnv_layout_dir)/ruby/RUBY_VERSION".
--- a/test/stdlib.bash
+++ b/test/stdlib.bash
@@ -37,7 +37,7 @@ test_name dotenv
   dotenv .env.non_existing_file && return 1
 
   # Try to source a file that exists
-  echo "export FOO=bar" > .env
+  echo "export FOO=bar" >.env
   dotenv .env
   [[ $FOO = bar ]]
 )
@@ -52,10 +52,10 @@ test_name dotenv_if_exists
   cd "$workdir"
 
   # Try to source a file that doesn't exist - should succeed
-  dotenv_if_exists .env.non_existing_file  || return 1
+  dotenv_if_exists .env.non_existing_file || return 1
 
   # Try to source a file that exists
-  echo "export FOO=bar" > .env
+  echo "export FOO=bar" >.env
   dotenv_if_exists .env
   [[ $FOO = bar ]]
 )
@@ -82,7 +82,7 @@ test_name direnv_apply_dump
   trap cleanup EXIT
 
   load_stdlib
-  FOO=bar direnv dump > "$tmpfile"
+  FOO=bar direnv dump >"$tmpfile"
   direnv_apply_dump "$tmpfile"
   assert_eq "$FOO" bar
 )
@@ -131,12 +131,12 @@ test_name semver_search
   mkdir "$versions/1.6.0"
 
   assert_eq "$(semver_search "$versions" "program-" "1.4.0")" "1.4.0"
-  assert_eq "$(semver_search "$versions" "program-" "1.4")"   "1.4.1"
-  assert_eq "$(semver_search "$versions" "program-" "1")"     "1.5.0"
-  assert_eq "$(semver_search "$versions" "program-" "1.8")"   ""
-  assert_eq "$(semver_search "$versions" "" "1.6")"           "1.6.0"
-  assert_eq "$(semver_search "$versions" "program-" "")"      "1.5.0"
-  assert_eq "$(semver_search "$versions" "" "")"              "1.6.0"
+  assert_eq "$(semver_search "$versions" "program-" "1.4")" "1.4.1"
+  assert_eq "$(semver_search "$versions" "program-" "1")" "1.5.0"
+  assert_eq "$(semver_search "$versions" "program-" "1.8")" ""
+  assert_eq "$(semver_search "$versions" "" "1.6")" "1.6.0"
+  assert_eq "$(semver_search "$versions" "program-" "")" "1.5.0"
+  assert_eq "$(semver_search "$versions" "" "")" "1.6.0"
 )
 
 test_name use_julia
@@ -152,10 +152,13 @@ test_name use_julia
     julia=$JULIA_VERSIONS/$version_prefix$version/bin/julia
     mkdir -p "$(dirname "$julia")"
     echo "#!/usr/bin/env bash
-    echo \"test-julia $version\"" > "$julia"
+    echo \"test-julia $version\"" >"$julia"
     chmod +x "$julia"
     # Locally disable set -u (see https://github.com/direnv/direnv/pull/667)
-    if ! [[ "$(set +u; use julia "$version" 2>&1)" =~ Successfully\ loaded\ test-julia\ $version ]]; then
+    if ! [[ "$(
+      set +u
+      use julia "$version" 2>&1
+    )" =~ Successfully\ loaded\ test-julia\ $version ]]; then
       return 1
     fi
   }
@@ -166,13 +169,13 @@ test_name use_julia
   test_julia "julia-" "1.1"
   # Custom JULIA_VERSION_PREFIX
   JULIA_VERSION_PREFIX="jl-"
-  test_julia "jl-"    "1.2.0"
-  test_julia "jl-"    "1.3"
+  test_julia "jl-" "1.2.0"
+  test_julia "jl-" "1.3"
   # Empty JULIA_VERSION_PREFIX
   # shellcheck disable=SC2034
   JULIA_VERSION_PREFIX=
-  test_julia ""    "1.4.0"
-  test_julia ""    "1.5"
+  test_julia "" "1.4.0"
+  test_julia "" "1.5"
 )
 
 test_name source_env_if_exists
@@ -188,13 +191,13 @@ test_name source_env_if_exists
   source_env_if_exists non_existing_file
 
   # Try to source a file that exists
-  echo "export FOO=bar" > existing_file
+  echo "export FOO=bar" >existing_file
   source_env_if_exists existing_file
   [[ $FOO = bar ]]
 
   # Expect correct path being logged
   export HOME=$workdir
-  output="$(source_env_if_exists existing_file 2>&1 > /dev/null)"
+  output="$(source_env_if_exists existing_file 2>&1 >/dev/null)"
   [[ "${output#*'loading ~/existing_file'}" != "$output" ]]
 )
 
@@ -209,7 +212,7 @@ test_name env_vars_required
   # shellcheck disable=SC2034
   BAR=1
   export BAZ=
-  output="$(env_vars_required BAR BAZ MISSING 2>&1 > /dev/null || echo "--- result: $?")"
+  output="$(env_vars_required BAR BAZ MISSING 2>&1 >/dev/null || echo "--- result: $?")"
 
   [[ "${output#*'--- result: 1'}" != "$output" ]]
   [[ "${output#*'BAR is required'}" != "$output" ]]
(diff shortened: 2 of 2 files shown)


dpprdan:
Would you consider making virtualenv activation optional? 
Currently most users probably prefer activation, so that should probably be the default.

dpprdan:
I'm refering to point 5 here, i.e. activation isn't necessary if you use `uv run` and might (and that's a big might) be even less so in the future.

> Or we could stick with the current activation. I don't see a lot of value in doing a half-layout here.

That's probably also fine. One could just use `uv sync` directly in the `.envrc` and be done with it. Actually that might not even be necessary, since `uv run` updates the environment before running the command.

offbyone:
Activation isn't necessary if you `uv run` but it also doesn't affect it or harm it, but activation _does_ make the direnv integrations in things like emacs et al work seamlessly, so on balance I think I'd prefer to keep it in.

stdlib.sh
```diff
@@ -1068,6 +1068,33 @@ layout_pyenv() {
...
+  # watch the uv configuration file for changes
+  watch_file .python-version
+  watch_file pyproject.toml
+  watch_file uv.lock
+
+  uv sync || return 1
+
+  # activate the virtualenv after syncing; this puts the newly-installed
+  # binaries on PATH.
+  venv_path=$(pwd)/.venv
```
thernstig, line 1090: Could the `.venv` be set via an env var by users? Meaning that you should opt to look for if the env var is set first or not?

stdlib.sh
```diff
@@ -1068,6 +1068,33 @@ layout_pyenv() {
...
+#
+# Enables the uv project layout in the current directory, and syncs
+# the dependencies in the project.
+#
+# This relies on the `uv` command being available in the PATH, and performs a
+# sync on cd because uv is fast enough it's not impactful. It relies on uv's
+# configuration file and environment variables, rather than arguments.
+#
+layout_uv() {
+  # watch the uv configuration file for changes
```
thernstig, line 1081: minor not: Capitalize sentences like done in the rest of the file/repo

stdlib.sh
```diff
@@ -1068,6 +1068,33 @@ layout_pyenv() {
   [[ -n "$PYENV_VERSION" ]] && export PYENV_VERSION
 }
 
+# Usage: layout uv
+#
+# Enables the uv project layout in the current directory, and syncs
+# the dependencies in the project.
+#
+# This relies on the `uv` command being available in the PATH, and performs a
```
thernstig, line 1076: I am not sure addinc `uv sync` is great. It might be fine for smaller repos, but for larger ones it could be annoying.

Personally I have git hooks that one rebase/merge checks the applicable files and only then syncs. This is much more intrusive then when I jump back/forth between folders and it syncs each time.

I think we should keep `layout uv` to a minimum of activating the venv.

dpprdan: > I think we should keep `layout uv` to a minimum of activating the venv.

But call it `use venv` or something like that? Because then it doesn't have any uv-specific features, does it?

stdlib.sh
```diff
@@ -1068,6 +1068,33 @@ layout_pyenv() {
...
+  watch_file uv.lock
+
+  uv sync || return 1
+
+  # activate the virtualenv after syncing; this puts the newly-installed
+  # binaries on PATH.
+  venv_path=$(pwd)/.venv
+  if [[ -f $venv_path ]]; then
+    # shellcheck source=/dev/null
+    source "$venv_path/bin/activate"
```
dpprdan, line 1093: Does this work for all shells and platforms? (e.g. direnv supports Windows, albeit with issues)

stdlib.sh
serl: `UV_PROJECT_ENVIRONMENT` changes the behavior of uv on this matter ([doc](https://docs.astral.sh/uv/concepts/projects/#configuring-the-project-environment-path)).

Something like this seems reasonable:

```suggestion
  venv_path=$(pwd)/${UV_PROJECT_ENVIRONMENT:-.venv}
```

thernstig: https://docs.astral.sh/uv/concepts/projects/#project-environments states:

> If a relative path is provided, it will be resolved relative to the workspace root. If an absolute path is provided, it will be used as-is, i.e. a child directory will not be created for the environment.

So a check should be made if it is a relative or absolute path. The relative can be considered to be the same path where `.envrc` resides is my guess, since we do that for `watch_file` invocations.

stdlib.sh
thernstig: I am not sure I agree. It seems that the Python specific `layout_` has unfortunately gone on a path to do installation/syncing. But if you look at most layouts, they do not do that. It is intrusive in a way that slows down the prompt.

The description of `layout <type>` even to me indicates it should not do things like installations such as with `uv sync`:

> A semantic dispatch used to describe common project layouts.

@zimbatm what is your take on this?

dpprdan: You seem to assume that I am against leaving sync out (but my interpretation might be wrong?). I am not. I am merely saying "don't call it uv if it isn't uv-specific".

stdlib.sh
serl: What about using direnv own [`expand_path`](https://direnv.net/man/direnv-stdlib.1.html#codeexpandpath-ltrelpathgt-ltrelativetogtcode)?

```suggestion
  venv_path="$(expand_path ${UV_PROJECT_ENVIRONMENT:-.venv})"
```

(if `UV_PROJECT_ENVIRONMENT` is an absolute path to start with, this would leave it as-is)

stdlib.sh
thernstig: Your interpretation was correct. I thought you were against leaving sync out.

Since you would/should most likely use `UV_PROJECT_ENVIRONMENT` for the virtual env path it would still be uv specific 🙂

stdlib.sh
thernstig: I think it should work on Linux, Mac and all Unix-like operating systems, which is what direnv supports: https://direnv.net/#prerequisites

Except for our comment previously about using `UV_PROJECT_ENVIRONMENT`.

stdlib.sh
offbyone: Reasonable!

stdlib.sh
offbyone: I think the value of this would be essentially nil without the sync + watch parts; it's just another location for a virtualenv, then, and that's a one-liner in any `.envrc` anyway.

As far as speed goes, `uv` is faster than nearly any other part of my environment setup, regardless of how complex a project I've thrown at it. Obviously I cannot predict how that performance will change over time, but the trends are positive.

I'd say that if syncing isn't desirable here -- I believe it is, but that's definitely up for debate! -- then this is probably not the ideal change to make to the stdlib.

As a personal anecdote, this is how I've been running direnv for a couple of months; what's in this PR is a cleaned up and streamlined version of my own direnv lib for uv Python, and it is *seamless*; I think my prompt generator takes more time than this does.

thernstig: @offbyone it is not fast if you need to download packages etc. faster than anything else surely, but not immediate. It becomes subjective what fast is if it is left there. It is also more intrusive since it does on-disc changes and not just _environment_ changes.

offbyone: Of course. That's why I didn't propose integrating it into the existing `layout python` suite; this is a different layout, with different expectations, that is entirely optional to use. People who _do_ use it will get what they expect, and those who don't won't be affected.

I am proposing this for the stdlib because I think it's a very useful and ergonomic pattern, and I'd like to use it out of the stock direnv, but if it's not the direction you want to take it in, then that's also fine too. 

Is there some kind of documentation change I could add to this that would make the expectations clear in a way that would make merging it reasonable?

thernstig: If a `layout uv` is added, we should strive for a smallest common denominator at least I think. Saying it is optional is not fair, it is optional for you as well in that sense to not use it.

Or rather:
The benefit of not adding it into this layout is that anyone, who want optionality, can add their own `watch_file` and `uv sync` in their own `.envrc`. But if `uv sync` is left in, it is added not optional for others.

I understand you want this from the vantage point of your workflow. I am not sure everyone agrees with that workflow though.

(I am not 100% opposed to having it in there, but I am not sure it is the intention of layouts to do such things from the start).

offbyone: I'm not a nix user, but how is this different from `use flake`? Does that not download / build if it's not already cached?

stdlib.sh
dpprdan: > I think it should work on Linux, Mac and all Unix-like operating systems, which is what direnv supports: [direnv.net#prerequisites](https://direnv.net/#prerequisites)

I've seen that line in the README, but it's 5+ years old and doesn't square with the fact that there have been releases for Windows in the meantime and that direnv is on winget. 🤷🏻‍♂️ 

That said, I didn't manage to get direnv to run on Windows out-of-the-box, so the pragmatic view is probably that it doesn't matter much?

stdlib.sh
```diff
@@ -1068,6 +1068,33 @@ layout_pyenv() {
...
+# sync on cd because uv is fast enough it's not impactful. It relies on uv's
+# configuration file and environment variables, rather than arguments.
+#
+layout_uv() {
+  # Watch the uv configuration file for changes
+  watch_file .python-version
+  watch_file pyproject.toml
+  watch_file uv.lock
+
+  uv sync || return 1
```
serl, line 1084: (I was tinkering with `uv sync --offline`, which could address some concerns about "the network not being as fast as uv". Probably not a good idea)

I think that we could have something more refined here. Here's a scenario:

- `.envrc` like so
```shell
strict_env
export UV_PROJECT_ENVIRONMENT=.fancy-venv
layout uv
```
- some network errors make `uv` fail
- the virtual environment is still created, but it's not activated and the `UV_PROJECT_ENVIRONMENT` is not set (that means that running `uv sync` in the resulting shell would use `.venv` instead of `.fancy-venv`).

I think that this should be a little more forgiving.

What about starting the script (before the `watch_file`) by checking that we could actually do something:

```shell
  if ! has uv; then
    echo "uv is required. Follow this steps to install it: https://docs.astral.sh/uv/getting-started/installation/"
    return 1
  fi
```

And then just assume that the error message is clear enough, but still trying to activate the venv and keeping all the other variables:

```suggestion
  uv sync || true
```

offbyone: Hm. I think I like your idea. I'm going to tweak it a tiny bit, though, to also avoid re-locking (so direnv doesn't ever change project files).

stdlib.sh
```diff
@@ -1068,6 +1068,34 @@ layout_pyenv() {
...
+  # user and the resolution is obvious.
+  uv sync --frozen || true
+
+  # activate the virtualenv after syncing; this puts the newly-installed
+  # binaries on PATH.
+  venv_path="$(expand_path "${UV_PROJECT_ENVIRONMENT:-.venv}")"
+  if [[ -f $venv_path ]]; then
+    # shellcheck source=/dev/null
+    source "$venv_path/bin/activate"
+  fi
```
serl, line 1095: I found a cleaner way to do the venv activation in [the doc](https://direnv.net/man/direnv-stdlib.1.html#codedirenvload-ltcommand-generating-dump-outputgtcode):

```suggestion
  direnv_load uv run --frozen direnv dump
```

Tried locally and it seems to work. It's more compact and it saves us from reimplementing the `UV_PROJECT_ENVIRONMENT` logic, also I guess it addresses the cross-platform concerns.

And btw `uv run` implicitly calls `uv sync`. We can pass `--no-sync` to avoid that, or `--locked`/`--frozen`. So yeah it's a two-liner in the end 😁 

(I didn't check what's the result in case of network/other errors. Maybe it's still better to split the `sync --frozen` and the `run --no-sync` to still have something to work with)

stdlib.sh
```diff
@@ -1068,6 +1068,34 @@ layout_pyenv() {
...
+
+  # Use --frozen so that direnv entry does not change the lock contents. This
+  # can fail if the lock file is badly out of sync, but it'll report that to the
+  # user and the resolution is obvious.
+  uv sync --frozen || true
+
+  # activate the virtualenv after syncing; this puts the newly-installed
+  # binaries on PATH.
+  venv_path="$(expand_path "${UV_PROJECT_ENVIRONMENT:-.venv}")"
+  if [[ -f $venv_path ]]; then
```
ericbn, line 1092: The above will check if `.venv` is a file, which will always be false. I suggest:
```suggestion
  if [[ -e $venv_path ]]; then
```
Also, shouldn't it create the venv using uv if if does not exist?

stdlib.sh
```diff
@@ -1068,6 +1068,34 @@ layout_pyenv() {
...
+# configuration file and environment variables, rather than arguments.
+#
+layout_uv() {
+  # Watch the uv configuration file for changes
+  watch_file .python-version pyproject.toml uv.lock
+
+  # Use --frozen so that direnv entry does not change the lock contents. This
+  # can fail if the lock file is badly out of sync, but it'll report that to the
+  # user and the resolution is obvious.
+  uv sync --frozen || true
```
ericbn, line 1087: This will not work for all cases. I have a monorepo with multiple small python applications, with multiple requirements.txt nested within the monorepo structure. I do the sync using `uv pip sync $(git rev-parse --show-toplevel)/**/requirements*.txt`.

offbyone: I don't expect it to work for cases where you are using a layout that isn't the `uv` workspace layout. That's a different layout.

stdlib.sh
offbyone: What is the goal here? What does this give us over simply activating the virtualenv?

offbyone: There is one difference between the two; `uv` doesn't set `VIRTUAL_ENV_PROMPT` in the same way that the `.venv/bin/activate` does. 

At the end of the day, the difference is slim enough that I'm not super worried about it. If using the `direnv_load` approach is a blocker to merging, then I'll switch to that. Is it?

serl: No idea. I was just reading the doc and it looked like a good idea

CGamesPlay: I notice that the `uv run --no-sync` approach works *even with no venv*, which makes it the purest "just set environment variables" option that I've seen. Specifically, I can use `direnv_load uv run --no-sync direnv dump`, no filesystem modifications are performed, I can run `uv sync` afterwards, and everything seems to work normally.

Konfekt:
How about a comparison to `uvp` in https://github.com/direnv/direnv/pull/1329 that uses uv's project implementation ?

anton-ryzhov:
I'm currently switching some legacy (pre-uv) projects to it and experimented with this change for my setup.

Since UV doesn't enforce its project system and provide other useful layers of operation, and there are many existing projects step-by-step switching to it, I'd want `direnv` to support these setups as well.
Projects like "`requirements.txt` + `uv` as venv" or "`poetry` / `pdm` + `uv` as venv". https://github.com/direnv/direnv/pull/1329 supports these cases with `layout_uv` and "`uv` only" with `layout_uvp`.

By saying that, I want a very generic name `layout_uv` to support low level "`uv` as `venv`" in the first place, and this PR's function to be called somehow differently.

What do you think?

PR #1354, opened 2024-11-16
Add EditorConfig setting for indent size in shell scripts.

gemmaro:


ccoVeille:
Adding this without adding a GitHub action such as editorconfig-checker is a problem, as it means someone can merge invalid .sh scripts, then someone will have to fix it later because their linter will report an issue

PR #1376, opened 2025-01-29
install script permits args now and uses printf

inktrap:
I want to use the installer to run it with ansible's script module. There i can't set an environment variable and I don't like the idea to just take the first thing from $PATH. I like to download install scripts and to have a look what they are doing. And maybe then I add them to an ansible role.

Now one can specify the `$bin_path` as `$1` or ask for `--help` or `-h`. Arguments overwrite environment variables. I checked this with shellcheck and everything was fine. And I checked the documented use case and my added use case.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -7,7 +7,7 @@ set -euo pipefail
 { # Prevent execution if this script was only partially downloaded
 
   log() {
-    echo "[installer] $*" >&2
+    printf "[installer] %s\n" "$*" >&2
   }
 
   die() {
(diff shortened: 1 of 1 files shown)


inktrap:
I realized that I can pass environment variables quite easily … however I also realized that the installer should probably try (in that order):

```
    $XDG_BIN_HOME
    $XDG_DATA_HOME/../bin
    $HOME/.local/bin
```

and that is not compatible to the previous behaviour of specifying `$bin_path`. So … I am going to leave this open, maybe it is still helpful and someone could provide some feedback if rewriting the installer to be more in line with conventions is helpful or not.

install.sh
```diff
@@ -18,15 +18,27 @@ set -euo pipefail
...
+      log "the script failed with error ${ret}.\n" \
+      "To report installation errors, submit an issue to\n" \
+      "https://github.com/direnv/direnv/issues/new/choose"
     fi
     exit "$ret"
   }
   trap at_exit EXIT
 
+  if [ $# -ge 1 ];then
+    if [ "${1}x" == "--helpx" ] || [ "${1}x" == "-hx" ];then
```
Mic92, line 30: It's an historic artifact and shells from this century should no longer need this hack:

```suggestion
    if [ "${1}" == "--help" ] || [ "${1}" == "-h" ]; then
```

install.sh
```diff
@@ -18,15 +18,27 @@ set -euo pipefail
...
 
+  if [ $# -ge 1 ];then
+    if [ "${1}" == "--help" ] || [ "${1}" == "-h" ]; then
+      printf "USAGE: ./install-direnv.sh [bin_path]\n"
+      printf "bin_path=~/.local/bin ./install-direnv.sh\n"
+      exit 0
+    elif [ -d "$(realpath "$1")" ] && [ -w "$(realpath "$1")" ]; then
+      bin_path="$(realpath "$1")"
+    else
+      log "bin_path ${1} does not exist or is not writeable" >&2
```
Mic92, line 37: Log already writes to stderr

```suggestion
      log "bin_path ${1} does not exist or is not writeable"
```

install.sh
```diff
@@ -18,15 +18,27 @@ set -euo pipefail
...
+      "https://github.com/direnv/direnv/issues/new/choose"
     fi
     exit "$ret"
   }
   trap at_exit EXIT
 
+  if [ $# -ge 1 ];then
+    if [ "${1}" == "--help" ] || [ "${1}" == "-h" ]; then
+      printf "USAGE: ./install-direnv.sh [bin_path]\n"
+      printf "bin_path=~/.local/bin ./install-direnv.sh\n"
```
Mic92, line 32: This doesn't:

```suggestion
      printf "USAGE: ./install-direnv.sh [bin_path]\n" >&2
      printf "bin_path=~/.local/bin ./install-direnv.sh\n" >&2
```

install.sh
```diff
@@ -66,9 +78,12 @@ set -euo pipefail
         break
       fi
     done
+    if [[ -z "$bin_path" || -w $bin_path ]]; then
```
Mic92, line 81: I think the `-w` condition is wrong here and shouldn't be needed

We can just do:

```suggestion
    if ! [[ -z "$bin_path"  ]]; then
```

Issue #320, opened 2017-12-17
encrypted variables with ssh-vault


nbari:
Hi, I have the need to export variables that are stored in an encrypted vault with [ssh-vault](https://ssh-vault.com), what I am doing now is to create a file with the variables something like:

    AWS_ACCESS_KEY_ID=foo
    AWS_SECRET_ACCESS_KEY=bar

Later I create the vault 

    cat my-vars | ssh-vault create > my-encypted-vars.ssh

That allows me to add to the repository `my-encrypted-vars.ssh`  in a safer way, later when I want to use them,  to automate the load of the variables, I have this in `.envrc`

    export $(ssh-vault view $HOME/path/to/my-encrpted-vars.ssh)

That works pretty well, but when dealing with multiple stacks/environments forces me to have different files, for example:

     export $(ssh-vault view $HOME/path/to/my-encrpted-vars-PRO.ssh)
  
or for development:

      export $(ssh-vault view $HOME/path/to/my-encrpted-vars-DEV.ssh)

the pattern could be by git branch:

      export $(ssh-vault view $HOME/path/to/my-encrpted-vars-BRANCH.ssh)

Therefore wondering if probably a `layout` on the side of **dirven** could be used to read the input on `yaml` format and based on the branch load the variables.

For example, instead of having to create multiple files with something like:

     AWS_ACCESS_KEY_ID=foo
     AWS_SECRET_ACCESS_KEY=bar

Could be nice to have a single file per project with the following structture:

    master:
       AWS_ACCESS_KEY_ID=foo
       AWS_SECRET_ACCESS_KEY=bar
    develop:
       AWS_ACCESS_KEY_ID=foo
       AWS_SECRET_ACCESS_KEY=bar
    default:
       AWS_ACCESS_KEY_ID=foo
       AWS_SECRET_ACCESS_KEY=bar

This simplifies and makes cleaner the repo since I could have only one single vault and to get all the vars, I could just do:

    $ ssh-vault view $HOME/path/to/my-encrpted-vars.ssh

But the missing part is the integration with **direnv**.  Probably the `yaml` could be parsed and always use the `default` vars but when doing something like:

    $ direnv allow -env "develop"

 reload vars in `.envrc` but only the ones for the `development` environment.

Better would get from the shell (zsh) the name of the existing branch by using the content of the variable `$vcs_info_msg_0_` and based on that load the variables if any, otherwise fallback to default.
 
This is only an idea but probably are betters way to enhance this.

[...]

PR #1378, opened 2025-02-01
Adding dotenv_context to stdlib for multiplexing dotenv loading.

amundhov:
Thought I'd share some functionality I found useful for myself.

I'm also polishing up some bash specific helper functions to cycle through available contexts and updating prompts. Is there a place to collect examples like that as well?

### DIFF
--- a/stdlib.sh
+++ b/stdlib.sh
@@ -264,6 +264,41 @@ dotenv_if_exists() {
   eval "$("$direnv" dotenv bash "$@")"
 }
 
+# Usage: dotenv_context [defaultcontext]
+#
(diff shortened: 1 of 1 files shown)


amundhov:
Before I forget, credit goes to @zimbatm  in [this issue](https://github.com/direnv/direnv/issues/320#issuecomment-388341274) for coming up with the `.envrc.context` idea.

There are also some [supporting bash functions on my master branch](https://github.com/amundhov/direnv/blob/master/contrib/bash/direnv_context_utils.sh) to switch between contexts and display them on the prompt.

Issue #455, opened 2019-02-03
In non-existent directory, reloading shell prompt -> error LoadConfig() Getwd failed: "getwd: no such file or directory"


rlue:
## Issue description

When the current working directory does not exist, loading the shell prompt (_e.g.,_ by hitting `<Enter>`) causes the following error to appear:

```
direnv: error LoadConfig() Getwd failed: "getwd: no such file or directory"
```

### Steps to reproduce

```sh
$ mkdir foo
$ cd foo
$ rmdir $(pwd)
direnv: error LoadConfig() Getwd failed: "getwd: no such file or directory"
```

zimbatm:
what is the behaviour that you would expect?

rlue:
Whoops, good point.

As a general principle, I would expect to see an error message only when a program has failed to perform one of its expected functions. Since I don't expect direnv to do anything in a non-existent directory, I would expect to see no error message at all.

henrebotha:
I agree, direnv shouldn't be doing anything (from a user's point of view) in a directory that doesn't exist.

HaleTom:
I'd suggest:

1. Store the CWD when running the hook (initialize to /?)
1. If the stored CWD directory doesn't exist next time, exit early

xerus2000:
> I'd suggest:
> 
>     1. Store the CWD when running the hook (initialize to /?)
> 
>     2. If the stored CWD directory doesn't exist next time, exit early

No need to store anything. Simply catch this error or check for existence beforehand.

W1M0R:
When I experienced this error, it turned out to be related to this WSL issue: https://github.com/microsoft/WSL/issues/6448

This worked for me:

> The workaround is to leave the directory and re-enter it:
> `cd; cd -`

PR #1395, opened 2025-03-16
Don't give an error when the current directory doesn't exist

jyn514:
This just indicates that there is no configuration for the current directory; it doesn't need to generate logging.

Fixes https://github.com/direnv/direnv/issues/455

### DIFF
--- a/internal/cmd/rc.go
+++ b/internal/cmd/rc.go
@@ -388,6 +388,9 @@ func findEnvUp(searchDir string, loadDotenv bool) (path string) {
 }
 
 func findUp(searchDir string, fileNames ...string) (path string) {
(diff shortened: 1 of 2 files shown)


internal/cmd/rc.go
```diff
@@ -388,6 +388,9 @@ func findEnvUp(searchDir string, loadDotenv bool) (path string) {
 }
 
 func findUp(searchDir string, fileNames ...string) (path string) {
+	if searchDir == "" {
+		return ""
```
jyn514, line 392: i tested and this actually works fine without the explicit return, but i was worried that it would search parent directories or something like that.

PR #1406, opened 2025-03-27
install.sh: fix empty array error

zimbatm:
See https://github.com/direnv/direnv/pull/1337#issuecomment-2756315464


d:
LGTM. Tested on an env with BASH 4

PR #1409, opened 2025-04-02
use_nix: unset TMPDIR variables

Mic92:
direnv uses temp directory variables that are not valid after the exit of nix-shell. nix-shell cleans up these environment variables.

PR #1424, opened 2025-04-24
use_nix: always restore special variables

timbertson:
https://github.com/direnv/direnv/pull/1409 added code to restore special envvars (yay), but it was inside the check which only happens when there are no arguments.

Moving the code block up means this works in cases like `use nix -p go`, not only `use nix`.


timbertson-zd:
Hi @Mic92 , any chance of a release for this one?

timbertson:
Hi @Mic92 , any chance of a release for this one?

PR #1439, opened 2025-05-26
Detect ble.sh and use blehook PRECMD instead of PROMPT_COMMAND if available

DoctorDalek1963:
I encountered a bug with [Starship](https://github.com/starship/starship) where it would load the prompt before direnv, meaning my prompt was wrong whenever I entered or left a directory that had direnv.

This change uses the `blehook` functionality provided by [ble.sh](https://github.com/akinomyoga/ble.sh) (which is also what Starship uses) to run direnv earlier in the sequence. This fixes that bug with Statship.

### DIFF
--- a/internal/cmd/shell_bash.go
+++ b/internal/cmd/shell_bash.go
@@ -15,7 +15,9 @@ _direnv_hook() {
   trap - SIGINT;
   return $previous_exit_status;
 };
-if [[ ";${PROMPT_COMMAND[*]:-};" != *";_direnv_hook;"* ]]; then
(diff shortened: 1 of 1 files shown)


Mic92:
Could this cause blehook to be called twice? How does this hook get added usually?

DoctorDalek1963:
Previously direnv just used Bash's native PROMPT_COMMAND, which doesn't interact directly with blehook PRECMD. So no, blehook wasn't called at all previously and this PR means it gets called exactly once if ble.sh is installed.

PR #1453, opened 2025-07-02
Automate release

Mic92:
Tested in https://github.com/Mic92/direnv/releases/tag/v2.37.0


Mic92:
This change depends on branch protection to be setup correctly. This is a repository setting that I currently don't have access to.

zimbatm (approved):
<3

PR #1476, opened 2025-07-26
Disable status logging with `log_format = "-"`

### LINKED ISSUE #1419: Unexpected output when using `log_format` to disable normal logging
**Describe the bug**
When `log_format = '-'` is used to disable normal logging in direnv.toml, the following output is shown:

```
-
%!!(MISSING)(EXTRA string=loading ~/path/to/.envrc)
-
%!!(string=+UV_ACTIVE ~PATH ~VIRTUAL_ENV)(EXTRA string=export %!s(MISSING))
```

This seems like a bug, especially because the literal `-` is being output.

In this case, my `.envrc` contains:

```
layout uv
```

**To Reproduce**
direnv.toml:

```toml
[global]
log_format = '-'
```

**Expected behavior**
Logging output should be disabled.

**Environment**
 - OS: macOS Sequoia
 - Shell: fish
 - Direnv version 2.36.0

antoineco:
Treat the `-` value of the `log_format` option specially to match the [documented behavior](https://github.com/direnv/direnv/blob/v2.37.1/man/direnv.toml.1.md#log_format).

Fixes #1419

### DIFF
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -155,8 +155,11 @@ func LoadConfig(env Env) (config *Config, err error) {
 		format, ok := env["DIRENV_LOG_FORMAT"]
 		if ok {
 			config.LogFormat = format
-		} else if global.LogFormat != "" {
-			config.LogFormat = global.LogFormat
+		} else if logFmt := global.LogFormat; logFmt != "" {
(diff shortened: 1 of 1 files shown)


svoop:
Great news, thanks a lot, @antoineco !!

PR #1480, opened 2025-08-08
feat: `global.flatSearchDirs` config option

### LINKED ISSUE #1108: Allow external dir mappings
I've recently been contributing to a few open source projects, and I'm repeatedly in the situation that I want to set some environment variables or enter a nix development shell when working on these. Many of these projects don't use `direnv` or have a different config to what I would prefer.

Thus, I need an easy way to load certain env files when entering a folder, without adding any files to that folder.

I've thought about this for a while, and I think the cleanest solution would be to add a section to `direnv.toml` which allows providing additional search paths per directory. This would allow loading configuration per folder without having to add any files to them. If there already is a direnv file in the directory, it could be prioritized treated the same way as direnv files in parent directories, i.e. adding a "virtual" direnv file in the parent directory.

The obvious solution would be to add direnv files to the global gitignore, and then add those files to the folder without running the risk of adding them to git. This doesn't work if the folder already contains a direnv config, and does not feel optimal whenever you do actually want to commit the file.

Additionally, this can lead to issues when the file is removed and re-added in the future, as `git add -f` always needs to be used.

**Additional Context**
I'm open to working on this (or a similar solution) myself, if you would consider including this in the project.

diniamo:
See the `direnv.toml.1.md` diff for how this works.

The intended use case is adding 1 directory to the list, but I made it a list to be more flexible, because I can imagine environments where that is necessary.

I considered the suggested solution in the issue, but in my opinion, that would introduce way too much unnecessary friction for the user. The way I implemented it still (sort of) allows for it though, so I think this should be fine.

Closes #1108

### DIFF
--- a/internal/cmd/cmd_exec.go
+++ b/internal/cmd/cmd_exec.go
@@ -52,7 +52,7 @@ func cmdExecAction(env Env, args []string, config *Config) (err error) {
 	previousEnv.CleanContext()
 
 	// Load the rc
-	if toLoad := findEnvUp(rcPath, config.LoadDotenv); toLoad != "" {
+	if toLoad := findEnv(rcPath, config.LoadDotenv, config.FlatSearchDirs); toLoad != "" {
 		if newEnv, err = config.EnvFromRC(toLoad, previousEnv); err != nil {
 			return
 		}
--- a/internal/cmd/cmd_export.go
+++ b/internal/cmd/cmd_export.go
@@ -45,7 +45,7 @@ func exportCommand(currentEnv Env, args []string, config *Config) (err error) {
 
 	logDebug("loading RCs")
 	loadedRC := config.LoadedRC()
-	toLoad := findEnvUp(config.WorkDir, config.LoadDotenv)
+	toLoad := findEnv(config.WorkDir, config.LoadDotenv, config.FlatSearchDirs)
 
 	if loadedRC == nil && toLoad == "" {
 		return
--- a/internal/cmd/config.go
+++ b/internal/cmd/config.go
@@ -32,6 +32,7 @@ type Config struct {
 	LogFilter       *regexp.Regexp
 	LogColor        bool
 	WarnTimeout     time.Duration
+	FlatSearchDirs  []string
 	WhitelistPrefix []string
 	WhitelistExact  map[string]bool
 }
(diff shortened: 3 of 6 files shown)


Lunderberg:
This change is exactly what I'd been looking for, as it would allow `direnv` to provide a per-directory environment without requiring the configuration to be decentralized.

I'm not familiar with direnv's codebase, but looking through the changes, a couple of things stood out to me:

* A config file located in the `flatSearchDirs` may be forbidden from use.  This seems a bit odd to me, as `flatSearchDirs` would only contain locations enabled by the user, and not `.env` files pulled in from some cloned repo.  Should these locations be trusted by default?
* If `flatSearchDirs` are not trusted by default, the interaction with `direnv allow` may be a bit confusing.  it would need to be executed within a directory that would cause `findEnv` to locate that rcfile, and couldn't be used to specify the `$FLAT_DIR/project.envrc` directly.
* A single `foo.envrc` could be used for multiple different paths.  (e.g. `/path/to/some/project/utils` and `/path/to/some/other/project/utils` would both use `utils.envrc`.)  I like this capability so that the same environment gets used even if a project is relocated to a new location, but I'm wondering if there should als be a way to specify an absolute path.
* The environment variable `DIRENV_DIR` is currently set based on the rcfile's path (`filepath.Dir(rc.path)` in the `Load` method of `rc.go`).  Since the rcfile could be located in the flat search path, this would incorrectly point to the flat search directory, rather than to the directory that had been entered.

diniamo:
1. I get your point, but I'm not so sure about just trusting everything by default. There might be cases (albeit rare) where you *do* add a cloned repo to `flatSearchDirs`. Let's have a maintainer decide this, once we get one on-board.
2. Yeah, I thought about this too, but I couldn't come up with anything better. It could be mentioned in the documentation for `flatSearchDirs`.
3. Also a valid concern, but I couldn't come up with a simple enough way to do that.
4. Didn't know about the variable honestly, I'll fix this.

Lunderberg:
> Let's have a maintainer decide this, once we get one on-board.

Makes sense, and agreed, especially as different projects have different security models.

> Didn't know about the variable honestly, I'll fix this.

To be entirely honest, I stumbled across it by accident this morning when looking up how `direnv allow` would be handled.  :D

discapes:
I would also like a way to centralize env file management, so i can change systems and cleanup git repos without having to think about backing up .env files.

TimJones:
I extensively use direnv for my projects but most are not owned by me and don't want to include `.envrc` in the project tree. This feature would be killer for me, what can we do to help get it merged?

Issue #1452, opened 2025-06-26
`direnv help` does not clearly indicate which commands are aliases


iloveitaly:
**Describe the bug**

`direnv help` output is confusing. It looks like `grant` and `permit` are commands without help text, instead of being aliases of `allow`.

**To Reproduce**
`direnv help`

**Expected behavior**

It seems like in most CLI tools aliases are _not_ displayed in the main help text and are instead only displayed in the help for a specific subcommand.

I think removing the aliases from the main help output would be an improvement. I'm happy to submit a PR for this.

**Additional context**
Honestly, I'm not a huge fan of aliases unless they reduce typing. I would vote in favor of deprecating the aliases and eventually removing them.

PR #1481, opened 2025-08-09
make it clear which commands are aliases in help output

RubikCubed:
This is one potential solution to #1452
Rather than displaying aliases as individual commands, it groups them under the appropriate command.
I personally feel like an additional newline between each command could be nicer to read, but didn't want to make extra changes.

Previous output:
```
Available commands
------------------
allow [PATH_TO_RC]:
permit [PATH_TO_RC]:
grant [PATH_TO_RC]:
  Grants direnv permission to load the given .envrc or .env file.
block [PATH_TO_RC]:
deny [PATH_TO_RC]:
disallow [PATH_TO_RC]:
revoke [PATH_TO_RC]:
  Revokes the authorization of a given .envrc or .env file.
```

Adjusted output:
```
Available commands
------------------
allow [PATH_TO_RC]:
  aliases: permit, grant
  Grants direnv permission to load the given .envrc or .env file.
block [PATH_TO_RC]:
  aliases: deny, disallow, revoke
  Revokes the authorization of a given .envrc or .env file.
```

### DIFF
--- a/internal/cmd/cmd_help.go
+++ b/internal/cmd/cmd_help.go
@@ -30,10 +30,14 @@ Available commands
 				}
 			} else {
 				fmt.Printf("%s%s:\n", cmd.Name, opts)
+				var aliases []string
 				for _, alias := range cmd.Aliases {
 					if alias[0:1] != "-" {
-						fmt.Printf("%s%s:\n", alias, opts)
+						aliases = append(aliases, alias)
 					}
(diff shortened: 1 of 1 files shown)


iloveitaly:
I ran into this problem as well, I think it would make things much more clear for users.

Mic92:
Thanks!

PR #1496, opened 2025-09-18
GetEnv: malformed variables trigger panic

gabriel-vasile:
The env variables returned by `os.Environ()` are supposed to follow a "key=value" format. But there are cases when the format is not followed and direnv panics.

The way I found this panic was by running direnv through `os/exec`:
```go
c := exec.Command("direnv")
c.Env = []string{"INVALID_BECAUSE_NO_EQUALS_SIGN"}
c.Run()
```
This commit makes `GetEnv` skip over malformed variables and log their values to stderr.

The root of the problem seems `exec.Command` not doing any validation, but I don't think that justifies a panic in direnv. I'll try to find why `exec.Command` allows invalid variables, but I don't think adding validation will be an easy/acceptable change on golang side.

I'm not sure how I could add a testcase for this change. Let me know where I could add a test.

### DIFF
--- a/internal/cmd/env.go
+++ b/internal/cmd/env.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"encoding/json"
+	"fmt"
 	"os"
 	"strings"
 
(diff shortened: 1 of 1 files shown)


internal/cmd/env.go
```diff
@@ -21,7 +23,10 @@ func GetEnv() Env {
 
 	for _, kv := range os.Environ() {
 		kv2 := strings.SplitN(kv, "=", 2)
-
+		if len(kv2) < 2 {
+			fmt.Fprintf(os.Stderr, "direnv: Skipping invalid environment variable: %s\n", kv)
```
gabriel-vasile, line 27: I'm logging an error, but let me know if there is any better way to handle.

PR #1497, opened 2025-09-30
docs: Add .gitignore tip to README (fixes #1434)

### LINKED ISSUE #1434: Missing documentation about what of direnv .gitignore and what not
I'm slightly confused and am missing documentation about what of direnv .gitignore and what not.

I'm guessing that...

* `.envrc` that typically SHOULD be committed to git
* `.envrc.local` (iff `.envrc` contains `source_env_if_exists .envrc.local`) should be added to `.gitignore`
* `.direnv/` (which contains `flake-profile*`, at least for me, with an `use flake` in `.envrc`) should also be in `.gitignore`

The first two are obvious, it's the last one that I'm not sure about - I don't really understand what `.direnv/` is.

Would you like a PR with an update to README about this?

vorburger:
Fixes #1434.

### DIFF
--- a/README.md
+++ b/README.md
@@ -84,6 +84,9 @@ $ echo ${FOO-nope}
 nope
 ```
 
+We recommend that you commit `.envrc` to your version control system, but add
+the `.envrc.local` file and the `.direnv/` directory e.g. to `.gitignore`.
+
 ### The stdlib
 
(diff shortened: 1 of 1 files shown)


vorburger:
@zimbatm 👋🏽 @Mic92

PR #1507, opened 2025-10-23
fix: stdlib: support named pipes

tvuotila:
Closes #1313

I was trying to get direnv to work with 1Password's new [Access secrets from 1Password through local .env files](https://developer.1password.com/docs/environments/local-env-file) feature. However, the 1Password .env files are named pipes and `dotenv` function does not support named pipes. I already did this change locally by re-refining the functions in `$HOME/.config/direnv/direnvrc` file. I decided to try to upstream this change so that others can also benefit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,4 +1,9 @@
 
+Unreleased
+==================
+
+  * fix: stdlib: dotenv, dotenv_if_exists: support named pipes (#1313)
+
 2.37.1 / 2025-07-20
 ==================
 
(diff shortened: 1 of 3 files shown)


tvuotila:
Named pipe modification time will keep changing. Thus, we cannot actually `direnv watch` them. I will make the necessary changes to the code.

tvuotila:
I am still not happy with this. I should change `direnv watch` to work sanely with named pipes.

tvuotila:
I am not sure how to make this work

PR #1526, opened 2025-12-03
Fix Nushell hook example in docs

ofek:
The previously referenced script was [removed](https://github.com/nushell/nu_scripts/pull/1084) in favor of the [cookbook example](https://www.nushell.sh/cookbook/direnv.html).


ofek:
I updated the branch; let me know what else is required to do here.

ofek:
Can someone please take a look at this?

PR #1536, opened 2025-12-31
fix: find `direnv` from PATH over absolute path in shell installation hook

### LINKED ISSUE #1447: _direnv_hook cannot find the binary when installed in distrobox
Hi,

I wanted to install direnv in a developer distrobox and export it from there to my host machine (as I have done with most clis). The binary is installed in the container, but when exported the binary shows up in `~/.locals/bin/direnv` on my host machine.

Running `direnv --version` works and shows the last version.

However, on each navigation I now get

`_direnv_hook:2: no such file or directory: /usr/bin/direnv`

I have a `direnvrc`-file but it is trivial and the contents is not the problem. It seems as if I have the binary correctly installed but that direnv hook logic fails to find the binary, even though that direnv is in the PATH. Why would it look for the absolute path `/usr/bin/` and not just `direnv` (and expect it to be in PATH)?

**EDIT**
I am using zsh if that matters. Is perhaps the problem to be found here?

```go
const zshHook = `
_direnv_hook() {
  trap -- '' SIGINT
  eval "$("{{.SelfPath}}" export zsh)"
  trap - SIGINT
}
typeset -ag precmd_functions
if (( ! ${precmd_functions[(I)_direnv_hook]} )); then
  precmd_functions=(_direnv_hook $precmd_functions)
fi
typeset -ag chpwd_functions
if (( ! ${chpwd_functions[(I)_direnv_hook]} )); then
  chpwd_functions=(_direnv_hook $chpwd_functions)
fi
```
https://github.com/direnv/direnv/blob/b817913459cd1093f58142755708ec7f4308b362/internal/cmd/shell_zsh.go

and

```go
	selfPath, err := os.Executable()
	if err != nil {
		return err
	}
```
https://github.com/direnv/direnv/blob/b817913459cd1093f58142755708ec7f4308b362/internal/cmd/cmd_hook.go

scarf005:
fixes #1447

### DIFF
--- a/internal/cmd/shell_bash.go
+++ b/internal/cmd/shell_bash.go
@@ -10,7 +10,7 @@ var Bash Shell = bash{}
 const bashHook = `
 _direnv_hook() {
   local previous_exit_status=$?;
-  vars="$("{{.SelfPath}}" export bash)";
+  vars="$(direnv export bash)";
   trap -- '' SIGINT;
   eval "$vars";
   trap - SIGINT;
--- a/internal/cmd/shell_elvish.go
+++ b/internal/cmd/shell_elvish.go
@@ -14,7 +14,7 @@ func (elvish) Hook() (string, error) {
 	return `## hook for direnv
 set @edit:before-readline = $@edit:before-readline {
 	try {
-		var m = [("{{.SelfPath}}" export elvish | from-json)]
+		var m = [("direnv" export elvish | from-json)]
 		if (> (count $m) 0) {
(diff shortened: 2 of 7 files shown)


zimbatm:
The issue is that if direnv is loaded from the PATH, and the PATH gets messed up in an environment, then direnv is not able to unload that environment. That's why we are using a fixed location.

BirdeeHub:
https://github.com/direnv/direnv/pull/1564

^ would this solve this issue?

PR #1556, opened 2026-03-04
feat(stdlib): source_up_secure

### LINKED ISSUE #1328: Feature Request: Consider adding `source_up_secure` to stdlib
This is a followup from https://github.com/direnv/direnv/issues/249.

Basically I think that https://github.com/direnv/direnv/issues/249#issuecomment-1322602805 is a reasonably complete solution to this issue, and is worth incorporating into the standard library. (I've been using it myself for years now without issue.) We may want to use the more broadly compatible version from @mattieb at https://github.com/direnv/direnv/issues/249#issuecomment-1907178014.

I figured I'd open a new issue for further discussion since the other one was closed. 🙂

I may try to open a PR for this, but feeling a tad bit overwhelmed about it, so we'll see.

vergenzt:
Resolves #1328. See that issue (and its linked threads) for rationale. 🙂 

I updated `man/direnv-stdlib.1.md` and then re-ran `make` from within a `nix develop` shell.

### DIFF
--- a/man/direnv-stdlib.1
+++ b/man/direnv-stdlib.1
@@ -119,6 +119,15 @@ found, nothing happens.
 .PP
 NOTE: the other \fB\&.envrc\fR is not checked by the security framework.
 
+.SS \fBsource_up_secure\fR
+Loads a \fB\&.env\fR or \fB\&.envrc\fR file from a parent directory by invoking direnv's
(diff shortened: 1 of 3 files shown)


mattieb:
I’ve been using this exact code in my “.direnvc” for a couple years now; it’s an irreplaceable part of my workflow.

Thanks, @vergenzt, for putting the PR together!

Issue #1548, opened 2026-01-29
does not unload the environment when `.envrc` is deleted


str50:
**Describe the bug**
When a loaded `.envrc` is deleted from the folder, the environment is not unloaded. 

**To Reproduce**
In any folder with a loaded `.envrc` file, delete the `.envrc` file. Observe that the environment is not modified no matter if running `direnv deny` or `cd` out of the folder or `direnv reload`.

**Expected behavior**
`direnv` should unload the environment when the file deletion is detected.

**Environment**
 - OS: macOS Tahoe
 - Shell: fish, zsh
 - Direnv version: 2.37.1

**Additional context**

jgfranco17:
Hello maintainers, I'd like to take a swing at this, if I may? 🙏

PR #1562, opened 2026-03-19
[1548] FIX: Unload environment on delete

jgfranco17:
## Fix: unload environment when `.envrc` is deleted

Addresses https://github.com/direnv/direnv/issues/1548.

When a loaded `.envrc` is deleted, direnv fails to unload the environment on the next prompt eval (or on `cd` out of the directory).

**Root cause:** In `cmd_export.go`, the early-return guard was:

```go
if loadedRC == nil && toLoad == "" {
    return
}
```

Meanwhile `DIRENV_DIFF` is still set in the shell from the previous load, so the exported variables remain in the environment indefinitely.

### Changes

- Extend the early-return guard to also require`DIRENV_DIFF` to be unset

## Testing

- Added test scenario for deleted `.envrc`

### DIFF
--- a/internal/cmd/cmd_export.go
+++ b/internal/cmd/cmd_export.go
@@ -47,7 +47,7 @@ func exportCommand(currentEnv Env, args []string, config *Config) (err error) {
 	loadedRC := config.LoadedRC()
 	toLoad := findEnvUp(config.WorkDir, config.LoadDotenv)
 
-	if loadedRC == nil && toLoad == "" {
+	if loadedRC == nil && toLoad == "" && currentEnv[DIRENV_DIFF] == "" {
 		return
 	}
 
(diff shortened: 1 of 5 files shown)


winterqt:
Hi! I [wrote a fix](https://github.com/winterqt/direnv/commit/c7342200c9726eb3fad98e1ec6b8a5d407393eb8) for this before seeing the issue + this PR.

Yours looks a bit simpler in terms of the actual implementation, but you might want to take my code for the tests, as yours has a chance of leaving the working tree dirty if it fails.

jgfranco17:
> but you might want to take my code for the tests, as yours has a chance of leaving the working tree dirty if it fails.

Hi @winterqt, thanks for the assist! I've updated the test scripts accordingly.

PR #1564, opened 2026-03-22
feat: Add DIRENV_EXE_PATH env variable

zenoli:
Add support for an environment variable `DIRENV_EXE_PATH`, which allows setting the executable path to something other than os.Executable.

The use case is extremely niche:

When creating a [nix-wrapper package](https://www.youtube.com/watch?v=Zzvn9uYjQJY)
 to bundle direnv's configuration (everything within `$XDG_CONFIG_HOME/direnv`) inside a wrapped direnv package, the direnv executable should be the wrapper (a Bash script that injects `DIRENV_CONFIG` and calls direnv with all arguments passed through), rather than the actual direnv executable. Otherwise, the wrapped configuration will not be picked up.

Despite the highly specific nature of this feature, I believe this PR is worthwhile to consider because:

- The feature is completely backward compatible: if `DIRENV_EXE_PATH` is unset, os.Executable is still used.
- It reduces code duplication by injecting Config into CmdHook.
- It is concise and does not introduce any technical debt.

Please let me know if this has any chance of being merged, and I will update the documentation and unit tests accordingly.

### DIFF
--- a/internal/cmd/cmd_hook.go
+++ b/internal/cmd/cmd_hook.go
@@ -28,10 +28,19 @@ func cmdHookAction(_ Env, args []string) (err error) {
 		target = args[1]
 	}
 
-	selfPath, err := os.Executable()
-	if err != nil {
-		return err
+	// Prefer DIRENV_EXE_PATH if set
+	selfPath := os.Getenv("DIRENV_EXE_PATH")
+	if selfPath == "" {
+		selfPath, err = os.Executable()
+		if err != nil {
+			return err
+		}
 	}
+
+	// selfPath, err := os.Executable()
+	// if err != nil {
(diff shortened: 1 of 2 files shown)


BirdeeHub:
Would this be a good solution to https://github.com/direnv/direnv/issues/1447 ?

zenoli:
I think it would solve it too. I cannot verify as I don't use distrobox but I asked the two people having the issue to verify this.

svenjacobs:
This would probably solve #1447, too, so it's not so niche after all 😉 
Thank you very much for this PR!

langsjo:
Using this patch myself and it works to solve the Nix wrapper issue :+1:

PR #1584, opened 2026-05-06
Use `-c` instead of `--` for `use_flox`

mkenigs:
Flox 1.9.0 replaced the behavior of `--` with that of `-c`, and `--` no longer runs profile scripts. Use `-c` to go back to the old behavior

Closes https://github.com/direnv/direnv/pull/1581


mkenigs:
Whoops thought that was an issue not a PR

PR #1586, opened 2026-05-18
docs: don't check in .envrc files

mightyiam:


docs/dont-check-in-dotenvrc.md
```diff
@@ -0,0 +1,21 @@
+# Don’t check in `.envrc` files
+
+`.envrc` files are intended to be a developer’s own personal means of customizing their environment for a project.
+Therefore, as much as the idea of checking in a `.envrc` file may look good, do not be tempted.
+A checked-in `.envrc` file makes for a figurative hoop to jump through before contributing to a project.
```
jfly, line 5: I contribute to a lot of random projects, and the first hoop is always "getting a dev environment". `.envrc` files are a good source of information about something that maintainers might actually use. Encouraging people to just delete that information feels like a net loss.

Could we instead encourage something like a `.envrc.example` file?

mightyiam: I knew someone more reasonable than myself would show up and give great feedback!

jfly (approved):
This seems reasonable to me. No idea if it's philosophically aligned with direnv's maintainers, though.

toastal:
This is the best way to avoid hacky workarounds. Programmers have this bad habit of always trying to code their way out of predicaments they created instead of taking a step back & rethinking the basic premise. The ‘less code’ option of avoiding committing the file in the first place is good for keeping setups simple without exposing oneself to security issues of those workarounds. You can even source the example file in your own `.envrc` if you truly trust upstream while not blocking other downstream users from making needed customization.

Personally I find this incredibly annoying but also quite selfish for upstream to be dictating how your machines environment can/should be as I very often use these files to prepare my environment specific to that project (changing credentials/keys, ports, setting tmpdir, & so on). It takes almost no work from upstream to one-time copy, symlink, or source the example. Adding 2 lines to a README is better than taking away user autonomy. A clear bad example is

```sh-session
$ cat .envrc
use nix
```

This could be replaced with `echo "use nix" >.envrc && direnv allow` in the README. Direnv + Nix users already know how to use/set this up & not all Direnv + Nix users want to start downloading from the cache or get the big, red consent banner when checking out code purely read locally (which I often do since web UIs are a slower experience that doesn’t have my tooling but also works offline), so upstream doesn’t need to provide this for anyone. This applies to $CORPORATE too,  where all ~~code monkeys~~ employees must have the same setup & want everything to be “turn-key”/zero-step setup, sourcing something like `.envrc.shared` would still allow user-level configuration.

mightyiam:
:pray:

Issue #1593, opened 2026-06-24
Add `layout uv` to stdlib


wolf:
This is a follow-up to #1250, #1338, and PR #1352. They've been open since early 2024 without resolution. I've read through all of them. The path forward is clearer now, partly because uv has matured and partly because several commenters already converged on a good answer.

**The disagreements**

Three positions came out of #1250.

The first: "this isn't direnv's job." Some users prefer `uv run` for everything and don't need a venv activated. That's a coherent workflow. It's an argument for not using the feature, not a constraint on its design. Easy answer here: just don't use `layout uv`.

The second: "just activate the venv, don't sync." Create it if it's missing, set `VIRTUAL_ENV` and `PATH`, done. Dependency management stays explicit.

The third: "sync on every cd." Run `uv sync` so the environment stays in lockstep with the lockfile. Fast enough not to matter, and it's the whole point of using uv's project system.

A warm `uv sync --frozen` completes in milliseconds now. The don't-sync objections carried more weight in 2024.

There was also a naming dispute. @anton-ryzhov argued that `layout uv` should be the venv-backend variant, analogous to `layout python`, with a different name for the project-centric one. That stalled PR #1352.

**Good stuff in earlier suggestions**

Several people working independently arrived at the same pieces. @pjz identified that `UV_PROJECT_ENVIRONMENT` needs to be exported so `uv` commands run after activation know which venv to use. @offbyone's PR #1352 had the right structure: `watch_file` on `.python-version`, `pyproject.toml`, and `uv.lock`; `uv sync --frozen`; and `expand_path` to resolve the venv path correctly. @injust landed on the same minimal pattern independently. @dpprdan articulated the `--frozen` requirement most clearly. Sync should read the lockfile as-is and fail if it's stale, never silently modifying a shared lockfile on `cd`.

Two things in #1352 were off: it used `source activate` instead of `PATH_add` + `export VIRTUAL_ENV` (inconsistent with how `layout python` works), and `|| true` swallowed sync errors instead of surfacing them.

**What's better now**

uv has been stable for a while now. The project system (`pyproject.toml` + `uv.lock` + `uv sync`) is the recommended workflow. The pip interface is [explicitly described as legacy](https://docs.astral.sh/uv/getting-started/features/#the-pip-interface). `uv sync` on a warm cache is fast enough to run on every shell entry without noticeable delay.

[...]

Issue #1338, opened 2024-10-12
A layout for uv using a custom project virtual environment path


NikosAlexandris:
The environment variable `UV_PROJECT_ENVIRONMENT` can be used to [configure the project (virtual) environment path](https://docs.astral.sh/uv/concepts/projects/#configuring-the-project-environment-path). Following is an alternative to the current (second) proposed layout for [uv](https://docs.astral.sh/uv) under https://github.com/direnv/direnv/wiki/Python#uv :

```  bash
layout_uv() {
    if [[ -d "$UV_PROJECT_ENVIRONMENT" ]]; then
        log_status "Existing project virtual environment \'$UV_PROJECT_ENVIRONMENT\'"
    else
        log_status "No project virtual environment exists."

        if [[ ! -f "pyproject.toml" ]]; then  # No project in-place 
            log_status "Initializing a new Python project via \`uv init\`."
            uv init --no-readme
        else
            log_status "Python project already initialized. Skipping \`uv init\`."
        fi
        [[ -f "hello.py" ]] && rm hello.py  # a file created by `uv init`

        if [[ ! -z "$UV_PROJECT_ENVIRONMENT" ]]; then
            log_status "Project virtual environment path set to : $UV_PROJECT_ENVIRONMENT"
            uv venv "$UV_PROJECT_ENVIRONMENT"
        else
            uv venv
            UV_PROJECT_ENVIRONMENT="$(pwd)/.venv"
        fi
    fi

    PATH_add "$UV_PROJECT_ENVIRONMENT/bin"
    export UV_ACTIVE=1  # or VENV_ACTIVE=1
    export UV_PROJECT_ENVIRONMENT
}
```

With the above layout command, it should suffice to define the `UV_PROJECT_ENVIRONMENT` in the `.envrc` file, for example : 

``` bash
export UV_PROJECT_ENVIRONMENT=.a_project_virtual_environment
layout uv
```

ssbarnea:
I would argue towards automatically using uv when available, obviously with an option to disable its use as there is a slight chance that some projects might want to avoid it.

PR #1594, opened 2026-06-24
stdlib: add layout_uv for uv project environments

### LINKED ISSUE #1250: official support for uv

wolf:
Implements `layout_uv` as described in #1593. Closes #1250. See also #1338.

Syncs with `--frozen` on every `cd` so the venv stays in lockstep with the lockfile without ever modifying it. Watches `.python-version`, `pyproject.toml`, and `uv.lock` for changes. Accepts an optional Python version argument; otherwise defers to uv's own resolution. Respects `UV_PROJECT_ENVIRONMENT` if set.


wolf:
These failures look like buildbot timeouts, maybe? Will they be re-run?

wolf:
Wow. I am not doing the right thing with respect to contributing guidelines, and not quite sure I understand them. Will attempt.

wolf:
As far as I can tell, contributing guidelines are instructions for maintainers, not for me. But looking around, I do see there are more tests needed than I originally provided. I made a unit tests, but i also needed an integration test. Added.

wolf:
More failures, again, timeouts where a compatible build machine couldn't be found.

stdlib.sh
```diff
@@ -1084,6 +1084,62 @@ layout_pyenv() {
...
+  if [[ ! -f pyproject.toml ]]; then
+    log_error "uv: no pyproject.toml found. Run \`uv init\` to create a project."
+    return 1
+  fi
+
+  local venv_path
+  venv_path="$(expand_path "${UV_PROJECT_ENVIRONMENT:-.venv}")"
+  export UV_PROJECT_ENVIRONMENT="$venv_path"
+
+  local python_arg=()
```
anton-ryzhov, line 1120: It seems [UV_PYTHON](https://github.com/direnv/direnv/pull/1329#discussion_r2619236974) solves that nicer. What do you think?

stdlib.sh
```diff
@@ -1084,6 +1084,62 @@ layout_pyenv() {
...
+  local sync_args=()
+  # uv python specifiers (versions, paths, implementations) never start with
+  # "--", so this distinguishes a python specifier from arguments intended for `uv sync`.
+  if [[ -n "${1:-}" && "${1:-}" != --* ]]; then
+    python_arg=(--python "$1")
+    sync_args=("${@:2}")
+  else
+    sync_args=("$@")
+  fi
```
anton-ryzhov, line 1129: And with UV_PYTHON this part [could be simpler](https://github.com/direnv/direnv/pull/1329#issuecomment-3659233834) as well

stdlib.sh
wolf: `UV_PYTHON` looks like it can have exactly the needed value, but I'm trying to understand the exact scope of where it is defined. The documentation says "defines and respects". So me setting `UV_PROJECT_ENVIRONMENT`, above, is appropriate cooperation with `uv`. While I don't see any rules [here](https://docs.astral.sh/uv/reference/environment/#uv_python) saying otherwise, this **feels** like a variable `uv` would set itself on entry to something... so I would need to test and search. Is this settable? Is that the best approach. If it is handled correctly, then yes absolutely, this would be a superior solution.

stdlib.sh
wolf: My problem with the example you point to is that:

```bash
export UV_PYTHON=3.13
layout uv
```
...is two lines instead of one, and also, doesn't match how `layout python` works.

stdlib.sh
```diff
@@ -1084,6 +1084,62 @@ layout_pyenv() {
...
+
+  # must use --frozen: we don't want to modify the lock file
+  uv sync --frozen "${python_arg[@]}" "${sync_args[@]}"
```
wolf, line 1132: And here's the one place we call `uv` itself. There are multiple ways to address @anton-ryzhov request:

  * have two layout functions with different names, e.g., to support `layout uv` and `layout uv_pip` (or `uv_legacy`, or something else acceptable)
    * ...then the mostly shared internals could be factored, and the caller gets what they want by using the appropriate layout
  * adjust this one call to `uv` within the single layout function; maybe with an `if`, maybe by constructing it (grafting in `pip` or whatever is needed` to get the right thing
    * in this case the caller would probably get what they wanted either with a special parameter, or else by setting an environment variable, or perhaps both would be supported. I would be concerned if support for this were determined by an external file, e.g., the exact shape of the venv, or the existence of specific files like requirements.txt, etc. Those seem more fragile to me.

I think the real question here should start at the human in the equation: how do legacy callers **want** to access this functionality? My personal opinion is two different layout names feels simplest for callers.

wolf (commented):
I will look into `UV_PYTHON` more, and @anton-ryzhov needs to speak for users of legacy behavior: how would you prefer to "get to it"?

stdlib.sh
anton-ryzhov: All these envs are exactly settable and control `uv`'s behavior. [UV_PYTHON](https://docs.astral.sh/uv/reference/environment/#uv_python) is equivalent to `--python` / `-p` argument. But you wouldn't need to parse and handle it.

For `pyproject.toml`-less packages that allows to set concrete package version (`UV_PYTHON=3.14`) without using `.python-version` (might be unwanted in some scenarios)

For `pyproject.toml`-enabled packages (including poetry/pdm) `uv` would respect version definition from that file, no explicit version setting would be needed.

`.python-version` could be used instead of `UV_PYTHON` as well where it suits.

stdlib.sh
anton-ryzhov: That's true, but it could be written as
```sh
UV_PYTHON=3.13 layout uv
```
This also hints that other `UV_` env vars could be used here to tune it, not just one.

> also, doesn't match how layout python works

True, but there is almost nothing to configure in `layout python`. But `uv` is more feature-rich

stdlib.sh
anton-ryzhov: > My personal opinion is two different layout names feels simplest for callers

I agree. Project-agnostic "uvenv" is just the same `venv` but supercharged with `uv`'s speed. While `layout uv` should support `uv`-native project layout. There is no point to mix them together

PR #1615, opened 2026-08-28
Fix PATH being emptied under pwsh due to case-insensitive collision

mschreibjambit:
## What
Fixes an intermittent bug where `direnv` could empty `PATH` under PowerShell (pwsh) on Windows.

## Why
On Windows, `PATH` is inherited from the native process as `Path`, while the Bash process direnv uses to evaluate `.envrc` always normalizes it back to `PATH`. This produced an `EnvDiff`/`ShellExport` containing both a removal of `Path` and an addition of `PATH` for the very same, case-insensitive Windows environment variable.

Since `ShellExport` is a Go map, the `Remove-Item`/`${env:...}=` statements emitted by `pwsh.Export()` were ordered non-deterministically (Go map iteration order is randomized). When the removal happened to run after the addition, it wiped out the `PATH` that had just been set — because PowerShell/Windows environment variables are case-insensitive, `Remove-Item -LiteralPath 'env:/Path'` and `${env:PATH}=...` refer to the same underlying variable. This explains why the failure was intermittent rather than consistent.

## Changes
- `internal/cmd/shell_pwsh.go`: `pwsh.Export()` now collects all keys being added (case-insensitively), and skips any removal whose key collides case-insensitively with one of them, so an addition can never be undone by an unordered removal of the same variable under a different casing.
- `internal/cmd/shell_pwsh_test.go`: new tests covering the `Path`/`PATH` collision (run repeatedly to account for the random map iteration order) and verifying unrelated removals still happen as expected.


mschreibjambit:
Closing — I'd like to keep this fix local to my fork for now rather than open a PR against the upstream project.

PR #1616, opened 2026-08-28
Fix pwsh hook breaking when direnv is installed under a path with spaces

mschreibjambit:
## What
Fixes the PowerShell (pwsh) hook silently failing to update the environment when `direnv` is installed under a path containing a space.

## Why
The pwsh hook substitutes `{{.SelfPath}}` (the absolute path to the `direnv` executable) **unquoted**:

```powershell
$export = ({{.SelfPath}} export pwsh) -join [Environment]::NewLine;
```

Every other shell's hook quotes it (`"{{.SelfPath}}"` in bash, zsh, fish, elvish, murex — see `shell_bash.go`, `shell_zsh.go`, `shell_fish.go`, `shell_elvish.go`, `shell_murex.go`), but pwsh does not.

If `direnv` is installed under a path with a space (very common on Windows, e.g. under `C:\Program Files\...`, or under a user profile with a space in the name), PowerShell splits the command on the space and fails to find it. The hook then silently produces an empty `$export`, so `.envrc` changes (including `PATH` modifications) never get applied — with no visible error, since the failure happens inside the `LocationChangedAction` event handler.

Reproduced locally with real `pwsh` 7.5.4:
```
$export = (/scratchpad/dir with space/direnv export pwsh) -join [Environment]::NewLine;
'/scratchpad/dir' is not recognized as a name of a cmdlet, function, script file, or executable program.
export was: []
```

After the fix (`& "{{.SelfPath}}" export pwsh`), the same path invokes correctly.

## Changes
- `internal/cmd/shell_pwsh.go`: invoke `{{.SelfPath}}` via the call operator with a quoted path: `(& "{{.SelfPath}}" export pwsh)`.
- `internal/cmd/shell_pwsh_test.go`: new test asserting the hook template quotes `{{.SelfPath}}` and uses the call operator.


mschreibjambit:
Closing — I'd like to keep this fix local to my fork for now rather than open a PR against the upstream project.

